PR #1, opened 2022-03-22
Add unit tests for helpers/metadata.go

johnbieren:
This adds basic unit tests. I ran `go fmt` on the file already and the tests pass with `go test .` from the `helpers/` directory

### DIFF
--- a/helpers/metadata_test.go
+++ b/helpers/metadata_test.go
@@ -1,0 +1,81 @@
+package helpers
+
+import (
+	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"testing"
+)
+
+var (
+	testObject      = &tektonv1beta1.PipelineRun{}
+	testAnnotations = map[string]string{
+		"a.test/annotation":              "foo",
+		"another.test/annotation-string": "bar",
+	}
+	testLabels = map[string]string{
+		"a.test/label":              "foo",
+		"another.test/label-string": "bar",
+	}
+)
+
+func init() {
+	testObject.SetAnnotations(testAnnotations)
+	testObject.SetLabels(testLabels)
+}
+
+func TestHasAnnotation(t *testing.T) {
+	checkExistingAnnotation := HasAnnotation(testObject, "a.test/annotation")
+	if checkExistingAnnotation != true {
+		t.Fatalf("HasAnnotation() failed to find existing annotation")
+	}
+
+	checkMissingAnnotation := HasAnnotation(testObject, "bar")
+	if checkMissingAnnotation != false {
+		t.Fatalf("HasAnnotation() found a non-existent annotation")
+	}
+}
+
+func TestHasAnnotationWithValue(t *testing.T) {
+	checkAnnotationProperValue := HasAnnotationWithValue(testObject, "a.test/annotation", "foo")
+	if checkAnnotationProperValue != true {
+		t.Fatalf("HasAnnotationWithValue() failed to find value of existing annotation")
+	}
+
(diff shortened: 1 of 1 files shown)


helpers/metadata_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+)
+
+func init() {
+	testObject.SetAnnotations(testAnnotations)
+	testObject.SetLabels(testLabels)
+}
+
+func TestHasAnnotation(t *testing.T) {
+	checkExistingAnnotation := HasAnnotation(testObject, "a.test/annotation")
```
davidmogar, line 26: It's ok like this, but just in case you weren't aware about a common patter when testing, you can go through multiple tests with something like:
```
var tests = []struct {
    annotation string
    want          bool
}{
    {"a.test/annotation", true}
    {"foo/bar", false}
}

for _, tt := range tests {
    if tt.want != HasAnnotation(testObject, tt.annotation) {
        t.Errorf("wanted %t for annotation '%s', got %t", tt.want, tt.annotation, !tt.want)
    }
}
```

You can see this pattern for repetitive tests here: https://gobyexample.com/testing

davidmogar: Btw, notice that I'm using `Errorf` here to go over every single test case. Fatal would break for the first failure.

helpers/metadata_test.go
```diff
@@ -0,0 +1,81 @@
...
+	if checkExistingAnnotation != true {
```
davidmogar, line 27: Applies to others, but in short:
```suggestion
	if !checkExistingAnnotation {
```
Variable name could be renamed to something that doesn't read like an action. Something like `hasAnnotation` or `exists`. That way the `if` condition would read better.

parasense:
This looks good. :+1: 

I would  suggest adding logging lines (e.g. `t.Logf()') so the tests emit logs to stdout when `go test -v ...` runs verbosely. 
Might be nice to have some verbose log lines at the top or bottom of the test file, or potentially on entry/exit of the Test* functions, and a nice secondary effect is logging lines tend to self-document the code.

https://pkg.go.dev/testing#B.Logf

helpers/metadata_test.go
```diff
@@ -0,0 +1,95 @@
...
+func TestHasAnnotation(t *testing.T) {
+	var tests = []struct {
+		annotation string
+		want       bool
+	}{
+		{"a.test/annotation", true},
+		{"bar", false},
+	}
+
+	t.Logf("Testing that HasAnnotation properly finds object annotations")
```
davidmogar, line 34: I´d say that all this logs are redundant. A test in Go points to a function most of the times, and the function is (or should) be properly defined. This should be changed by proper documentation for each test function. Generally, for a test output, I'm just interested in seeing which test failed (if any). But this is my personal opinion.

helpers/metadata_test.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, test := range tests {
```
davidmogar, line 35: For table driven tests, this variable is called `tt` usually. Kinda like an unwritten rule.

That's it:
```suggestion
	for _, tt := range tests {
```

davidmogar (changes requested):

helpers/metadata_test.go
johnbieren: @parasense wdyt?

helpers/metadata_test.go
johnbieren: Ack, didn't know that. Fixed, thank you

davidmogar (approved):

PR #2, opened 2022-03-29
Add test to ensure all files gofmt'ed in new pull requests

johnbieren:
This will run the test when a PR is opened, reopened, or a commit is added/amended. It has no comment functionality. The test ensures that the files being added abide by the `gofmt` rules


johnbieren:
@davidmogar @parasense please take a look when you have a chance

PR #4, opened 2022-03-30
Add github actions for pull requests

johnbieren:
This commit adds a github action that runs whenever a pull request
is opened, reopened, or a commit is added/updated. It runs go mod,
go fmt, and a static check tests. It does not run a manifest or
go test test. It also ensures that the Dockerfile can be built


johnbieren:
Depends on https://github.com/redhat-appstudio/release-service/pull/3 to start passing.
We have a glibc version issue in the ubuntu container that prevents the manifest test and `go test` from passing, so I left those out of this pr. I will add them in a future pr that also addresses the glibc piece

.github/workflows/pr.yml
```diff
@@ -0,0 +1,101 @@
...
+    steps:
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - name: Check out code into the Go module directory
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Cache Operator SDK ${{ env.OPERATOR_SDK_VERSION }}
```
davidmogar, line 24: Silly comment, but could you move this one step down so all the operator sdk stuff is together?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,101 @@
...
+            echo "generated sources are not up to date:"
+            diff -r ../release-service /tmp/release-service
+            exit 1
+          fi
+      - name: Run Go Tests
+        run: |
+          make test
+  docker:
+    name: Check docker build
+    runs-on: ubuntu-latest
```
davidmogar, line 94: Use ubuntu-20.04 or change the previous one (line 11) to use ubuntu-latest.

johnbieren: Good catch

PR #5, opened 2022-04-07
Add ReleaseLink CRD

johnbieren:
This adds the custom resource definition for the ReleaseLink

The webhook validating that the CR namespace is not equal to the target namespace will be in a separate PR


api/v1alpha1/releaselink_types.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// ReleaseLinkSpec defines the desired state of ReleaseLink
+type ReleaseLinkSpec struct {
+	// DisplayName refers to the name of the Release Link to link a user and managed workspace together.
+	DisplayName string `json:"displayName"`
+
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	// Application to add the component to
```
davidmogar, line 30: Would it make sense to move the description to the first line of the doc block? That way you don’t go through the kubebuilder tags before seeing what it’s actually for. Easier to read IMHO.

api/v1alpha1/releaselink_types.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ReleaseLinkSpec defines the desired state of ReleaseLink
+type ReleaseLinkSpec struct {
+	// DisplayName refers to the name of the Release Link to link a user and managed workspace together.
+	DisplayName string `json:"displayName"`
+
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
```
davidmogar, line 28: I actually don’t like it but code generated by operator sdk doesn’t add an space between `//` and the tag itself. We should be consistent (even if that means change the generated ones).

api/v1alpha1/releaselink_types.go
```diff
@@ -0,0 +1,70 @@
...
+	// Application to add the component to
+	Application string `json:"application"`
+
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	// Target workspace to establish a link with
+	Target string `json:"target"`
+
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// Release Strategy for the managed workspace to execute, not needed in user workspace
```
davidmogar, line 39: Should we add `//+optional`? I’ve seen it in other controllers even when they specified `omitempty`.

api/v1alpha1/releaselink_types.go
```diff
@@ -0,0 +1,70 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
```
davidmogar, line 22: Function docs should end in a period (even when it’s autogenerated). It applies to others in the same file (and maybe existing code 😅).

api/v1alpha1/releaselink_types.go
davidmogar:
```suggestion
	// ReleaseStrategy is a reference to the strategy which will be used on releases of the application. The field has no effect on ReleaseLinks created on unmanaged workspaces
```

api/v1alpha1/releaselink_types.go
davidmogar, line 30:
```suggestion
	// Application is a reference to the application to be releases in a managed workspace
```

api/v1alpha1/releaselink_types.go
```diff
@@ -0,0 +1,70 @@
...
+	DisplayName string `json:"displayName"`
+
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	// Application to add the component to
+	Application string `json:"application"`
+
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	// Target workspace to establish a link with
```
davidmogar, line 35:
```suggestion
	// Target is a reference to the workspace to establish a link with
```

davidmogar (commented):
I added some comments on how docs should be. Go has a particular way to document everything as those comments are used to generate the documentation. In the case of docs for fields, the format should be something like `Name what_it_is` (eg. Name defines…, Name is…). You can see it on existing and generated docs for this very same file.

api/v1alpha1/releaselink_types.go
johnbieren: so you're saying s|// +kube|//+kube| ?

api/v1alpha1/releaselink_types.go
johnbieren: Adding optional is a good idea

api/v1alpha1/releaselink_types.go
johnbieren: I assumed description last was best practice because of https://github.com/redhat-appstudio/application-service/blob/main/api/v1alpha1/component_types.go#L82 if that is not the case though, I can change it. I don't have a strong opinion, but you are probably right about easier to read

PR #8, opened 2022-04-19
Add checks for make build and bundle

johnbieren:
Add checks to our per PR CI that ensures both the
make build and make bundle targets run successfully

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -85,8 +85,11 @@ jobs:
             exit 1
           fi
       - name: Run Go Tests
-        run: |
-          make test
+        run: make test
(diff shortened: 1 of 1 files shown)


johnbieren:
This is blocked from merging for two things:
1) https://github.com/redhat-appstudio/release-service/pull/6 because it needs the change to make the `make bundle` command non-interactive
2) @davidmogar PR about the release CRD so config/samples/kustomization.yaml doesn't complain about the release CRD

johnbieren:
Checked that the new tests pass on HEAD here https://github.com/johnbieren/release-service/pull/19

davidmogar (approved):

PR #10, opened 2022-04-19
Add Realese CRD and controller

davidmogar:
This change adds the new Release CRD which will keep track of releases statuses. The commit includes changes to the previous controller, spreading the old functionality and the new one into two different controllers.

The first one, pipelinerun_controller watches PipelineRuns both from the Build and the Release service. It is in charge of creating new Release objects and of updating their status.

The second one, release_controller, is quite similar to the one we had before. Once a Release object is detected, it will create a new PipelineRun to trigger the release.


controllers/release_controller.go
```diff
@@ -45,103 +44,105 @@ type ReleaseReconciler struct {
...
-		}
-		log.Info("Release triggered",
-			"PipelineRun.Name", pipelineRun.Name,
-			"PipelineRun.Namespace", pipelineRun.Namespace)
-	}
+	return releaseStrategy, nil
 }
 
 // SetupWithManager sets up the controller with the Manager which monitors PipelineRuns and filter them
 // using a predicate to filter out everything but just succeeded Build PipelineRuns.
```
johnbieren, line 131: comment needs update

controllers/release_controller.go
```diff
@@ -45,103 +44,105 @@ type ReleaseReconciler struct {
...
 func (r *ReleaseReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&tektonv1beta1.PipelineRun{}).
-		WithEventFilter(tekton.BuildPipelineRunSucceededPredicate()).
+		For(&v1alpha1.Release{}).
+		WithEventFilter(predicate.GenerationChangedPredicate{}).
```
johnbieren, line 135: what does this call? I guess something with the CreateFunc?

controllers/pipelinerun_controller.go
```diff
@@ -0,0 +1,205 @@
...
+
+// Reconcile is part of the main kubernetes reconciliation loop which aims to
+// move the current state of the cluster closer to the desired state.
+func (r *PipelineRunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	log := r.Log.WithValues("PipelineRun", req.NamespacedName)
+
+	pipelineRun := &tektonv1beta1.PipelineRun{}
+	err := r.Get(ctx, req.NamespacedName, pipelineRun)
+	if err != nil {
+		if errors.IsNotFound(err) {
```
johnbieren, line 55: I see this code is new in both controllers. Why do we need it? Shouldn't the event filter stuff guarantee we have a pipeline run (or in the other case, a release)?

controllers/pipelinerun_controller.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// createRelease creates a new Release by using the information provided in the Build PipelineRun.
+func (r *PipelineRunReconciler) createRelease(ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun) (ctrl.Result, error) {
+	log := r.Log.WithValues()
+
+	log.Info("Creating new Release",
```
johnbieren, line 79: What's the advantage to logging like this instead of passing it as arguments to Log.WithValues like we did before? Or, what is the difference, rather

api/v1alpha1/release_types.go
```diff
@@ -0,0 +1,112 @@
...
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Release `json:"items"`
+}
+
+// NewRelease returns a new Release object created in the same namespace that the given Component.
+func NewRelease(component *v1alpha1.Component, releaseLink *ReleaseLink) *Release {
+	return &Release{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: "release-",
```
johnbieren, line 100: does this add some like uuid after the - or something?

controllers/pipelinerun_controller.go
```diff
@@ -0,0 +1,205 @@
...
+	return helpers.UpdateStatus(r.Client, ctx, release)
+}
+
+// getComponent loads from the cluster the Component referenced in the given PipelineRun. If the PipelineRun doesn't specify
+// a Component or this is not found in the cluster, an error will be returned.
+func (r *PipelineRunReconciler) getComponent(ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun) (*hasv1alpha1.Component, error) {
+	if componentName, found := pipelineRun.Labels["build.appstudio.openshift.io/component"]; found {
+		component := &hasv1alpha1.Component{}
+		err := r.Get(ctx, types.NamespacedName{
+			Namespace: pipelineRun.Namespace,
```
johnbieren, line 146: so the pipeline runs will all be in the namespace of the component, not in some "release-service namespace"

I know I heard about some release-service namespace in meetings a while ago but no idea if that was dropped or not

johnbieren (changes requested):
Some small changes requested, but overall looks good to me. Nicely done. One other thing is that I am not really grasping what we are doing with the release pipeline runs in the pipelinerun_controller. If we are triggered by a release pipeline run, what are we doing exactly?

api/v1alpha1/release_types.go
```diff
@@ -0,0 +1,112 @@
...
+	// Trigger reflects whether the Release is manual or automated
+	// +kubebuilder:validation:Enum=automated;manual
+	// +optional
+	Trigger string `json:"trigger,omitempty"`
+
+	// StartTime is the time when the Release PipelineRun was created and set to run
+	// +optional
+	StartTime *metav1.Time `json:"startTime,omitempty"`
+
+	// CompletionTime is the time the release completed
```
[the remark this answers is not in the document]
davidmogar: Yeah, both actually. The release finishes when the PipelineRun finishes (at the moment).

config/crd/patches/cainjection_in_releases.yaml
```diff
@@ -0,0 +1,7 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
```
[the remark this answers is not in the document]
davidmogar: That’s generated, but I’ll take a look. Thanks!

helpers/status.go
```diff
@@ -0,0 +1,17 @@
+package helpers
```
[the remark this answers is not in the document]
davidmogar: Absolutely!

controllers/release_controller.go
davidmogar: Good eye!

controllers/release_controller.go
davidmogar: This is to ignore status changes. This changes don’t modify the generation.

controllers/pipelinerun_controller.go
davidmogar: It seems like a common pattern across operator sdk so I decided to implement it just to be safe. Take into account that the reconciliation could be repeated multiple times in case of errors and the user could delete resources in the meantime.

controllers/pipelinerun_controller.go
davidmogar: In this case I wanted to print different things. In some logs I wanted the PipelineRun but in others the component.

johnbieren:
Also, by the way, for CI purposes you have to rebase on latest main in order to merge

api/v1alpha1/release_types.go
davidmogar: Yes, it adds a randomly generated string (5 characters?)

controllers/pipelinerun_controller.go
davidmogar: Build ones in the component namespace. Ours in the target workspace.

davidmogar:
> Also, by the way, for CI purposes you have to rebase on latest main in order to merge

Absolutely! My bad.

davidmogar:
> One other thing is that I am not really grasping what we are doing with the release pipeline runs in the pipelinerun_controller. If we are triggered by a release pipeline run, what are we doing exactly?

We are basically monitoring the state of the PipelineRun to update the state of the Release. That way we can show if the release succeeds or not without having to check the PipelineRun (which could disappear at some point).

config/crd/patches/cainjection_in_releases.yaml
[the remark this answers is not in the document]
johnbieren: At least for me, the webhook operator-sdk command generated it, not just the normal crd one

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
johnbieren: Maybe capital R then for clarity (I see you used capitals for the CRs in other spots in this PR and it made reading easier for me)

controllers/pipelinerun_controller.go
johnbieren: ah, this function is only used in the build pipeline tree of logic. Got it, thanks

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
davidmogar: I was referring to the general concept of release, not to the object or the PipelineRun, but yeah, this can be capital R if that’s clearer.

api/v1alpha1/release_types.go
```diff
@@ -0,0 +1,112 @@
...
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec ReleaseSpec `json:"spec,omitempty"`
+
+	//+kubebuilder:default={trigger: "manual"}
+	Status ReleaseStatus `json:"status,omitempty"`
+}
+
+//+kubebuilder:object:root=true
```
johnbieren, line 87: Needs space for consistency

johnbieren (approved):
You need to regenerate your manifests and one small change I added

PR #12, opened 2022-04-21
Add ReleaseLink validation to release controller

johnbieren:
Add validations to the release controller to check that the
ReleaseLink referenced in the Release both targets an existing
workspace and that there is a corresponding ReleaseLink in the
target workspace pointing back to the ReleaseLink. If either is
untrue, error out.

### DIFF
--- a/controllers/release_controller.go
+++ b/controllers/release_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"github.com/go-logr/logr"
 	"github.com/redhat-appstudio/release-service/api/v1alpha1"
 	"github.com/redhat-appstudio/release-service/tekton"
@@ -74,9 +75,18 @@ func (r *ReleaseReconciler) triggerReleasePipeline(ctx context.Context, release 
 		return ctrl.Result{}, nil
 	}
 
+	_, err = r.getTargetReleaseLink(ctx, releaseLink)
+	if err != nil {
+		log.Error(err, "Failed to find a matching ReleaseLink in target workspace", "ReleaseLink.Target", releaseLink.Spec.Target)
+		release.Status.SetErrorCondition(err)
+
+		return ctrl.Result{}, nil
+	}
+
 	releaseStrategy, err := r.getReleaseStrategy(ctx, releaseLink)
 	if err != nil {
 		log.Error(err, "Failed to get ReleaseStrategy")
+		release.Status.SetErrorCondition(err)
 
 		return ctrl.Result{}, nil
 	}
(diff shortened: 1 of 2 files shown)


johnbieren:
@davidmogar so we did need the changes to `SetupWithManager` as you predicted. Two areas I'd like to highlight for your review
1) Do the errors make sense? Are they too redundant?
2) I changed the order in the controller. Instead of getting the `releaseStrategy` before validating the `ReleaseLink`, I did all the validation first. I think it logically makes more sense this way

controllers/release_controller.go
```diff
@@ -74,9 +76,26 @@ func (r *ReleaseReconciler) triggerReleasePipeline(ctx context.Context, release
...
+		log.Error(err, "Failed to find ReleaseLink's target workspace")
+		release.Status.SetErrorCondition(err)
+
+		return ctrl.Result{}, nil
+	}
+
+	targetReleaseLink, err = r.getTargetReleaseLink(ctx, releaseLink)
+	if err != nil {
+		log.Error(err, "Failed to find a matching ReleaseLink in Target workspace", "ReleaseLink.Target", releaseLink.Spec.Target)
+		release.Status.SetErrorCondition(err)
```
johnbieren, line 90: @davidmogar we still need to talk about what condition we want to set here

controllers/release_controller.go
```diff
@@ -74,9 +76,26 @@ func (r *ReleaseReconciler) triggerReleasePipeline(ctx context.Context, release
...
 
+	err = r.checkReleaseLinkTargetExists(ctx, releaseLink)
+	if err != nil {
+		log.Error(err, "Failed to find ReleaseLink's target workspace")
+		release.Status.SetErrorCondition(err)
+
+		return ctrl.Result{}, nil
+	}
+
+	_, err = r.getTargetReleaseLink(ctx, releaseLink)
```
johnbieren, line 87: @davidmogar This line should be
`targetReleaseLink, err := r.getTargetReleaseLink(ctx, releaseLink)`
but then it fails for unused variable (it won't be used until your PR). So you'll have to change to that line as part of your PR

controllers/release_controller.go
```diff
@@ -127,8 +137,41 @@ func (r *ReleaseReconciler) getReleaseStrategy(ctx context.Context, releaseLink
...
+			return &foundReleaseLink, nil
+		}
+	}
+	return nil, fmt.Errorf("no ReleaseLink found in target workspace %s with application %s",
+		releaseLink.Spec.Target, releaseLink.Spec.Application)
+}
+
 // SetupWithManager sets up the controller with the Manager which monitors new Releases and filters out status updates.
 func (r *ReleaseReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	// Add a caches to be able to search for ReleaseLinks by Target or Application value
```
davidmogar, line 167:
```suggestion
	// Add a cache to be able to search for ReleaseLinks by target
```

controllers/release_controller.go
```diff
@@ -127,8 +137,41 @@ func (r *ReleaseReconciler) getReleaseStrategy(ctx context.Context, releaseLink
...
+	}
+
+	if err := r.List(ctx, releaseLinks, opts...); err != nil {
+		return nil, err
+	}
+	if len(releaseLinks.Items) == 0 {
+		return nil, fmt.Errorf("no ReleaseLink found with target %s in %s workspace",
+			releaseLink.Namespace, releaseLink.Spec.Target)
+	}
+	for _, foundReleaseLink := range releaseLinks.Items {
```
davidmogar, line 156: Do you mind to add an empty line before this one?

controllers/release_controller.go
```diff
@@ -127,8 +137,41 @@ func (r *ReleaseReconciler) getReleaseStrategy(ctx context.Context, releaseLink
...
+	releaseLinks := &v1alpha1.ReleaseLinkList{}
+	opts := []client.ListOption{
+		client.InNamespace(releaseLink.Spec.Target),
+		client.MatchingFields{"spec.target": releaseLink.Namespace},
+	}
+
+	if err := r.List(ctx, releaseLinks, opts...); err != nil {
+		return nil, err
+	}
+	if len(releaseLinks.Items) == 0 {
```
davidmogar, line 152: Let's remove this and improve the comment in line 161. The check is already covered by the loop.

controllers/release_controller.go
```diff
@@ -74,9 +75,18 @@ func (r *ReleaseReconciler) triggerReleasePipeline(ctx context.Context, release
 		return ctrl.Result{}, nil
 	}
 
+	_, err = r.getTargetReleaseLink(ctx, releaseLink)
+	if err != nil {
+		log.Error(err, "Failed to find a matching ReleaseLink in Target workspace", "ReleaseLink.Target", releaseLink.Spec.Target)
```
davidmogar, line 80:
```suggestion
		log.Error(err, "Failed to find a matching ReleaseLink in target workspace", "ReleaseLink.Target", releaseLink.Spec.Target)
```

PR #17, opened 2022-04-24
Add release webhook

davidmogar:
The new webhook will forbid updates to Releases as these resources should not be updated once created. At that point the controller is already processing them and adding data specific to the fields in the resource specification.

### DIFF
--- a/api/v1alpha1/release_webhook.go
+++ b/api/v1alpha1/release_webhook.go
@@ -1,0 +1,55 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


api/v1alpha1/release_webhook.go
```diff
@@ -0,0 +1,55 @@
...
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *Release) ValidateCreate() error {
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *Release) ValidateUpdate(old runtime.Object) error {
+	releaselog.Info("validate update", "name", r.Name)
+
+	return fmt.Errorf("release resources cannot be udpated")
```
johnbieren, line 49: s/udpated/updated/

davidmogar: Autogenerated code. It's fine ;)

johnbieren (approved):

api/v1alpha1/release_webhook.go
davidmogar: My bad, fixed.

PR #22, opened 2022-04-26
Remove cert-manager

davidmogar:
This change follows what application-service did for enabling webhooks on staging [1]. With this change, webhooks will not work on CRC.

[1] https://github.com/redhat-appstudio/application-service/pull/52


johnbieren:
With this change, they don't work on my local PSI setup either. If we're going to break our local setups, can we have a flag in the Makefile for this? Having to change the repo locally and reapply things all the time would be pretty non-ideal IMO (having to recreate the `bin/` files is like this already and it is annoying)

johnbieren (approved):
I suppose it is okay

PR #25, opened 2022-05-04
feat(HACBS-402): add conventional-commit checks

johnbieren:
Add conventional-commit checking that will be run on commit messages
that are added in new pull requests. Run the check as part of pr github
action.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -100,3 +100,17 @@ jobs:
           fetch-depth: 0
       - name: Check if dockerimage build is working
         run: docker build -f ./Dockerfile .
+  gitlint:
+    name: Run gitlint checks
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Install gitlint into container
+        run: python -m pip install gitlint
+      - name: Run gitlint check
+        run: gitlint --commits origin/${{ github.event.pull_request.base.ref }}..HEAD
+ 
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,14 @@
+[general]
+contrib=contrib-title-conventional-commits,CC1
+ignore=T5
+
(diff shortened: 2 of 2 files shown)


johnbieren:
I took all the rules product CI is using

.gitlint
```diff
@@ -0,0 +1,12 @@
+[general]
+ignore=T5
```
davidmogar, line 2: Could you add `contrib=contrib-title-conventional-commits,CC1` here and remove it from the action?
```suggestion
contrib=contrib-title-conventional-commits,CC1
ignore=T5
```

.gitlint
```diff
@@ -0,0 +1,12 @@
...
+
+[title-max-length]
+line-length=72
```
davidmogar, line 5: Both this and the next one are the default values, but yeah, probably better to make it clear.

.gitlint
```diff
@@ -0,0 +1,12 @@
...
+
+[body-max-line-length]
+line-length=72
+
+[body-first-line-empty]
+
+[contrib-title-conventional-commits]
```
davidmogar, line 12: Missing new line at the end.

.github/workflows/pr.yml
```diff
@@ -100,3 +100,16 @@ jobs:
...
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Install gitlint into container
+        run: python -m pip install gitlint
+      - name: Run gitlint check
+        run: gitlint --config .gitlint --contrib contrib-title-conventional-commits,CC1 --commits origin/${{ github.event.pull_request.base.ref }}..HEAD
```
davidmogar, line 115: Missing new line at the end.

Also, as per the comment below:
```suggestion
        run: gitlint --commits origin/${{ github.event.pull_request.base.ref }}..HEAD
```
I also removed the `--config` option as we are using the default.

davidmogar (commented):
Neat change! Just some minor comments.

davidmogar (approved):

PR #26, opened 2022-05-05
feat(HACBS-390): create mock ApplicationSnapshot CRD

johnbieren:
Create an ApplicationSnapshot CRD to unblock us for now. This commit
is meant to be reverted at a later date once the Gitops team provides
an ApplicationSnapshot CRD.


api/v1alpha1/applicationsnapshot_types.go
```diff
@@ -0,0 +1,67 @@
...
+	// +required
+	Component string `json:"component"`
+
+	// PullSpec is the particular version of the component's container image
+	// +required
+	PullSpec string `json:"pullSpec"`
+}
+
+// ApplicationSnapshotSpec defines the desired state of ApplicationSnapshot
+type ApplicationSnapshotSpec struct {
```
johnbieren, line 35: @davidmogar do we want a reference to the Application in here too? I went back and forth in my head on it

davidmogar: This is actually a good point. I think we still need to discuss about how these three CRDs (Application, ApplicationSnapshotEnvironmentBinding and ApplicationSnapshot) should be linked together. We went over it briefly in the past but nothing was agreed.

In my opinion having something like the following would make things easier:
* Snapshot CRD: List of images not linked to any Application.
* Bindings field in the Application CRD: It would be a list of named tuples:
```
bindings:
  - snapshot: ID
    environment: prod
```

johnbieren:
> Do we need the component or that can be inferred from the pullspec?

You tell me. I was thinking you would need it if you want to use the mapping. Like given pullspec and mapping: A->Z, B->Y, what component do you have? If it is programmatic to get the component from pullspec, I am fine to drop it but I don't know that

johnbieren:
@davidmogar I am also assuming that the name field suffices for "snapshot id"

davidmogar (approved):

PR #29, opened 2022-05-12
fix: remove bin directory

davidmogar:
It seems that the bin directory was pushed at some point before adding it to the gitignore file. This might be causing some CI problems.


johnbieren (approved):
Thank goodness. Now I can stop stashing my bin files whenever I rebase

PR #31, opened 2022-05-13
fix: remove deploy-key secret

mbharatk:
This PR deletes the infra-deployments-deploy-key secret, as there is a sharedSecret created for all the namespaces, there is no need for an individual secret to be added to each namespace.

Signed-off-by: Bharat Kumar Mallavarapu mbharatk@redhat.com

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -30,6 +30,3 @@ spec:
      persistentVolumeClaim: 
         claimName: app-studio-default-workspace 
      subPath: release-service-push-{{ revision }}
-   - name: deploy-key
-     secret:
-        secretName: infra-deployments-deploy-key


davidmogar:
Hey Bharat, thank you for the contribution!

We are requiring signed commits following the [Conventional Commits specification](https://www.conventionalcommits.org/en/v1.0.0/). We are ok merging this PR even if unsigned, but could you please amend your commit and fix it so it adders to the specification?

Content of the commit following to this specification would look like:
```
fix: remove deploy-key secret

This PR deletes the infra-deployments-deploy-key secret, as there is a
sharedSecret created for all the namespaces, there is no need for an
individual secret to be added to each namespace.

Signed-off-by: Bharat Kumar Mallavarapu <mbharatk@redhat.com>
```

I'm pretty sure that you are aware of it, but just in case, you can just run `git commit --amend -s` to edit your commit.

Michkov (approved):

PR #32, opened 2022-05-16
feat(HACBS-392): pass ApplicationSnapshot to Release pipeline

johnbieren:
Extract the ApplicationSnapshot from the Release CR that is being processed
and pass that as a parameter to the Release pipeline.

### DIFF
--- a/controllers/release/release_adapter.go
+++ b/controllers/release/release_adapter.go
@@ -19,6 +19,8 @@ package release
 import (
 	"context"
 	"fmt"
+	"time"
+
 	"github.com/go-logr/logr"
 	"github.com/redhat-appstudio/release-service/api/v1alpha1"
 	"github.com/redhat-appstudio/release-service/controllers/results"
@@ -29,7 +31,6 @@ import (
 	"knative.dev/pkg/apis"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"time"
 )
 
 // Adapter holds the objects needed to reconcile a Release.
@@ -120,7 +121,13 @@ func (a *Adapter) EnsureReleasePipelineRunExists() (results.OperationResult, err
 			return results.RequeueOnErrorOrStop(a.updateStatus())
 		}
 
-		pipelineRun, err = a.createReleasePipelineRun(releaseStrategy)
+		applicationSnapshot, err := a.getApplicationSnapshot()
+		if err != nil {
+			a.release.Status.MarkInvalid(v1alpha1.ReleaseReasonValidationError, err.Error())
+			return results.RequeueOnErrorOrStop(a.updateStatus())
+		}
+
+		pipelineRun, err = a.createReleasePipelineRun(releaseStrategy, applicationSnapshot)
 		if err != nil {
 			return results.RequeueWithError(err)
 		}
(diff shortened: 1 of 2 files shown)


tekton/pipeline_run.go
```diff
@@ -66,6 +68,13 @@ func CreateReleasePipelineRun(release *v1alpha1.Release, releaseStrategy *v1alph
 				Name:   releaseStrategy.Spec.Pipeline,
 				Bundle: releaseStrategy.Spec.Bundle,
 			},
+			Params: []tektonv1beta1.Param{{ // This is silently ignored if pipeline called doesn't declare this parameter
```
davidmogar, line 71: If we go through with my PR, this could be done with something like the following:

```
func (pr *ReleasePipelineRun) WithApplicationSnapshot(snapshot *v1alpha1.ApplicationSnapshot) *ReleasePipelineRun {
	snapshotString, err := json.Marshal(snapshot.Spec)
	if err != nil {
		return nil
	}

	pr.Spec.Params = append(pr.Spec.Params, tektonv1beta1.Param{
		Name: snapshot.Kind,
		Value: tektonv1beta1.ArrayOrString{
			Type:      tektonv1beta1.ParamTypeString,
			StringVal: string(snapshotString),
		},
	})

	return pr
}
```

johnbieren: Yes, will wait to base this PR on that one

controllers/release/release_adapter.go
```diff
@@ -194,6 +207,22 @@ func (a *Adapter) finalizeRelease() error {
 	return nil
 }
 
+// getApplicationSnapshot returns the ApplicationSnapshot referenced by the Release being processed. If the
+// ApplicationSnapshot is not found or the Get operation failed, an error will be returned.
+func (a *Adapter) getApplicationSnapshot() (*v1alpha1.ApplicationSnapshot, error) {
+	applicationSnapshot := &v1alpha1.ApplicationSnapshot{}
+	err := a.client.Get(a.context, types.NamespacedName{
+		Namespace: a.release.Namespace,
+		Name:      a.release.Spec.ApplicationSnapshot,
```
davidmogar, line 216: ABC 215-216 please :wink:

johnbieren: :slightly_smiling_face: sure thing, will do when I rebase on !33

tekton/pipeline_run.go
```diff
@@ -101,6 +103,23 @@ func (r *ReleasePipelineRun) WithReleaseLabels(releaseName, releaseNamespace str
 	return r
 }
 
+// WithApplicationSnapshot adds a param containing the ApplicationSnapshot as a json string
+// to the release PipelineRun.
+func (r *ReleasePipelineRun) WithApplicationSnapshot(snapshot *v1alpha1.ApplicationSnapshot) *ReleasePipelineRun {
+	snapshotString, err := json.Marshal(snapshot.Spec)
+	if err != nil {
+		return nil
```
davidmogar, line 111: This is wrong. You should never return nil in a chain or it will break the code. If we need to send an error back, we can do it by adding code. In this particular case, it should always work as a CRD will always generate a valid JSON so I'd say that we can remove it.

tekton/pipeline_run.go
```diff
@@ -101,6 +103,23 @@ func (r *ReleasePipelineRun) WithReleaseLabels(releaseName, releaseNamespace str
...
+	}
+
+	snapshotValue := tektonv1beta1.ArrayOrString{
```
davidmogar, line 114: I wouldn't create a new variable for this one.

davidmogar (changes requested):

tekton/pipeline_run.go
```diff
@@ -101,6 +103,19 @@ func (r *ReleasePipelineRun) WithReleaseLabels(releaseName, releaseNamespace str
...
 }
 
+// WithApplicationSnapshot adds a param containing the ApplicationSnapshot as a json string
+// to the release PipelineRun.
+func (r *ReleasePipelineRun) WithApplicationSnapshot(snapshot *v1alpha1.ApplicationSnapshot) *ReleasePipelineRun {
+	snapshotString, _ := json.Marshal(snapshot.Spec)
+
+	r.WithExtraParam(snapshot.Kind, tektonv1beta1.ArrayOrString{
+		Type:      tektonv1beta1.ParamTypeString,
+		StringVal: string(snapshotString),
```
johnbieren, line 113: This line is WIP because tekton does crazy things with string variables

davidmogar: I converted this PR into a draft as per your comment.

tekton/pipeline_run.go
```diff
@@ -102,6 +104,19 @@ func (r *ReleasePipelineRun) WithReleaseLabels(releaseName, releaseNamespace str
 	return r
 }
 
+// WithApplicationSnapshot adds a param containing the ApplicationSnapshot as a json string
+// to the release PipelineRun.
+func (r *ReleasePipelineRun) WithApplicationSnapshot(snapshot *v1alpha1.ApplicationSnapshot) *ReleasePipelineRun {
+	snapshotString, _ := json.Marshal(snapshot.Spec)
```
davidmogar, line 110: Can we maybe add a comment here saying that we ignore the error as this should not happen when marshalling specs and that if we end up needing this we will need to pass the errors through the chain and something like a `Complete` function returning the final object and an error?

davidmogar (approved):

PR #33, opened 2022-05-17
feat(HACBS-470): refactor code to add ReleaseStrategy params

davidmogar:
This change adds some changes into the pipeline_run.go file so it's possible to create empty ReleasePipelineRuns (without spec). To actually pass information to the PipelineRun, With* functions can be used. The new functions allow to specify the release labels, the pipeline reference, parameters, etc.


config/crd/bases/appstudio.redhat.com_releasestrategies.yaml
```diff
@@ -36,20 +36,39 @@ spec:
...
+                  description: Param declares an ArrayOrString to use for the parameter
+                    called name.
                   properties:
                     name:
                       type: string
                     value:
-                      items:
-                        type: string
-                      type: array
+                      description: ArrayOrString is a type that can hold a single
```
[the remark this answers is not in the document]
davidmogar: Similarly to the other ones, they come from the code documentation. In this case, those comments are in the upstream Tekton PipelineRun CRD.

tekton/pipeline_run.go
```diff
@@ -48,24 +49,56 @@ var (
 	ReleaseWorkspaceLabel = fmt.Sprintf("%s/%s", releaseLabelPrefix, "workspace")
 )
 
-// CreateReleasePipelineRun creates a PipelineRun from a given Release and ReleaseStrategy.
-// Release details are added to the labels of the new PipelineRun to be able to reference it later on.
-func CreateReleasePipelineRun(release *v1alpha1.Release, releaseStrategy *v1alpha1.ReleaseStrategy) *tektonv1beta1.PipelineRun {
-	return &tektonv1beta1.PipelineRun{
+type ReleasePipelineRun struct {
```
johnbieren, line 52: It is not clear to me why we are defining this struct instead of just using the normal `tektonv1beta1.PipelineRun`

controllers/release/release_adapter.go
```diff
@@ -279,7 +282,7 @@ func (a *Adapter) getTargetReleaseLink() (*v1alpha1.ReleaseLink, error) {
 // registerReleasePipelineRunStatus updates the status of the Release being processed by monitoring the status of the
 // associated release PipelineRun and setting the appropriate state in the Release. If the PipelineRun hasn't
 // started/succeeded, no action will be taken.
-func (a *Adapter) registerReleasePipelineRunStatus(pipelineRun *v1beta1.PipelineRun) error {
+func (a *Adapter) registerReleasePipelineRunStatus(pipelineRun *tekton.ReleasePipelineRun) error {
```
johnbieren, line 285: See my comment pin `pipeline_run.go` about the new type but if we are using a new type, can't we just change the one last reference to `v1beta1` and remove the import entirely from this file?

tekton/pipeline_run.go
davidmogar: You cannot add new methods in a different file from where the receiver is defined.

controllers/release/release_adapter.go
```diff
@@ -161,8 +161,11 @@ func (a *Adapter) EnsureReleasePipelineStatusIsTracked() (results.OperationResul
 
 // createReleasePipelineRun creates and returns a new release PipelineRun. The new PipelineRun will include owner
 // annotations, so it triggers Release reconciles whenever it changes.
```
johnbieren, line 163: I'd like a quick couple words about release strategy in the comment too

controllers/release/release_adapter.go
davidmogar: I don’t really get this comment, sorry. Could you rephrase it?

config/crd/bases/appstudio.redhat.com_releasestrategies.yaml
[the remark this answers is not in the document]
johnbieren: Oh that's sweet

controllers/release/release_adapter.go
davidmogar: Will change it.

tekton/pipeline_run.go
johnbieren: That's unfortunate. Can you comment pretty much exactly that before defining this then?

davidmogar: Can do.

controllers/release/release_adapter.go
davidmogar: Oh, got you now. I was confused because the discussion was for a different line.

No, I don't think it's possible to do that as a ReleasePipelineRun is not a PipelineRun (it's something more), so listing would not work.

johnbieren: I thought the same but figured it was worth asking the go expert

davidmogar: I’d like to meet that guy too.

tekton/pipeline_run.go
```diff
@@ -48,24 +49,69 @@ var (
...
+
+	return &ReleasePipelineRun{pipelineRun}
+}
+
+// AsPipelineRun casts the ReleasePipelineRun to PipelineRun, so it can be used in the Kubernetes client.
+func (r *ReleasePipelineRun) AsPipelineRun() *tektonv1beta1.PipelineRun {
+	return &r.PipelineRun
+}
+
+// WithExtraParam adds an extra param to the release PipelineRun.
```
johnbieren, line 75: What do you think about adding a comment along the lines of "If the parameter is not part of the invoked release Pipeline definition, it is silently ignored"

I say this just because when I was implementing this, it is something I thought of and tested

johnbieren (approved):
One small comment but otherwise lgtm

PR #37, opened 2022-05-24
fix(HACBS-509): prevent spec updates exclusively

davidmogar:
The Release admission webhook was preventing the objects to be updated to add a finalizer. The admission webhook for updates should exclusively prevent spec updates.

### DIFF
--- a/api/v1alpha1/release_webhook_test.go
+++ b/api/v1alpha1/release_webhook_test.go
@@ -17,78 +17,62 @@ package v1alpha1
 
 import (
 	"context"
-	"reflect"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
 	//+kubebuilder:scaffold:imports
 )
 
 var _ = Describe("Release validation webhook", func() {
-	// Define utility constants for object names
-	const (
-		Name                = "test-release-"
-		Namespace           = "default"
-		ApplicationSnapshot = "test-snapshot"
-		ReleaseLink         = "test-releaselink"
-	)
+	var release *Release
+
+	BeforeEach(func() {
+		release = &Release{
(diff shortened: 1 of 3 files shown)


johnbieren (commented):
Can you add a test that updates something besides the spec and does not throw error?

api/v1alpha1/release_webhook_test.go
```diff
@@ -17,78 +17,60 @@ package v1alpha1
...
-			// Look up the Release resource that was created
-			releaseLookupKey := types.NamespacedName{Name: release.Name, Namespace: Namespace}
-			createdRelease := &Release{}
-			Eventually(func() bool {
-				k8sClient.Get(ctx, releaseLookupKey, createdRelease)
-				return !reflect.DeepEqual(createdRelease, &Release{})
-			}, timeout, interval).Should(BeTrue())
+			release.Spec.ApplicationSnapshot = "another-snapshot"
 
-			// Try to update the Release application snapshot
```
johnbieren, line 67: Can we put this comment back before new line 57?

api/v1alpha1/release_webhook_test.go
```diff
@@ -17,78 +17,60 @@ package v1alpha1
...
-	})
 
-})
+		It("Should not error out when updating the resource metadata", func() {
+			ctx := context.Background()
+
+			Expect(k8sClient.Create(ctx, release)).Should(Succeed())
 
-// deleteReleaseCR deletes the specified Release resource and verifies it was properly deleted.
-func deleteReleaseCR(releaseLookupKey types.NamespacedName) {
```
johnbieren, line 81: Why is this test being removed?

johnbieren: Sorry. I thought it was a test for the `ValidateDelete` function of the webhook https://github.com/redhat-appstudio/release-service/blob/main/api/v1alpha1/release_webhook.go#L53
If it just removes the test resource, agreed let's remove it

johnbieren (approved):

PR #38, opened 2022-05-24
fix(HACBS-510): fail on description starting with uppercase

davidmogar:
We are following Conventional Commits and using Gitlint to enforce the specification. Although the description doesn't have a set-on-stone format, examples on Conventional Commits spec seem to use always a sentence starting with lowercase. This is not enforced by Gitlint, but I think it would be good to enforce it. First because all our commits will follow the same convention. Second because a new user following the spec will already think this is the proper way.

### DIFF
--- a/.github/gitlint/contrib_description_conventional_commits.py
+++ b/.github/gitlint/contrib_description_conventional_commits.py
@@ -1,0 +1,26 @@
+# -*- coding: utf-8 -*-
+from gitlint.rules import CommitRule, RuleViolation
+
+
+class ConventionalCommitsDescription(CommitRule):
+    """This rule will enforce that each commit description starts with
+    a lowercase."""
+
+    id = "UC1"
+    name = "contrib-description-conventional-commits"
+
+    def validate(self, commit):
+        """Validate the given commit checking that the description starts with
+        a lowercase."""
+        if ":" not in commit.message.title:
+            self.log.debug("Title does not follow conventional commit specification")
+            return
+
+        description = commit.message.title.split(":")[1].strip()
+
+        if len(description) > 0 and description[0].isupper():
+            return [
+                RuleViolation(
+                    self.id, "Description should start with lowercase", line_nr=1
+                )
+            ]
--- a/.gitlint
+++ b/.gitlint
@@ -1,5 +1,6 @@
 [general]
 contrib=contrib-title-conventional-commits,CC1
+extra-path=.github/gitlint
 ignore=T5
 
 [title-max-length]


.github/gitlint/contrib_description_conventional_commits.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from gitlint.rules import CommitRule, RuleViolation
+
+
+class SignedOffBy(CommitRule):
```
johnbieren, line 5: Why is this called signed off by? It has nothing to do with signing

davidmogar: Because I love to copy paste brainlessly.

parasense: SignOffBy means `git commit -s ...`

Also, yes it has nothing to-do with cryptographic signing.

Git doc ref:
https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--s

Commit conventions:
https://git.wiki.kernel.org/index.php/CommitMessageConventions

.github/gitlint/contrib_description_conventional_commits.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def validate(self, commit):
+        if ':' not in commit.message.title:
+            self.log.debug("Title does not follow conventional commit specification")
+            return
+
+        description = commit.message.title.split(':')[1].strip()
+
+        if len(description) > 0 and description[0].isupper():
+            return [RuleViolation(self.id, "Description should start with lowercase", line_nr=1)]
```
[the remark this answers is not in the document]
davidmogar: I don't see what is the actual change. I don't even know why the whole file appears like that.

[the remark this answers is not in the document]
parasense: Please consider using a python linter, potentially from github action or through your text editor.
```
python3 -m pip install --user --no-cache-dir --upgrade pip
python3 -m pip install --user --no-cache-dir flake8 black
```

And then run `flake8` to see existing issues.
```
$ flake8 .github/gitlint/contrib_description_conventional_commits.py
.github/gitlint/contrib_description_conventional_commits.py:15:80: E501 line too long (85 > 79 characters)
.github/gitlint/contrib_description_conventional_commits.py:21:80: E501 line too long (97 > 79 characters)
```

Or, one can simply run `black` to auto-correct python mistakes, or as a linter:
```
$ black --diff .github/gitlint/contrib_description_conventional_commits.py
--- .github/gitlint/contrib_description_conventional_commits.py	2022-05-26 13:27:17.032813 +0000
+++ .github/gitlint/contrib_description_conventional_commits.py	2022-05-26 13:39:37.650458 +0000
@@ -1,21 +1,25 @@
 # -*- coding: utf-8 -*-
 from gitlint.rules import CommitRule, RuleViolation
 
 
 class ConventionalCommitsDescription(CommitRule):
-    """ This rule will enforce that each commit description starts with
+    """This rule will enforce that each commit description starts with
     a lowercase.
     """
 
     id = "UC1"
     name = "contrib-description-conventional-commits"
 
     def validate(self, commit):
-        if ':' not in commit.message.title:
+        if ":" not in commit.message.title:
             self.log.debug("Title does not follow conventional commit specification")
             return
 
-        description = commit.message.title.split(':')[1].strip()
+        description = commit.message.title.split(":")[1].strip()
 
         if len(description) > 0 and description[0].isupper():
-            return [RuleViolation(self.id, "Description should start with lowercase", line_nr=1)]
+            return [
+                RuleViolation(
+                    self.id, "Description should start with lowercase", line_nr=1
+                )
+            ]
would reformat .github/gitlint/contrib_description_conventional_commits.py

All done! ✨ 🍰 ✨
1 file would be reformatted.
```

Run `black` without the `--check` or `--diff` to get in-place auto-correct.
Then run `flake8` again to see the lack of any linting issues.
I would recommend using `--max-line-length 88` so `flake8` and `black` agree on line-length.

.github/gitlint/contrib_description_conventional_commits.py
```diff
@@ -0,0 +1,21 @@
...
+    a lowercase.
+    """
+
+    id = "UC1"
+    name = "contrib-description-conventional-commits"
+
+    def validate(self, commit):
+        if ':' not in commit.message.title:
+            self.log.debug("Title does not follow conventional commit specification")
+            return
```
parasense, line 16: I suspect this rule wants to impose the idea of overloading the git commit subject line with metadata?

Can you briefly explain what the subject line would ideally look like using this rule? Maybe paste a few examples?

The reason I ask, is because I want to ensure the subject-lines stay high quality:
- have enough room to use the imperative mood inside 80'ish characters, and 
- not get over-loaded with metadata that could potentially go elsewhere in the git commit message body.

https://en.wikipedia.org/wiki/Imperative_mood

If the idea is to have something like:
- `Bug fix: stop doing that thing that broke something`
- `Enhancement: implement cool new feature`

Then I'm cool..

But, if the idea is to inject Jira issues into the subject-line, then not cool...

Instead, I would suggest adding a rule to parse for jira `project-id` string in the commit message, something like:
```
New feature: imperative mood subject-line

Paragraphs of message text
. . .

jira: hacbs-99999

By-Lines: Captain Crepe <cpt.crepe@example.com>
...
```

https://wiki.openstack.org/wiki/GitCommitMessages#Information_in_commit_messages

davidmogar: Specification: https://www.conventionalcommits.org/en/v1.0.0/

davidmogar:
@parasense can this be merged?

davidmogar:
> If we agreed on using the conventional commit I don't see issue on merging this one. I think you are even being soft on it. I would as well enforce checking that the ticket ID is present, but I don't know if we have any edge case that would make this rule impossible.

Ticket id (scope) is optional according to the [specification](https://www.conventionalcommits.org/en/v1.0.0/).

theflockers (approved):

PR #39, opened 2022-05-25
fix(HACBS-513): use go install to fetch binaries

davidmogar:
The operator Makefile was outdated for recent Go versions. Specifically, the routine to install Go binaries was deprecated and should be using `go install` instead of `go get`.


johnbieren (commented):
Are these changes operator-sdk generated or manual by you? I see the kustomize and controller-gen versions stay the same, which is good. If manual, what distro did you test on? I just want to make sure we don't break the rhel ecosystem distros and only have it work on arch linux after merging this

davidmogar:
Manual. Tested on Fedora with Go 1.18. You should test this as well, but given that is currently passing the CI it should be fine.

johnbieren:
We don't use 1.18 https://github.com/redhat-appstudio/release-service/blob/main/Dockerfile#L2 so I will test it after meetings. I have 1.17 installed

johnbieren (approved):

PR #41, opened 2022-05-27
feat(HACBS-542): group pipeline information

davidmogar:
This commit cleans up the ReleaseStrategy by merging related fields like Pipeline, Bundle and Params.


johnbieren (changes requested):
This breaks the m5 demo files

davidmogar:
I can fix them in a different PR.

johnbieren:
> I can fix them in a different PR.

Why not fix them here so there are no commits in the repo with them broken? :slightly_smiling_face:

davidmogar:
> > I can fix them in a different PR.
> 
> Why not fix them here so there are no commits in the repo with them broken? 🙂

I meant a separate PR to merge both at the same time, but yeah, I can do it in this one.

PR #42, opened 2022-05-27
feat(HACBS-541): add serviceAccount field

davidmogar:
This commit allows to specify a service account for the release PipelineRun. We are going to need to pass a service account to be able to clone private repositories (at least).

### DIFF
--- a/tekton/pipeline_run.go
+++ b/tekton/pipeline_run.go
@@ -148,5 +148,15 @@ func (r *ReleasePipelineRun) WithReleaseStrategy(strategy *v1alpha1.ReleaseStrat
 		})
 	}
 
+	r.WithServiceAccount(strategy.Spec.ServiceAccount)
+
+	return r
+}
+
+// WithServiceAccount adds a reference to the service account to be used to gain elevated privileges during the
+// execution of the different Pipeline tasks.
+func (r *ReleasePipelineRun) WithServiceAccount(serviceAccount string) *ReleasePipelineRun {
+	r.Spec.ServiceAccountName = serviceAccount
+
 	return r
 }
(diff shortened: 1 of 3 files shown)


tekton/pipeline_run.go
```diff
@@ -140,5 +140,15 @@ func (r *ReleasePipelineRun) WithReleaseStrategy(strategy *v1alpha1.ReleaseStrat
...
 
+	r.WithServiceAccount(strategy.Spec.ServiceAccount)
+
+	return r
+}
+
+// WithServiceAccount adds a reference to the service account to be used to gain elevated privileges during the
+// execution of the different Pipeline tasks.
+func (r *ReleasePipelineRun) WithServiceAccount(serviceAccount string) *ReleasePipelineRun {
+	r.Spec.ServiceAccountName = serviceAccount
```
johnbieren, line 151: I am undecided if we should just have this line be in place of line 143 or not. I'll think on it

davidmogar: Having it this way will allow us to set a ServiceAccount independently of the ReleaseStrategy. Not like we are going to need it, but it's cleaner.

johnbieren: I don't follow. line 143 just calls line 151. What makes it independent/dependent of the ReleaseStrategy?

davidmogar: You can call WithServiceAccount without needing to call WithReleaseStrategy.

johnbieren: I doubt we do that, but fair

davidmogar:
Same, but given that we are defining a flexible file, let’s keep it flexible ;)

theflockers (approved):
From what I can tell, it looks good. Hope soon I can review ones like this better!

PR #47, opened 2022-06-02
fix(HACBS-596): add missing Application name

davidmogar:
With the recent changes in the ApplicationSnapshot CRD, we are now missing the Application name.


theflockers (commented):
It looks good, although one more question from me: Why we use such generic name as `component`?

davidmogar:
> It looks good, although one more question from me: Why we use such generic name as `component`?
Component references a [HAS Component](https://github.com/redhat-appstudio/application-service/blob/main/api/v1alpha1/component_types.go#L148)

theflockers (approved):

PR #50, opened 2022-06-05
fix(HACBS-605): use camel case for snapshot parameter

davidmogar:
The parameter name for the snapshot was ApplicationSnapshot, when it should be applicationSnapshot.

### DIFF
--- a/tekton/pipeline_run.go
+++ b/tekton/pipeline_run.go
@@ -19,6 +19,7 @@ package tekton
 import (
 	"encoding/json"
 	"fmt"
+	"unicode"
 
 	libhandler "github.com/operator-framework/operator-lib/handler"
 	"github.com/redhat-appstudio/release-service/api/v1alpha1"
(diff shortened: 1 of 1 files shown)


tekton/pipeline_run.go
```diff
@@ -112,7 +113,11 @@ func (r *ReleasePipelineRun) WithApplicationSnapshot(snapshot *v1alpha1.Applicat
 	// add something like a `Complete` function that returns the final object and error.
 	snapshotString, _ := json.Marshal(snapshot.Spec)
 
-	r.WithExtraParam(snapshot.Kind, tektonv1beta1.ArrayOrString{
+	// Get snapshot.Kind runes to make the first letter lowercase
+	snapshotRunes := []rune(snapshot.Kind)
+	snapshotRunes[0] = unicode.ToLower(snapshotRunes[0])
```
parasense, line 118: Just an FYI.
There used to be a function for that, but it was depreciated.
https://pkg.go.dev/strings#Title

parasense: (resolve conversation once you read, this is just informational)

parasense (approved):
LGTM

tekton/pipeline_run.go
davidmogar: Thanks. I saw another library for doing the same, but as it is an small change I went for this alternative.

PR #52, opened 2022-06-05
fix: fix ServiceAccount field comment

davidmogar:
The comment for this field was incorrect.


parasense (approved):
LGTM

PR #54, opened 2022-06-07
revert: pass release to the Release pipeline

johnbieren:
This reverts commit 31f250f3e1e09916c8def8a570768b497817d01b.


davidmogar:
The commit is not good according to the Conventional commits spec. Weird to see it passing the CI. In any case, I think we should follow [what they say here](https://www.conventionalcommits.org/en/v1.0.0/#how-does-conventional-commits-handle-revert-commits).
```
revert: pass release to the Release pipeline

Refs: 31f250f
```

johnbieren:
> The commit is not good according to the Conventional commits spec. Weird to see it passing the CI. In any case, I think we should follow [what they say here](https://www.conventionalcommits.org/en/v1.0.0/#how-does-conventional-commits-handle-revert-commits).
> 
> ```
> revert: pass release to the Release pipeline
> 
> Refs: 31f250f
> Signed-off-by: Johnny Bieren <jbieren@redhat.com>
> ```
> 
> I changed the PR title already.

Yeah, I was wondering about how it would work with conventional-commits. I was surprised to see it pass the check. Nonetheless, I will change the commit to be like what you pasted after retro

davidmogar (approved):

PR #58, opened 2022-06-13
feat(HACBS-665): switch to using GitOps ApplicationSnapshot CRD

johnbieren:
This repository was temporarily maintaining an ApplicationSnapshot CRD
for testing purposes while the production one was developed by the
GitOps team. The production one is now ready to use, so this commit
removes the temporary one and switches over to using the production one.


controllers/release/release_adapter.go
```diff
@@ -22,6 +22,7 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
```
johnbieren, line 25: @kasemAlem FYI this is how the new import looks

controllers/release/release_adapter.go
```diff
@@ -169,7 +170,7 @@ func (a *Adapter) EnsureReleasePipelineStatusIsTracked() (results.OperationResul
 // annotations, so it triggers Release reconciles whenever it changes. The Pipeline information and the parameters to it
 // will be extracted from the given ReleaseStrategy. The Release's ApplicationSnapshot will also be passed to the
 // release PipelineRun.
-func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStrategy, applicationSnapshot *v1alpha1.ApplicationSnapshot) (*v1beta1.PipelineRun, error) {
+func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStrategy, applicationSnapshot *appstudioshared.ApplicationSnapshot) (*v1beta1.PipelineRun, error) {
```
johnbieren, line 173: @kasemAlem FYI this is how the snapshot is referenced once this merges

tekton/pipeline_run.go
```diff
@@ -19,13 +19,15 @@ package tekton
 import (
 	"encoding/json"
 	"fmt"
+	"os"
+	"unicode"
+
 	libhandler "github.com/operator-framework/operator-lib/handler"
+	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
```
davidmogar, line 26: Could we use the same name used in the main file? That's it, `gitopsv1alpha1`.

controllers/release/release_adapter.go
kasemAlem: @johnbieren Thanks, this was very helpful 
I've created a new branch to fix test ApplicationSnapshot part to be in GitOps api

tekton/pipeline_run.go
davidmogar: I'm ok with you using `appstudioshared` as the GitOps service have API definitions in at least 3 different places, but please add the API version to the name (`appstudiosharedv1alpha1) so it's clear what version are we using.

This PR is blocking progress in e2e, so let's get it changed and merged as soon as possible.

johnbieren: It is written exactly how the gitops team says to write it https://github.com/redhat-appstudio/managed-gitops/tree/main/appstudio-shared#remember-to-add-the-scheme-to-your-operators-maingo 
There is no reason to not follow their documentation

davidmogar: My concerns are that the section in the README is only an example and not using a versioned import will only make more difficult to follow code specially during reviews. It will also require changes in the case of using multiple versions. In addition it breaks with the strategy followed up to this point in which every single CRD import is using a versioned name.

Having said that, let's move on to unblock progress on e2e and reevaluate in the future.

davidmogar (approved):

PR #60, opened 2022-06-16
feat(HACBS-555): add auto-release label to ReleaseLinks

johnbieren:
This commit adds a mutating webhook for ReleaseLink CRs. The webhook
checks to see if the release.appstudio.openshift.io/auto-release label
exists. If it does not exist, it adds it with a value of true. If it
does exist (meaning the user added the label when creating the CR),
it will do nothing.

### DIFF
--- a/api/v1alpha1/releaselink_webhook.go
+++ b/api/v1alpha1/releaselink_webhook.go
@@ -18,6 +18,7 @@ package v1alpha1
 
 import (
 	"fmt"
+
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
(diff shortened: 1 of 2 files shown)


api/v1alpha1/releaselink_webhook.go
```diff
@@ -29,12 +30,19 @@ func (r *ReleaseLink) SetupWebhookWithManager(mgr ctrl.Manager) error {
 		Complete()
 }
 
-// +kubebuilder:webhook:path=/mutate-appstudio-redhat-com-v1alpha1-releaselink,mutating=true,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releaselinks,verbs=create;update,versions=v1alpha1,name=mreleaselink.kb.io,admissionReviewVersions=v1
+// +kubebuilder:webhook:path=/mutate-appstudio-redhat-com-v1alpha1-releaselink,mutating=true,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releaselinks,verbs=create,versions=v1alpha1,name=mreleaselink.kb.io,admissionReviewVersions=v1
```
scoheb, line 33: for my understanding....why was "update" removed?

johnbieren: `update` being there is the default, but we had this webhook turned off before (as you can see by the function being empty). To my best knowledge, our desired behavior is to add a label if it doesn't exist when a ReleaseLink is created. I don't believe we have any desired behavior on ReleaseLink updates, right?

It would have made more sense if this PR introduced this entire line (containing just `create`) but for whatever reason we already had the code in our operator, but we didn't have it doing anything

scoheb: Thinking about it more in detail...

I guess someone could create a ReleaseLink and our webhook would add the label.
then he/she could edit it and remove the label.
We'd want to catch that too, no?

johnbieren: You tell me. My understanding was that integration controller should create a Release if the label exists and == true OR doesn't exist. So the only way to be set to `manual` mode is to create your ReleaseLink with the label set to false, or set it to false at some point after it is created before the integration controller runs. Do we want to add it set it true if the label is removed entirely? This would eliminate the "doesn't exist" scenario from my second sentence here.

@davidmogar thoughts?

api/v1alpha1/releaselink_webhook.go
```diff
@@ -29,12 +30,19 @@ func (r *ReleaseLink) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
 
 var _ webhook.Defaulter = &ReleaseLink{}
 
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
 func (r *ReleaseLink) Default() {
+	label := "release.appstudio.openshift.io/auto-release"
```
davidmogar, line 39: I don't like this hard-coding here. We should revamp it moving `release.appstudio.openshift.io` labels to a single place. Right now we have something like that in the tekton package but that doesn't look like the right place after this change. Having something like this in, for example, the Release type as exported constants, would reduce the hard-coding in the integration service as well, which would use our constant directly.

Having said this, I don't want to address this at this time. I'll create a ticket for it.

davidmogar (approved):

PR #62, opened 2022-06-16
test(HACBS-668): add the first set of Unit Tests for `PipelineRun`s

theflockers:
This commit adds the first set of Unit Tests for `PipelineRun`s

short note: This is a work in progress - just adding the PR for earlier review as it is 1st one.

### DIFF
--- a/tekton/pipeline_run_test.go
+++ b/tekton/pipeline_run_test.go
@@ -24,6 +24,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	"github.com/redhat-appstudio/release-service/api/v1alpha1"
+
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -39,30 +40,26 @@ type ReleasePipelineParams struct {
 	Extra     ExtraParams
 }
 
+const (
+	PipelineRunPrefixName = "mypipeline"
+	Namespace             = "default"
+)
+
 var _ = Describe("PipelineRun", func() {
 
-	var params *ReleasePipelineParams
-	var releasePipelineRun *ReleasePipelineRun
-
 	var release *v1alpha1.Release
-	var releaseParams *v1alpha1.ReleaseSpec
+	var extraParams *ExtraParams
+	var releasePipelineRun *ReleasePipelineRun
 
 	BeforeEach(func() {
 
-		params = &ReleasePipelineParams{
-			Prefix:    "mypipeline",
-			Namespace: "default",
-			// @todo: use a valid Pipeline params
-			Extra: ExtraParams{
-				Name:  "ExtraParam",
-				Value: tektonv1beta1.ArrayOrString{Type: tektonv1beta1.ParamTypeString, StringVal: "ExtraValue"},
+		extraParams = &ExtraParams{
+			Name: "extraConfigPath",
+			Value: tektonv1beta1.ArrayOrString{
+				Type:      tektonv1beta1.ParamTypeString,
+				StringVal: "path/to/extra/config.yaml",
 			},
 		}
-		releaseParams = &v1alpha1.ReleaseSpec{
-			ApplicationSnapshot: "mysnapshot",
-			ReleaseLink:         "myreleaselink",
-		}
-
 		release = &v1alpha1.Release{
 			TypeMeta: metav1.TypeMeta{
 				APIVersion: "appstudio.redhat.com/v1alpha1",
@@ -70,10 +67,12 @@ var _ = Describe("PipelineRun", func() {
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				GenerateName: "myrelease-",
-				Namespace:    params.Namespace,
-				UID:          "8ba3294a-2e1b-4a13-9b3a-faf3cfe3d31b", // This is an arbitrary random UUID
+				Namespace:    Namespace,
 			},
-			Spec: *releaseParams,
+			Spec: v1alpha1.ReleaseSpec{
+				ApplicationSnapshot: "mysnapshot",
+				ReleaseLink:         "myreleaselink",
+			},
 		}
 
 		ctx := context.Background()
@@ -83,16 +82,17 @@ var _ = Describe("PipelineRun", func() {
 		gvk := v1alpha1.GroupVersion.WithKind(kind)
 		controllerRef := metav1.NewControllerRef(release, gvk)
 
-		// creating a release
+		// Creating a release
 		Expect(k8sClient.Create(ctx, release)).Should(Succeed())
 		release.SetOwnerReferences([]metav1.OwnerReference{*controllerRef})
 
-		// need to set the Kind as it loses it for reasons...
+		// Need to set the Kind and APIVersion as it loses it due to:
+		// https://github.com/kubernetes-sigs/controller-runtime/issues/1870
 		release.TypeMeta.APIVersion = "appstudio.redhat.com/v1alpha1"
 		release.TypeMeta.Kind = "Release"
 
 		// Creates the PipelineRun Object
-		releasePipelineRun = NewReleasePipelineRun(params.Prefix, params.Namespace)
+		releasePipelineRun = NewReleasePipelineRun(PipelineRunPrefixName, Namespace)
 		Expect(k8sClient.Create(ctx, releasePipelineRun.AsPipelineRun())).Should(Succeed())
 
 	})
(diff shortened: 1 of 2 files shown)


tekton/tekton_suite_test.go
```diff
@@ -0,0 +1,93 @@
+package tekton
```
theflockers, line 1: Noticed the missing header.

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	var release *v1alpha1.Release
+	var releaseParams *v1alpha1.ReleaseSpec
+
+	BeforeEach(func() {
+
+		params = &ReleasePipelineParams{
+			Prefix:    "mypipeline",
+			Namespace: "default",
+			// @todo: use a valid Pipeline params
```
johnbieren, line 55: Here are some examples. Any of them should work fine here
```
    - name: extraConfigGitUrl
      value: https://github.com/davidmogar/strategy-configs.git
    - name: extraConfigPath
      value: m5.yaml
    - name: extraConfigRevision # optional field, default is main
      value: main
```
if you pick the path one, please name it something besides mX.yaml (we are trying to avoid tying our names to milestones so we don't have to change stuff every milestone).

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+		release = &v1alpha1.Release{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "appstudio.redhat.com/v1alpha1",
+				Kind:       "Release",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "myrelease-",
+				Namespace:    params.Namespace,
+				UID:          "8ba3294a-2e1b-4a13-9b3a-faf3cfe3d31b", // This is an arbitrary random UUID
```
johnbieren, line 74: This line doesn't seem needed

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,137 @@
...
+	var params *ReleasePipelineParams
+	var releasePipelineRun *ReleasePipelineRun
+
+	var release *v1alpha1.Release
+	var releaseParams *v1alpha1.ReleaseSpec
+
+	BeforeEach(func() {
+
+		params = &ReleasePipelineParams{
+			Prefix:    "mypipeline",
```
johnbieren, line 53: This is only used once in the file. I don't necessarily dislike it as a variable at the top, but I think we should be consistent. There are plenty of other strings also used once that aren't variables in the params

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,137 @@
...
+		Expect(k8sClient.Create(ctx, releasePipelineRun.AsPipelineRun())).Should(Succeed())
+
+	})
+
+	AfterEach(func() {
+		_ = k8sClient.Delete(ctx, release)
+		_ = k8sClient.Delete(ctx, releasePipelineRun.AsPipelineRun())
+	})
+
+	Context("Pipeline Run", func() {
```
johnbieren, line 105: Can we improve this string a bit? Maybe something like `Create PipelineRun and modify its attributes`

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+		It("Can create a ReleasePipelineRun", func() {
+			Expect(releasePipelineRun.ObjectMeta.Name).
+				Should(MatchRegexp("mypipeline-" + `[a-z0-9]{5}`))
+			Expect(releasePipelineRun.ObjectMeta.Namespace).To(Equal("default"))
```
johnbieren, line 110: I think this and line 54 should be changed to use a constant or something. If we some day find out we need to switch the namespace or something, it would be nice to have it all in one spot instead of having to grep for it

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,137 @@
...
+		})
+
+		It("Can add extra params to ReleasePipelineRun", func() {
+			releasePipelineRun.WithExtraParam(params.Extra.Name, params.Extra.Value)
+
+			Expect(releasePipelineRun.Spec.Params[0].Name).To(Equal("ExtraParam"))
```
johnbieren, line 116: Same comment as line 110

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+		It("Can add the release Owner annotations to ReleasePipelineRun", func() {
+			releasePipelineRun.WithOwner(release)
+			Expect(releasePipelineRun.Annotations).NotTo(BeNil())
+		})
+
+		It("Can add the release Labels to ReleasePipelineRun", func() {
+			releasePipelineRun.WithReleaseLabels(release.Name, release.Namespace)
+			Expect(releasePipelineRun.Labels["release.appstudio.openshift.io/name"]).
+				Should(MatchRegexp(release.GenerateName + `[a-z1-9]{5}`))
```
johnbieren, line 128: I believe this is cleaner as
`Should(MatchRegexp(release.Name))`

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,137 @@
...
+		// The code below sets the ownership for the Release Object
+		kind := reflect.TypeOf(v1alpha1.Release{}).Name()
+		gvk := v1alpha1.GroupVersion.WithKind(kind)
+		controllerRef := metav1.NewControllerRef(release, gvk)
+
+		// creating a release
+		Expect(k8sClient.Create(ctx, release)).Should(Succeed())
+		release.SetOwnerReferences([]metav1.OwnerReference{*controllerRef})
+
+		// need to set the Kind as it loses it for reasons...
```
johnbieren, line 90: I tested this and you're right. Do you know the reasons? If so, can we write them? If not, totally fine :slightly_smiling_face:  I personally have no clue

tekton/tekton_suite_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"testing"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+
+	// "k8s.io/apimachinery/pkg/runtime"
```
johnbieren, line 32: What is this for? Leftover from testing?

tekton/tekton_suite_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+var (
+	cfg       *rest.Config
+	k8sClient client.Client
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+)
+
+func TestPipelineRun(t *testing.T) {
```
johnbieren, line 54: This file is confusing to me. Where is the test? I assume it is this function, but I don't see it called anywhere. It looks like this just has a before and after with no actual test

theflockers: According to the ginkgo doc: 
```
Next we define a single testing test: func TestBooks(t *testing.T). This is the entry point for Ginkgo - the go test runner will run this function when you run go test or ginkgo.
```
It is implicitly called. I need to read a bit more to understand it properly, but it seems it initiates `t *testing.T` and then all the other functions might use it through its pointer as it is not called anywhere else. It would be much clear if all the other functions were explicitly called inside it.

tekton/pipeline_run_test.go
theflockers: The only "kind of" explanation I got researching it was: "EnvTest behaves differently from an actual cluster". 

https://github.com/kubernetes-sigs/controller-runtime/issues/1870

But I got some others (maybe) more _elegant_ ways to fix it (maybe as not as clear as this one). I will give them a try.

tekton/pipeline_run_test.go
theflockers: Sure. I'll go exactly with this string. I did not have much idea what exactly put there.

tekton/pipeline_run_test.go
theflockers: You are right. This is because I was creating `PipelineRuns` for each test, than later I moved it to top and indeed, this does not need to be a variable anymore.

tekton/pipeline_run_test.go
johnbieren: Up to you. I am fine with you just putting a link to that issue in the comments. If you want to spend more time on it and do the elegant fix, that works too

tekton/tekton_suite_test.go
johnbieren: Ah, I've never used ginkgo, cool. I am not able to review this file as closely, but nothing *looks* bad in it to me

tekton/pipeline_run_test.go
theflockers: I will keep it like this for the moment. Adding it manually is simple and clear to understand. If required in the future I can change it.

tekton/pipeline_run_test.go
```diff
@@ -0,0 +1,136 @@
...
+		// The code below sets the ownership for the Release Object
+		kind := reflect.TypeOf(v1alpha1.Release{}).Name()
+		gvk := v1alpha1.GroupVersion.WithKind(kind)
+		controllerRef := metav1.NewControllerRef(release, gvk)
+
+		// Creating a release
+		Expect(k8sClient.Create(ctx, release)).Should(Succeed())
+		release.SetOwnerReferences([]metav1.OwnerReference{*controllerRef})
+
+		// Need to set the Kind and APIVersion as it loses it due:
```
johnbieren, line 89: `s/due/due to/`

johnbieren (approved):

PR #63, opened 2022-06-16
feat(HACBS-681): force ReleaseLink auto-release to be only true or false

johnbieren:
The release.appstudio.openshift.io/auto-release label for ReleaseLinks can be
manually set or edited once the ReleaseLink is created. This commit forces updates
to the label to only set the value to true or false.

### DIFF
--- a/api/v1alpha1/releaselink_webhook.go
+++ b/api/v1alpha1/releaselink_webhook.go
@@ -24,6 +24,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
+const (
+	// autoReleaseLabel is the label name for the auto-release setting. This will probably be moved elsewhere eventually.
+	autoReleaseLabel = "release.appstudio.openshift.io/auto-release"
+)
+
 func (r *ReleaseLink) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
@@ -36,12 +41,12 @@ var _ webhook.Defaulter = &ReleaseLink{}
 
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
 func (r *ReleaseLink) Default() {
-	label := "release.appstudio.openshift.io/auto-release"
-	if _, found := r.GetLabels()[label]; !found {
+	if _, found := r.GetLabels()[autoReleaseLabel]; !found {
 		if r.Labels == nil {
-			r.Labels = make(map[string]string)
+			r.Labels = map[string]string{
+				autoReleaseLabel: "true",
+			}
 		}
-		r.Labels[label] = "true"
 	}
 }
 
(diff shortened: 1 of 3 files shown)


api/v1alpha1/releaselink_webhook.go
```diff
@@ -36,12 +41,11 @@ var _ webhook.Defaulter = &ReleaseLink{}
...
 // Default implements webhook.Defaulter so a webhook will be registered for the type.
 func (r *ReleaseLink) Default() {
-	label := "release.appstudio.openshift.io/auto-release"
-	if _, found := r.GetLabels()[label]; !found {
+	if _, found := r.GetLabels()[autoReleaseLabel]; !found {
 		if r.Labels == nil {
 			r.Labels = make(map[string]string)
 		}
-		r.Labels[label] = "true"
+		r.Labels[autoReleaseLabel] = "true"
```
parasense, line 48: This looks fine, but I have a trivial question.
Looks like `r.GetLabels()` would potentially not succeed if `r.Labels == nil`.
So I feel like maybe one would want to check if the pointer is `nil` first, then check if the value is empty?
So I'm wondering if this ordering is more optimal, less complex, or does less work? 

Also, I wonder if the `r.Labels = make(map[string]string)` could instead be a [map literal](https://go.dev/ref/spec#Composite_literals) instead?
Something like:
```
r.Labels = map[string]string{autoReleaseLabel : "true"}
```
Note: this is not asking for code changes, feel free to ignore.

api/v1alpha1/releaselink_webhook.go
```diff
@@ -51,12 +55,20 @@ var _ webhook.Validator = &ReleaseLink{}
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type.
 func (r *ReleaseLink) ValidateCreate() error {
-	return r.isTargetValid()
+	err := r.isTargetValid()
+	if err != nil {
+		return err
+	}
```
parasense, line 61: As seen elsewhere in the code, Go allows for a [short statements to be placed ahead of the conditional expression](https://go.dev/ref/spec#If_statements).
Something like:
```
if err := r.isTargetValid() ; err != nil {
    return err
}
```
It may seem trivial, but it shaves off one line, here, another there, and it adds-up.
The only potential gotcha is `err` becomes limited to the scope of `if`'s curly brace, but I dont' see that being a problem here.

Note: this is not asking for code changes, just something I do... feel free to ignore.

davidmogar: Although I agree for this particular case, I actually have more concerns about the auxiliary functions themselves. I missed this when `isTargetValid` was added but the name itself needs an improvement. That name suggests that the function returns a boolean, which is not the case.

Unless there's a better alternative, let's rename the functions to `validateTarget` and `validateAutoReleaseLabel`.

api/v1alpha1/releaselink_webhook.go
```diff
@@ -72,3 +84,13 @@ func (r *ReleaseLink) isTargetValid() error {
 
 	return nil
 }
+
+// isAutoReleaseLabelValid throws an error if the auto-release label value is set to anything besides true and false.
+func (r *ReleaseLink) isAutoReleaseLabelValid() error {
+	if _, found := r.GetLabels()[autoReleaseLabel]; found {
+		if r.Labels[autoReleaseLabel] != "true" && r.Labels[autoReleaseLabel] != "false" {
```
davidmogar, line 91: You are already getting the value in the previous line. Replace `_` with `autoReleaseLabel` so you don't search multiple times.

davidmogar (changes requested):

api/v1alpha1/releaselink_webhook.go
johnbieren: Line 45 already checks for nil, but I did make the change to set the label when the labels are created instead of in another line. Thanks

api/v1alpha1/releaselink_webhook.go
johnbieren: I changed it to be the shorter syntax and renamed the functions

api/v1alpha1/releaselink_webhook.go
johnbieren: Done

davidmogar (approved):

PR #64, opened 2022-06-21
test(HACBS-726): add unit tests to tekton functions in tekton

theflockers:
- adds UnitTests for tekton directory

### DIFF
--- a/tekton/pipeline_run_test.go
+++ b/tekton/pipeline_run_test.go
@@ -18,6 +18,7 @@ package tekton
 
 import (
 	"context"
+	"encoding/json"
 	"reflect"
 
 	. "github.com/onsi/ginkgo"
@@ -25,6 +26,7 @@ import (
 
 	"github.com/redhat-appstudio/release-service/api/v1alpha1"
 
+	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -40,17 +42,23 @@ type ReleasePipelineParams struct {
 	Extra     ExtraParams
 }
 
-const (
-	PipelineRunPrefixName = "mypipeline"
-	Namespace             = "default"
-)
-
 var _ = Describe("PipelineRun", func() {
-
-	var release *v1alpha1.Release
-	var extraParams *ExtraParams
-	var releasePipelineRun *ReleasePipelineRun
-
+	const (
+		pipelineRunPrefixName = "test-pipeline"
+		namespace             = "default"
+		workspace             = "test-workspace"
+		persistentVolumeClaim = "test-pvc"
+		serviceAccountName    = "test-service-account"
+		apiVersion            = "appstudio.redhat.com/v1alpha1"
+	)
+	var (
+		release                            *v1alpha1.Release
+		extraParams                        *ExtraParams
+		releasePipelineRun                 *ReleasePipelineRun
+		snapshot                           *appstudioshared.ApplicationSnapshot
+		strategy                           *v1alpha1.ReleaseStrategy
+		unmarshaledApplicationSnapshotSpec *appstudioshared.ApplicationSnapshotSpec
+	)
 	BeforeEach(func() {
 
 		extraParams = &ExtraParams{
@@ -62,16 +70,40 @@ var _ = Describe("PipelineRun", func() {
 		}
 		release = &v1alpha1.Release{
 			TypeMeta: metav1.TypeMeta{
-				APIVersion: "appstudio.redhat.com/v1alpha1",
+				APIVersion: apiVersion,
 				Kind:       "Release",
 			},
 			ObjectMeta: metav1.ObjectMeta{
 				GenerateName: "myrelease-",
-				Namespace:    Namespace,
+				Namespace:    namespace,
 			},
 			Spec: v1alpha1.ReleaseSpec{
-				ApplicationSnapshot: "mysnapshot",
-				ReleaseLink:         "myreleaselink",
+				ApplicationSnapshot: "testsnapshot",
+				ReleaseLink:         "testreleaselink",
+			},
+		}
+		snapshot = &appstudioshared.ApplicationSnapshot{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "testsnapshot-",
+				Namespace:    "default",
+			},
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: apiVersion,
+				Kind:       "ApplicationSnapshot",
+			},
+			Spec: appstudioshared.ApplicationSnapshotSpec{
+				Application: "testapplication",
+				DisplayName: "Test application",
+				Components:  []appstudioshared.ApplicationSnapshotComponent{},
+			},
+		}
+		strategy = &v1alpha1.ReleaseStrategy{
+			Spec: v1alpha1.ReleaseStrategySpec{
+				Pipeline:              "release-pipeline",
+				Bundle:                "testbundle",
+				Policy:                "testpolicy",
+				PersistentVolumeClaim: persistentVolumeClaim,
+				ServiceAccount:        serviceAccountName,
 			},
 		}
 
@@ -88,13 +120,12 @@ var _ = Describe("PipelineRun", func() {
 
 		// Need to set the Kind and APIVersion as it loses it due to:
 		// https://github.com/kubernetes-sigs/controller-runtime/issues/1870
-		release.TypeMeta.APIVersion = "appstudio.redhat.com/v1alpha1"
+		release.TypeMeta.APIVersion = apiVersion
 		release.TypeMeta.Kind = "Release"
 
 		// Creates the PipelineRun Object
-		releasePipelineRun = NewReleasePipelineRun(PipelineRunPrefixName, Namespace)
+		releasePipelineRun = NewReleasePipelineRun(pipelineRunPrefixName, namespace)
 		Expect(k8sClient.Create(ctx, releasePipelineRun.AsPipelineRun())).Should(Succeed())
-
 	})
 
 	AfterEach(func() {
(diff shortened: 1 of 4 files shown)


tekton/pipeline_run_test.go
```diff
@@ -41,15 +43,22 @@ type ReleasePipelineParams struct {
 }
 
 const (
-	PipelineRunPrefixName = "mypipeline"
+	PipelineRunPrefixName = "test-pipeline"
```
davidmogar, line 46: You probably don't need to export these variables. These could cause you problems if other tests under the same package.

tekton/pipeline_run_test.go
```diff
@@ -132,5 +164,55 @@ var _ = Describe("PipelineRun", func() {
...
+			Expect(json.Unmarshal(
+				[]byte(releasePipelineRun.Spec.Params[0].Value.StringVal),
+				&unmarshaledApplicationSnapshotSpec)).Should(Succeed())
+
+			// check if the unmarshaled data has what we expect
+			Expect(unmarshaledApplicationSnapshotSpec.Application).To(Equal("testapplication"))
+		})
+
+		It("Can add the ReleaseStrategy to the PipelineRun", func() {
+			strategy.Spec.Pipeline = releasePipelineRun.AsPipelineRun().Name
```
davidmogar, line 185: You could set it to any random value like `release-pipeline` when you declare it and skip this line.

tekton/pipeline_run_test.go
```diff
@@ -132,5 +164,55 @@ var _ = Describe("PipelineRun", func() {
...
+				To(Equal(persistentVolumeClaim))
+		})
+
+		It("Is it a PipelineRun", func() {
+			Expect(isReleasePipelineRun(releasePipelineRun.
+				WithReleaseLabels(release.Name, release.Namespace).
+				AsPipelineRun())).To(Equal(true))
+		})
+
+		It("Can make sure if PipelineRun has has Succeeded", func() {
```
davidmogar, line 209: Do we need this? This looks like a test I would expect to see in Tekton to be honest.

tekton/predicates_test.go
```diff
@@ -0,0 +1,122 @@
...
+		releasePipelineRun = NewReleasePipelineRun(pipelineRunPrefixName, namespace)
+		Expect(k8sClient.Create(ctx, releasePipelineRun.AsPipelineRun())).Should(Succeed())
+	})
+
+	AfterEach(func() {
+		_ = k8sClient.Delete(ctx, release)
+		_ = k8sClient.Delete(ctx, releasePipelineRun.AsPipelineRun())
+	})
+
+	Context("Testing ReleasePipelineRunSucceededPredicate", func() {
```
davidmogar, line 85: I think that the descriptions in the tests should be a bit more verbose. For example, in this one I'd write something like:
```suggestion
	Context("when using the ReleasePipelineRunSucceededPredicate predicate", func() {
```
And then the tests would be something like:
```
It("should ignore creation events")
It("should ignore deletion events")
It("should ignore generic events")
It("returns true when an update event is received for a succeeded release PipelineRun")
```

tekton/pipeline_run_test.go
theflockers: The result of `hasPipelineSucceeded` is important for `setupControllerWithManager`. Check https://github.com/redhat-appstudio/release-service/blob/main/controllers/release/release_controller.go#L142

I assume it should behave properly.

tekton/predicates_test.go
theflockers: Ack'ing here for all the comments with the same context. I will changes these labels.

tekton/pipeline_run_test.go
```diff
@@ -106,14 +137,15 @@ var _ = Describe("PipelineRun", func() {
 
 		It("Can create a ReleasePipelineRun", func() {
 			Expect(releasePipelineRun.ObjectMeta.Name).
-				Should(MatchRegexp(PipelineRunPrefixName + `-[a-z0-9]{5}`))
-			Expect(releasePipelineRun.ObjectMeta.Namespace).To(Equal(Namespace))
+				Should(MatchRegexp(pipelineRunPrefixName + `-[a-z0-9]{5}`))
```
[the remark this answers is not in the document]
theflockers: I agree. Although I think once we stick with a format we should make sure it is consistent. For instance, here it says `[a-z0-9]{5}`. If we change this to an `uuid` or a `md5` for instance, we should make sure that it won't accept something different than that.

tekton/pipeline_run_test.go
davidmogar: The actual use case it to use the name from the strategy so let's stick to it. The tests it's ok though. It's just the value used.

tekton/pipeline_run_test.go
davidmogar: Understood, but in that case the this test should be in utils_test.go as it applies to any PipelineRun.

tekton/pipeline_run_test.go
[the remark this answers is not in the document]
theflockers: TLDR: I will change it

Long version: 

The objects that were created by `EnvTest` were always in the format `GenerateName-[a-z0-9]{5}`. What I meant was:  if the suffix format (I don't know from where it comes) changes from `[a-z0-9]{5}` to something else as an `uuid` or a `md5` hash (because 5 chars in the suffix seems too easy to clash) we *eventually* would need to make sure the generated names match the pattern.

That said, it is indeed irrelevant as I the suffix is generated by k8s itself. 

I started thinking that we were managing this somehow as you mentioned "_revisit because the name generation rules change in the future_". So I guessed we were doing atm more than just using the `GenerateName`.

tekton/pipeline_run_test.go
theflockers: Fair enough! I thought I had moved it already tbh! Moving it!

theflockers: wow, just noticed it is a dup! It is already in `utils_test.go`.

tekton/pipeline_run_test.go
```diff
@@ -62,16 +70,40 @@ var _ = Describe("PipelineRun", func() {
 		}
 		release = &v1alpha1.Release{
```
davidmogar, line 71: I thought I commented on this but probably I forgot. Although outside of this patch, it seems that the release doesn't change per test. You could have a `BeforeAll` for it.

tekton/pipeline_run_test.go
```diff
@@ -88,49 +120,84 @@ var _ = Describe("PipelineRun", func() {
...
 			releasePipelineRun.WithReleaseLabels(release.Name, release.Namespace)
 			Expect(releasePipelineRun.Labels["release.appstudio.openshift.io/name"]).
 				To(Equal(release.Name))
 		})
 
-		It("Can cast ReleasePipelineRun type to PipelineRun", func() {
-			Expect(reflect.TypeOf(releasePipelineRun.AsPipelineRun())).
-				To(Equal(reflect.TypeOf(&tektonv1beta1.PipelineRun{})))
+		It("can return a PipelineRun object from a ReleasePipelineRun object",
+			func() {
```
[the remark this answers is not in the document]
theflockers: Leftover. I was testing breaking string lines in go, and in the end forgot to move this on back.

tekton/pipeline_run_test.go
```diff
@@ -62,16 +70,40 @@ var _ = Describe("PipelineRun", func() {
...
+			},
+			Spec: appstudioshared.ApplicationSnapshotSpec{
+				Application: "testapplication",
+				DisplayName: "Test application",
+				Components:  []appstudioshared.ApplicationSnapshotComponent{},
+			},
+		}
+		strategy = &v1alpha1.ReleaseStrategy{
+			Spec: v1alpha1.ReleaseStrategySpec{
+				Pipeline:              "", // added while running the test
```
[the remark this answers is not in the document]
theflockers: I was actually setting it afterwards with a value of the PipelineRun. However I have changed this to a fixed name as suggested in another comment.

tekton/pipeline_run_test.go
```diff
@@ -88,49 +120,84 @@ var _ = Describe("PipelineRun", func() {
...
 		})
 
-		It("Can add the release Owner annotations to ReleasePipelineRun", func() {
+		It("can append owner release information to the object as annotations", func() {
 			releasePipelineRun.WithOwner(release)
 			Expect(releasePipelineRun.Annotations).NotTo(BeNil())
 		})
 
-		It("Can add the release Labels to ReleasePipelineRun", func() {
+		It("can append the release Name and Namespace to a ReleasePipelineRun object and that these label key names matches the correct label format", func() {
```
johnbieren, line 156: nit: `s/matches/match/`

tekton/predicates_test.go
```diff
@@ -0,0 +1,122 @@
...
+		})
+
+		It("should ignore generic events", func() {
+			contextEvent := event.GenericEvent{
+				Object: releasePipelineRun.AsPipelineRun(),
+			}
+			Expect(instance.Generic(contextEvent)).To(BeFalse())
+		})
+
+		It("should return true when an updated even is received for a succeeded release PipelineRun", func() {
```
johnbieren, line 109: `s/even/event/`

tekton/utils_test.go
```diff
@@ -0,0 +1,100 @@
...
+	const (
+		pipelineRunPrefixName = "test-pipeline"
+		apiVersion            = "appstudio.redhat.com/v1alpha1"
+		namespace             = "default"
+	)
+
+	var release *v1alpha1.Release
+	var releasePipelineRun *ReleasePipelineRun
+
+	BeforeEach(func() {
```
johnbieren, line 42: This is a question, not suggestion - there is no way to make this public to the package and not have to redefine it in each test? It is the same as the one in `predicates_test.go`

johnbieren (changes requested):

tekton/utils_test.go
theflockers: This was sort of the same question David made to me but in the opposite direction :D. He was concerning that some constants I had defined globally would eventually clash with other parts of the `tekton` package, therefore I moved it into the function.

johnbieren: @davidmogar thoughts?

davidmogar: Sharing variables in code is ok but not a fan of sharing them in tests. Nevertheless I don't have an strong opinion on this. Your call Leandro.

theflockers: Thanks for the feedback @davidmogar and @johnbieren. I would prefer keep it like this for now. If it become necessary (as tests tends to increase) I might refactor it.

PR #66, opened 2022-06-27
feat(HACBS-769): use gosec in release-service ci 

parasense:
Please see upstream docs here:
https://github.com/securego/gosec/

Jira: HACBS-769

### DIFF
--- a/.github/workflows/gosec.yaml
+++ b/.github/workflows/gosec.yaml
@@ -1,0 +1,27 @@
+---
+name: Go Test on Pull Requests
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+    paths:
+      - '**.go'
+  workflow_dispatch:
+jobs:
+  gosec:
+    name: Check GO security
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Run Gosec Security Scanner
+        uses: securego/gosec@master
+        with:
+          args: -exclude-generated ./...
+        env:
+          GOROOT: ""


davidmogar:
I've been taking a look to every rule applied by gosec and to be honest, most of it doesn't/won't apply to us. We get just a small benefit using this. It doesn't hurt though...

parasense:
> I've been taking a look to every rule applied by gosec and to be honest, most of it doesn't/won't apply to us. We get just a small benefit using this. It doesn't hurt though...

For reference non the list of rules:
https://github.com/securego/gosec/#available-rules

@davidmogar We could specify the rules we want to run, for example on the cmdline `gosec -include=G101,G203,G401 ...` , and for the GitHub action we would pass the above as an `arg` parameter. I think we could skip the SQL rules (G201, G202).

.github/workflows/pr.yml
```diff
@@ -84,6 +84,12 @@ jobs:
             git --no-pager diff ':(exclude)bin/kustomize' ':(exclude)bin/controller-gen'
             exit 1
           fi
+      - name: Run Gosec Security Scanner
+        uses: securego/gosec@master
+        with:
+          args: ./...
+        env:
+          GOROOT: ""
```
davidmogar, line 92: For my understanding. Why you need to change the GOROOT?

.github/workflows/pr.yml
```diff
@@ -84,6 +84,12 @@ jobs:
             git --no-pager diff ':(exclude)bin/kustomize' ':(exclude)bin/controller-gen'
             exit 1
           fi
+      - name: Run Gosec Security Scanner
+        uses: securego/gosec@master
+        with:
+          args: ./...
```
davidmogar, line 90: Do we want to scan tests files? We have some hits there.

.github/workflows/pr.yml
parasense: Using gosec in GitHub action seems to need GOROOT unset.
See here for specific details:
https://github.com/securego/gosec/issues/469

.github/workflows/pr.yml
parasense: I would default to say yes, test everything, but honestly I have no strong feeling either way.
What is your preference @davidmogar ?

davidmogar: I'd say yes skipping generated files.

davidmogar:
We've been discussing lately about cleaning up the CI file. This is a good candidate in my opinion. The new stuff in this PR can be in a new workflow (gosec.yaml). Can you make the change in this PR?

parasense:
> We've been discussing lately about cleaning up the CI file. This is a good candidate in my opinion. The new stuff in this PR can be in a new workflow (gosec.yaml). Can you make the change in this PR?

I think that's a great idea, however I think it might be more appropriate to split the workflow in a separate Jira/PR. Right now I don't see any existing progress towards that goal, so I'm not comfortable doing that in this PR. 

That said I'd be happy to work on splitting the workflow... I'll setup a Jira to capture the work.

.github/workflows/pr.yml
parasense: Done. I found an option to `gosec` to ignore generated files.
```
$ ~/go/bin/gosec -exclude-generated ./...
...
[gosec] 2022/08/16 12:07:48 Checking file: /home/jdisnard/scm/github/forks/release-service/api/v1alpha1/releasestrategy_types.go
[gosec] 2022/08/16 12:07:48 Ignoring generated file: /home/jdisnard/scm/github/forks/release-service/api/v1alpha1/zz_generated.deepcopy.go
[gosec] 2022/08/16 12:07:48 Checking package: tekton
...
```
It works as expected, but one note is the option is not well documented:
https://github.com/securego/gosec/issues/709

parasense:
After some more consideration, I decided refactored this as a separate CI workflow, per one of the suggestions.
One big reason was to only trigger the Gosec scan when GO code was updated, using path triggers.

.github/workflows/gosec.yaml
```diff
@@ -0,0 +1,27 @@
...
+      - reopened
+    paths:
+      - '**.go'
+  workflow_dispatch:
+jobs:
+  gosec:
+    name: Check GO security
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check out code
```
johnbieren, line 17: One last question then I am good to approve. This seems like the same as https://github.com/redhat-appstudio/release-service/blob/main/.github/workflows/pr.yml#L19
However we are using different fetch-depths and this one specifies the ref (maybe this is the default since the linked one seems to work just fine?) Is there a reason fetch-depth:1 is better than 0? I just think consistency would be ideal here since as I understand it, it is the same action to do the same purpose

parasense: It's an optimization for speed. 
I don't think `gosec` needs to have the entire git history, just the most recent HEAD of the given branch in the PR.

johnbieren: gotcha. makes sense to me then. maybe we should change the `pr.yaml` depth at some point but that is out of scope of this work IMO

johnbieren (approved):

davidmogar (approved):

PR #67, opened 2022-06-27
feat(HACBS-694): generate APIResourceSchema

johnbieren:
This commit adds the APIResourceSchema for the release service. This
is needed for KCP integration. The generate-kcp-api.sh script is copied
from the has and spi repos.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -41,6 +41,11 @@ jobs:
         run: |
           rm -f bin/kustomize bin/controller-gen
           make kustomize controller-gen
+      - name: Install kcp tooling
+        run: |
+          git clone https://github.com/kcp-dev/kcp ~/kcp
+          cd ~/kcp
+          WHAT=./cmd/kubectl-kcp make install
       - name: Cache go modules
         id: cache-mod
         uses: actions/cache@v2
@@ -75,12 +80,21 @@ jobs:
         with:
           version: "2021.1.2"
           install-go: false
-      - name: Check manifests
+      - name: Check kcp manifests
         run: |
-          make generate manifests
+          CONTROL_PLANE=kcp make generate manifests
           if [[ ! -z $(git status ':(exclude)bin/kustomize' ':(exclude)bin/controller-gen' -s) ]]
           then
-            echo "generated sources are not up to date:"
+            echo "generated kcp sources are not up to date:"
+            git --no-pager diff ':(exclude)bin/kustomize' ':(exclude)bin/controller-gen'
+            exit 1
+          fi
+      - name: Check kubernetes manifests
+        run: |
+          CONTROL_PLANE=kubernetes make generate manifests
+          if [[ ! -z $(git status ':(exclude)bin/kustomize' ':(exclude)bin/controller-gen' -s) ]]
+          then
+            echo "generated k8s sources are not up to date:"
             git --no-pager diff ':(exclude)bin/kustomize' ':(exclude)bin/controller-gen'
             exit 1
           fi
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,9 @@ TAG_NAME ?= next
 CERT_MANAGER_VERSION ?= v1.8.0
 ENABLE_WEBHOOKS ?= true
 
+# CONTROL_PLANE defines the type of cluster that will be used. Possible values are kubernetes (default) and kcp.
+CONTROL_PLANE ?= kubernetes
+
 # DEFAULT_PERSISTENT_VOLUME_CLAIM defines the default PVC to be used in the Release pipeline workspace declaration.
 DEFAULT_RELEASE_PVC ?= release-pvc
 
@@ -98,6 +101,9 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+ifeq ($(CONTROL_PLANE), kcp)
+	hack/generate-kcp-api.sh ## Generate KCP APIResourceSchema from CRDs
+endif
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
--- a/config/kcp/kustomization.yaml
+++ b/config/kcp/kustomization.yaml
@@ -1,0 +1,3 @@
+resources:
+  - apiexport_release.yaml
+  - apiresourceschema_release.yaml
--- a/hack/generate-kcp-api.sh
+++ b/hack/generate-kcp-api.sh
@@ -1,0 +1,68 @@
+#!/bin/sh
+
+set -e
+
+if ! which kubectl-kcp; then
+  echo "kubectl-kcp required on path"
+  echo "you can install it with running:"
+  echo "    $ git clone https://github.com/kcp-dev/kcp && cd kcp && make install"
+  exit 1
+fi
+
+THIS_DIR="$(dirname "$(realpath "$0")")"
+CRD_DIR="$( realpath ${THIS_DIR}/../config/crd/bases)"
+KCP_API_DIR="$( realpath ${THIS_DIR}/../config/kcp)"
+
+KCP_API_SCHEMA_FILE_CURRENT="${KCP_API_DIR}/apiresourceschema_release.yaml"
+KCP_API_SCHEMA_FILE_NEW="${KCP_API_DIR}/apiresourceschema_release.yaml_new"
+cat << EOF > ${KCP_API_SCHEMA_FILE_NEW}
+# This file is generated from CRDs by ./hack/generate-kcp-api.sh script.
+# Please do not modify!
+
+EOF
+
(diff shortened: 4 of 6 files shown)


scoheb (commented):
Two of the files are named _has
Is this intentional?

johnbieren:
> Two of the files are named _has Is this intentional?

Nope, I thought I fixed that. Thanks for pointing that out

hack/generate-kcp-api.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/sh
+
+set -e
+
+if ! which kubectl-kcp; then
+  echo "kubectl-kcp required on path"
+  echo "you can install it with running:"
```
davidmogar, line 7:
```suggestion
  echo "you can install it running:"
```

hack/generate-kcp-api.sh
```diff
@@ -0,0 +1,68 @@
...
+  echo "    $ git clone https://github.com/kcp-dev/kcp && cd kcp && make install"
+  exit 1
+fi
+
+THIS_DIR="$(dirname "$(realpath "$0")")"
+CRD_DIR="$( realpath ${THIS_DIR}/../config/crd/bases)"
```
davidmogar, line 13:
```suggestion
CRD_DIR="$(realpath ${THIS_DIR}/../config/crd/bases)"
```

hack/generate-kcp-api.sh
```diff
@@ -0,0 +1,68 @@
...
+KCP_API_DIR="$( realpath ${THIS_DIR}/../config/kcp)"
```
davidmogar, line 14:
```suggestion
KCP_API_DIR="$(realpath ${THIS_DIR}/../config/kcp)"
```

hack/generate-kcp-api.sh
```diff
@@ -0,0 +1,68 @@
...
+
+EOF
+
+# APIResourceSchema is immutable so when we want to update something, we actually have to create new version.
+# Version is defined by this prefix, which is taken from date. This will allow us to do new version each minute, which
+# should be hopefully enough granularity :)
+PREFIX=$( TZ="Etc/UTC" date +%Y%m%d%H%M )
+
+I=0
+for CRD in $( ls ${CRD_DIR} ); do
```
davidmogar, line 30:
```suggestion
for CRD in $(ls ${CRD_DIR}); do
```

hack/generate-kcp-api.sh
```diff
@@ -0,0 +1,68 @@
...
+cat << EOF > ${KCP_API_SCHEMA_FILE_NEW}
+# This file is generated from CRDs by ./hack/generate-kcp-api.sh script.
+# Please do not modify!
+
+EOF
+
+# APIResourceSchema is immutable so when we want to update something, we actually have to create new version.
+# Version is defined by this prefix, which is taken from date. This will allow us to do new version each minute, which
+# should be hopefully enough granularity :)
+PREFIX=$( TZ="Etc/UTC" date +%Y%m%d%H%M )
```
davidmogar, line 27: As an alternative, you could use the output of md5sum for each CRD.

Although the date sounds good, this could become confusing really easily when two PRs, both modifying the CRDs and the APIResourceSchema, go to main. Depending on when the code was pushed (commit date) and the order followed to merge the PRs, the actual date could mean nothing (last PR merged could have a date previous to the other PR date and this is not fixed by a rebase). At least I think this would be the case... What do you think?

hack/generate-kcp-api.sh
```diff
@@ -0,0 +1,68 @@
...
+spec:
+  latestResourceSchemas:
+EOF
+
+# match APIResourceSchema name pattern with date prefix
+for SCHEMA in $( cat ${KCP_API_SCHEMA_FILE_CURRENT} | grep -o "v[0-9]\{12\}\..*\.appstudio\.redhat\.com" ); do
+  echo "    - ${SCHEMA}" >> ${KCP_API_EXPORT_FILE}
+done
+
+cat << EOF > ${KCP_API_EXPORT_FILE}
```
davidmogar, line 63: Why not to add this directly in line 50?

hack/generate-kcp-api.sh
```diff
@@ -0,0 +1,68 @@
...
+# If there are some changes in new generated file, we replace old one. Otherwise just remove new file.
+# The regex is there to ignore name change, because we're updating date there so it is expected to change.
+# Ignored line looks like this:
+# '  name: v202206151654.applications.appstudio.redhat.com'
+if ! diff -I '^  name: v[0-9]\{12\}\..*\.appstudio\.redhat\.com$' ${KCP_API_SCHEMA_FILE_CURRENT} ${KCP_API_SCHEMA_FILE_NEW} > /dev/null; then
+  mv ${KCP_API_SCHEMA_FILE_NEW} ${KCP_API_SCHEMA_FILE_CURRENT}
+  echo "updated KCP APIResourceSchema for release saved at '${KCP_API_SCHEMA_FILE_CURRENT}'"
+else
+  echo "no changes in KCP API"
+  rm ${KCP_API_SCHEMA_FILE_NEW}
```
davidmogar, line 43: This should be a trap to be sure it's deleted even if the script breaks for some reason. For example, if you run this script without logging in your cluster, the file will be created with the heading, the script will break in line 31 and the file won't be deleted.

hack/generate-kcp-api.sh
johnbieren: This script is directly copied from the spi and has workstreams (just with spi/has replaced with release). I think we should keep it as is so we are like the other teams in hacbs

davidmogar: This looks prone to failure but if everyone else is using the same script I guess consistency wins. To be revisited in case of problems.

config/kcp/kustomization.yaml
```diff
@@ -1,3 +1,44 @@
...
+- ../manager
+- ../kcp-rbac
+
 resources:
-  - apiexport_release.yaml
-  - apiresourceschema_release.yaml
+- apiexport_release.yaml
+- apiresourceschema_release.yaml
+
+patches:
```
johnbieren, line 24: Still looking at the patches

Makefile
```diff
@@ -98,6 +98,11 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(MAKE) manifests-kcp
+
+.PHONY: manifests-kcp
+manifests-kcp: ## Generate KCP APIResourceSchema from CRDs
```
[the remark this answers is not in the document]
davidmogar: I think that CONTROL_PLACE should default to kubernetes and we should keep testing kubernetes given that we don't have anything else yet. Wondering how we will test it once we move completely to kcp given the blockers we discussed... Will envtest still work?

.github/workflows/pr.yml
```diff
@@ -75,9 +80,9 @@ jobs:
         with:
           version: "2021.1.2"
           install-go: false
-      - name: Check manifests
+      - name: Check all manifests
         run: |
-          make generate manifests
+          CONTROL_PLANE=kcp make generate manifests
```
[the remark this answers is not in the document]
davidmogar: This would ignore anything in the `kubernetes` directory. If rbac is going to be different not sure if it will even work. Webhooks will be disabled too.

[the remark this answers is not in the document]
johnbieren: I changed it to two separate checks. wdyt?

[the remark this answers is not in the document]
davidmogar: :+1:

johnbieren:
I am not too sure how to test this with KCP's current state. I will add the deploy and rbac parts in a followup PR so we don't slow down mering the APIResourceSchemas (which the program asks for links to)

Makefile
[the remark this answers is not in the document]
johnbieren: Resolving this as the only outstanding parts of it will be addressed in the next PR

ralphbean:
Reach out to Matous Jobanek if you need help figuring out how to test.

johnbieren:
@davidmogar I tested this based on the instructions I got from Matous, as Ralph suggested. It all applied cleanly and looks to have worked correctly, so I think we can declare this as tested and should be good to merge (assuming you have no pending requested changes for this PR)

.github/workflows/pr.yml
```diff
@@ -75,12 +80,21 @@ jobs:
...
+          if [[ ! -z $(git status ':(exclude)bin/kustomize' ':(exclude)bin/controller-gen' -s) ]]
+          then
+            echo "generated kcp sources are not up to date:"
+            git --no-pager diff ':(exclude)bin/kustomize' ':(exclude)bin/controller-gen'
+            exit 1
+          fi
+      - name: Check kubernetes manifests
         run: |
-          make generate manifests
+          CONTROL_PLANE=kubernetes make generate manifests
```
davidmogar, line 94: You don't really need to specify the control plane here although if you feel strong about keeping it here I'm fine with it.

davidmogar (approved):

.github/workflows/pr.yml
johnbieren: I'd like to keep it. I think it makes it easier to read when contrasting with the kcp check

PR #68, opened 2022-06-28
feat(HACBS-776): submit coverage reports to codecov

davidmogar:
This change configures the CI to push coverage reports to Codecov whenever a PR to main is created or there is a push to the main branch.

A codecov.yml file has been added to ignore those files that should not be part of the report.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -86,6 +86,8 @@ jobs:
           fi
       - name: Run Go Tests
         run: make test
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v2
       - name: Ensure make build succeeds
         run: make build
       - name: Ensure make bundle succeeds
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,21 @@
+name: Code Coverage Report
+on:
+  push:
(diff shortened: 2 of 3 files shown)


johnbieren (commented):
The code itself lgtm. The description about a new report whenever a PR to main is opened does not make sense to me though. The trigger for the action seems to just be pushes to main

davidmogar:
> The code itself lgtm. The description about a new report whenever a PR to main is opened does not make sense to me though. The trigger for the action seems to just be pushes to main

Maybe you missed one of the files? It also adds the action to the pr workflow. Although it's true that this is for a PR to any branch. I'll update that.

.github/workflows/pr.yml
```diff
@@ -86,6 +86,8 @@ jobs:
           fi
       - name: Run Go Tests
         run: make test
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v2
```
johnbieren, line 90: Why is this not `v2.1.0` like in `codecov.yml`?

davidmogar: Because I didn't want to pin the version in any of them but I forgot to remove it in one of them. Fixing it.

PR #69, opened 2022-06-29
fix(HACBS-791): fix isDone function

davidmogar:
Whenever markInvalid is called, the output of isDone is checked to know if we have to send a metric or not. The isDone function was properly checking the value of the condition but was not taking into account the case in which the condition didn't exist.


johnbieren (approved):
`or not` doesn't feel needed in the description to me, but that is a super nit pick

PR #71, opened 2022-07-12
test(HACBS-774): increase coverage in controllers/release/ directory

theflockers:
adds release_controller tests
adds release_adapter tests
bumps up ginkgo to v2

### DIFF
--- a/controllers/release/release_adapter_test.go
+++ b/controllers/release/release_adapter_test.go
@@ -1,0 +1,381 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package release
+
+import (
+	"reflect"
+	"time"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/operator-framework/operator-lib/handler"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
+	appstudiov1alpha1 "github.com/redhat-appstudio/release-service/api/v1alpha1"
+	"github.com/redhat-appstudio/release-service/tekton"
+	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Release Adapter", Ordered, func() {
+	var (
+		adapter *Adapter
+
+		release             *appstudiov1alpha1.Release
+		releaseStrategy     *appstudiov1alpha1.ReleaseStrategy
+		releaseLink         *appstudiov1alpha1.ReleaseLink
+		applicationSnapshot *appstudioshared.ApplicationSnapshot
+	)
+
+	BeforeAll(func() {
+		releaseStrategy = &appstudiov1alpha1.ReleaseStrategy{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "test-releasestrategy-",
+				Namespace:    testNamespace,
+			},
+			Spec: appstudiov1alpha1.ReleaseStrategySpec{
+				Pipeline:              "release-pipeline",
+				Bundle:                "test-bundle",
+				Policy:                "test-policy",
+				PersistentVolumeClaim: "test-pvc",
+				ServiceAccount:        "test-account",
+			},
+		}
+		Expect(k8sClient.Create(ctx, releaseStrategy)).Should(Succeed())
+
+		releaseLink = &appstudiov1alpha1.ReleaseLink{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "test-releaselink-",
+				Namespace:    testNamespace,
+				Labels: map[string]string{
+					"release.appstudio.openshift.io/auto-release": "false",
+				},
+			},
+			Spec: appstudiov1alpha1.ReleaseLinkSpec{
+				Application:     "test-app",
+				Target:          "default",
+				ReleaseStrategy: releaseStrategy.GetName(),
+			},
+		}
+		Expect(k8sClient.Create(ctx, releaseLink)).Should(Succeed())
+
+		applicationSnapshot = &appstudioshared.ApplicationSnapshot{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "test-snapshot-",
+				Namespace:    "default",
+			},
+			Spec: appstudioshared.ApplicationSnapshotSpec{
+				Application: "testapplication",
+				Components:  []appstudioshared.ApplicationSnapshotComponent{},
+			},
+		}
+		Expect(k8sClient.Create(ctx, applicationSnapshot)).Should(Succeed())
+	})
+
+	BeforeEach(func() {
+		release = &appstudiov1alpha1.Release{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "test-release-",
+				Namespace:    testNamespace,
+			},
+			Spec: appstudiov1alpha1.ReleaseSpec{
+				ApplicationSnapshot: applicationSnapshot.GetName(),
+				ReleaseLink:         releaseLink.GetName(),
+			},
+		}
+		Expect(k8sClient.Create(ctx, release)).Should(Succeed())
+
+		Eventually(func() error {
+			err := k8sClient.Get(ctx, types.NamespacedName{
+				Name:      release.Name,
+				Namespace: testNamespace,
+			}, release)
+			return err
+		}).ShouldNot(HaveOccurred())
+
+		adapter = NewAdapter(release, ctrl.Log, k8sClient, ctx)
+		Expect(reflect.TypeOf(adapter)).To(Equal(reflect.TypeOf(&Adapter{})))
+	})
+
+	AfterEach(func() {
+		err := k8sClient.Delete(ctx, release)
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+	})
+
+	AfterAll(func() {
+		err := k8sClient.Delete(ctx, releaseLink)
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+
+		err = k8sClient.Delete(ctx, applicationSnapshot)
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+
+		Expect(k8sClient.Delete(ctx, releaseStrategy)).Should(Succeed())
+	})
+
+	It("can create a new Adapter instance", func() {
+		Expect(reflect.TypeOf(NewAdapter(release, ctrl.Log, k8sClient, ctx))).To(Equal(reflect.TypeOf(&Adapter{})))
+	})
+
+	It("ensures the finalizer is added to the Release object", func() {
+		result, err := adapter.EnsureFinalizerIsAdded()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+		Expect(release.ObjectMeta.GetFinalizers()).To(ContainElement(finalizerName))
+
+		// Call should not fail if the finalizer already exists
+		result, err = adapter.EnsureFinalizerIsAdded()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+	})
+
+	It("ensures a PipelineRun object exists", func() {
+		Eventually(func() bool {
+			result, err := adapter.EnsureReleasePipelineRunExists()
+			return (!result.CancelRequest && err == nil)
+		}, time.Second*10).Should(BeTrue())
+
+		pipelineRun, err := adapter.getReleasePipelineRun()
+		Expect(pipelineRun != nil && err == nil).To(BeTrue())
+		Expect(k8sClient.Delete(ctx, pipelineRun)).Should(Succeed())
+	})
+
+	It("ensures the Release Pipeline status is tracked", func() {
+		result, err := adapter.EnsureReleasePipelineRunExists()
+		Expect(!result.CancelRequest && err == nil).Should(BeTrue())
+
+		pipelineRun, err := adapter.getReleasePipelineRun()
+		Expect(pipelineRun != nil && err == nil).Should(BeTrue())
+
+		result, err = adapter.EnsureReleasePipelineStatusIsTracked()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		// It should return results.ContinueProcessing() in case the release is done
+		release.MarkSucceeded()
+		result, err = adapter.EnsureReleasePipelineStatusIsTracked()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		// The Release is running but has no pipelineRun matching the release
+		release.MarkRunning()
+
+		// avoiding "(SA5011)"
+		if pipelineRun != nil {
+			pipelineRun.ObjectMeta.Labels = make(map[string]string)
+		}
+		Expect(k8sClient.Update(ctx, pipelineRun)).Should(Succeed())
+
+		release.Status.Conditions = []metav1.Condition{}
+		result, err = adapter.EnsureReleasePipelineStatusIsTracked()
+		Expect(err).NotTo(HaveOccurred())
+
+		Eventually(func() bool {
+			pipelineRun, err := adapter.getReleasePipelineRun()
+			return pipelineRun == nil && err == nil
+
+		}, time.Second*10).Should(BeTrue())
+		Expect(k8sClient.Delete(ctx, pipelineRun)).Should(Succeed())
+	})
+
+	It("ensures finalizers are called when a release is deleted", func() {
+		// Call to EnsureFinalizersAreCalled should succeed if the Release was not set to be deleted
+		result, err := adapter.EnsureFinalizersAreCalled()
+		Expect(!result.CancelRequest && !result.RequeueRequest).To(BeTrue())
+		Expect(err).NotTo(HaveOccurred())
+
+		// Add finalizer
+		result, err = adapter.EnsureFinalizerIsAdded()
+		Expect(result.CancelRequest).To(BeFalse())
+		Expect(err).NotTo(HaveOccurred())
+
+		// Create a ReleasePipelineRun to ensure it gets finalized
+		result, err = adapter.EnsureReleasePipelineRunExists()
+		Expect(result.CancelRequest).To(BeFalse())
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(k8sClient.Delete(ctx, release)).NotTo(HaveOccurred())
+		Eventually(func() bool {
+			err = k8sClient.Get(ctx, types.NamespacedName{
+				Name:      release.Name,
+				Namespace: release.Namespace,
+			}, release)
+			return err == nil && release.GetDeletionTimestamp() != nil
+		}, time.Second*10).Should(BeTrue())
+
+		result, err = adapter.EnsureFinalizersAreCalled()
+		Expect(result.RequeueRequest).To(BeTrue())
+		Expect(err).NotTo(HaveOccurred())
+
+		Eventually(func() error {
+			return k8sClient.Get(ctx, types.NamespacedName{
+				Name:      release.Name,
+				Namespace: release.Namespace,
+			}, release)
+		}, time.Second*10).Should(HaveOccurred())
+
+		Eventually(func() bool {
+			pipelineRun, _ := adapter.getReleasePipelineRun()
+			return pipelineRun == nil
+		}, time.Second*10).Should(BeTrue())
+	})
+
+	It("can create a ReleasePipelineRun", func() {
+		applicationSnapshot.TypeMeta.Kind = "applicationSnapshot"
+		Expect(k8sClient.Update(ctx, applicationSnapshot)).Should(Succeed())
+
+		pipelineRun, err := adapter.createReleasePipelineRun(
+			releaseStrategy,
+			applicationSnapshot)
+		Expect(err).ShouldNot(HaveOccurred())
+
+		// AsPipelineRun()
+		Expect(reflect.TypeOf(pipelineRun)).To(Equal(reflect.TypeOf(&tektonv1beta1.PipelineRun{})))
+
+		// WithOwner()
+		Expect(pipelineRun.GetAnnotations()[handler.NamespacedNameAnnotation]).To(ContainSubstring(release.Name))
+		Expect(pipelineRun.GetAnnotations()[handler.TypeAnnotation]).To(ContainSubstring("Release"))
+
+		// WithReleaseLabels()
+		Expect(pipelineRun.GetLabels()[tekton.PipelinesTypeLabel]).To(Equal("release"))
+		Expect(pipelineRun.GetLabels()[tekton.ReleaseNameLabel]).To(Equal(release.Name))
+		Expect(pipelineRun.GetLabels()[tekton.ReleaseWorkspaceLabel]).To(Equal(testNamespace))
+
+		// WithReleaseStrategy()
+		Expect(pipelineRun.Spec.PipelineRef.Name).To(Equal(releaseStrategy.Spec.Pipeline))
+		Expect(pipelineRun.Spec.PipelineRef.Bundle).To(Equal(releaseStrategy.Spec.Bundle))
+		Expect(pipelineRun.Spec.Params[0].Name).To(Equal("policy"))
+		Expect(pipelineRun.Spec.Params[0].Value.StringVal).Should(ContainSubstring(releaseStrategy.Spec.Policy))
+		Expect(pipelineRun.Spec.ServiceAccountName).Should(ContainSubstring("test-account"))
+
+		// WithApplicationSnapshot()
+		Expect(pipelineRun.Spec.Params[1].Name).To(Equal("applicationSnapshot"))
+		Expect(pipelineRun.Spec.Params[1].Value.StringVal).To(ContainSubstring(applicationSnapshot.Spec.Application))
+
+		Expect(k8sClient.Delete(ctx, pipelineRun)).Should(Succeed())
+	})
+
+	It("can return the pipelineRun from the release", func() {
+		result, err := adapter.EnsureReleasePipelineRunExists()
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(result.CancelRequest).To(BeFalse())
+
+		pipelineRun, err := adapter.getReleasePipelineRun()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(pipelineRun)).To(Equal(reflect.TypeOf(&tektonv1beta1.PipelineRun{})))
+	})
+
+	It("can return the releaseStrategy", func() {
+		strategy, _ := adapter.getReleaseStrategyFromRelease()
+		link, _ := adapter.getReleaseLink()
+
+		Expect(reflect.TypeOf(strategy)).To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseStrategy{})))
+		Expect(strategy.Namespace).To(Equal(link.Spec.Target))
+	})
+
+	It("can return the releaseStrategy referenced in a given releaseLink", func() {
+		strategy, err := adapter.getReleaseStrategy(releaseLink)
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(strategy)).To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseStrategy{})))
+	})
+
+	It("can return the target ReleaseLink from a given ReleaseLink", func() {
+		link, _ := adapter.getTargetReleaseLink()
+		Expect(reflect.TypeOf(link)).To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseLink{})))
+		Expect(link.Spec.Application).To(Equal("test-app"))
+		Expect(link.Spec.ReleaseStrategy).To(Equal(releaseStrategy.Name))
+		Expect(link.Namespace).Should(Equal(releaseLink.Spec.Target))
+	})
+
+	It("can mark the status for a given Release according to the pipelineRun status", func() {
+		pipelineRun := tekton.NewReleasePipelineRun("test-pipeline", testNamespace).AsPipelineRun()
+		pipelineRun.Name = "foo" // PipelineRun needs a name before registering the Release status
+		err := adapter.registerReleaseStatusData(pipelineRun, releaseStrategy)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(adapter.release.Status.StartTime).ToNot(BeNil())
+		pipelineRun.Status.MarkSucceeded("succeeded", "set it to succeeded")
+		Expect(pipelineRun.IsDone()).To(BeTrue())
+
+		err = adapter.registerReleasePipelineRunStatus(pipelineRun)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(release.HasSucceeded()).To(BeTrue())
+
+		// Double check that we still have a valid StartTime
+		Expect(adapter.release.Status.StartTime).ToNot(BeNil())
+
+		// Clear up previous condition so isDone returns false
+		release.Status.Conditions = []metav1.Condition{}
+		pipelineRun.Status.MarkFailed("failed", "set it to failed")
+		err = adapter.registerReleasePipelineRunStatus(pipelineRun)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(release.HasSucceeded()).To(BeFalse())
+	})
+
+	It("can add the release information to its status and mark it as Running", func() {
+		// The call should succeed if releaseStrategy or pipelineRun are nil
+		Expect(adapter.registerReleaseStatusData(nil, nil)).To(Succeed())
+
+		pipelineRun := tekton.NewReleasePipelineRun("test-pipeline", testNamespace).AsPipelineRun()
+		pipelineRun.Name = "new-name"
+		err := adapter.registerReleaseStatusData(pipelineRun, releaseStrategy)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(release.Status.ReleasePipelineRun).To(Equal(testNamespace + "/" + pipelineRun.Name))
+		Expect(release.Status.Conditions[0].Status == metav1.ConditionUnknown)
+		Expect(release.Status.Conditions[0].Reason == string(appstudiov1alpha1.ReleaseReasonRunning))
+	})
+
+	It("can update the Status of a release", func() {
+		patch := client.MergeFrom(adapter.release.DeepCopy())
+		adapter.release.Status.ReleasePipelineRun = "foo/test"
+		Expect(k8sClient.Status().Patch(ctx, release, patch)).Should(Succeed())
+		Eventually(func() bool {
+			err := k8sClient.Get(ctx, types.NamespacedName{
+				Name:      release.Name,
+				Namespace: testNamespace,
+			}, release)
+			return err == nil && release.Status.ReleasePipelineRun == "foo/test"
+		}).Should(BeTrue())
+	})
+
+	It("can get an existing ApplicationSnapshot", func() {
+		snapshot, err := adapter.getApplicationSnapshot()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(snapshot)).To(Equal(reflect.TypeOf(&appstudioshared.ApplicationSnapshot{})))
+
+		Expect(k8sClient.Delete(ctx, snapshot)).Should(Succeed())
+		Eventually(func() bool {
+			_, err = adapter.getApplicationSnapshot()
+			return errors.IsNotFound(err)
+		}, time.Second*10).Should(BeTrue())
+	})
+
+	It("can return the releaseLink from the release", func() {
+		link, err := adapter.getReleaseLink()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(link)).To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseLink{})))
+
+		// It should err when releaseLink does not exist
+		Expect(k8sClient.Delete(ctx, releaseLink)).Should(Succeed())
+
+		Eventually(func() bool {
+			_, err = adapter.getReleaseLink()
+			return errors.IsNotFound(err)
+		}, time.Second*10).Should(BeTrue())
+	})
+})
--- a/controllers/release/release_controller_test.go
+++ b/controllers/release/release_controller_test.go
@@ -1,0 +1,136 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package release
+
+import (
+	"reflect"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	appstudiov1alpha1 "github.com/redhat-appstudio/release-service/api/v1alpha1"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+)
+
+var _ = Describe("Release Controller", func() {
+	var (
+		manager    ctrl.Manager
+		release    *appstudiov1alpha1.Release
+		reconciler *Reconciler
+		scheme     runtime.Scheme
+		req        ctrl.Request
+	)
+
+	BeforeEach(func() {
+		release = &appstudiov1alpha1.Release{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: testApiVersion,
+				Kind:       "Release",
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "test-release-",
+				Namespace:    testNamespace,
+			},
+			Spec: appstudiov1alpha1.ReleaseSpec{
+				ApplicationSnapshot: "test-snapshot",
+				ReleaseLink:         "test-releaselink",
+			},
+		}
+		Expect(k8sClient.Create(ctx, release)).Should(Succeed())
+
+		req = ctrl.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: release.Namespace,
+				Name:      release.Name,
+			},
+		}
+
+		webhookInstallOptions := &testEnv.WebhookInstallOptions
+
+		var err error
+		manager, err = ctrl.NewManager(cfg, ctrl.Options{
+			Scheme:             clientsetscheme.Scheme,
+			Host:               webhookInstallOptions.LocalServingHost,
+			Port:               webhookInstallOptions.LocalServingPort,
+			CertDir:            webhookInstallOptions.LocalServingCertDir,
+			MetricsBindAddress: "0", // this disables metrics
+			LeaderElection:     false,
+		})
+		Expect(err).NotTo(HaveOccurred())
+		reconciler = NewReleaseReconciler(k8sClient, &logf.Log, &scheme)
+	})
+
+	AfterEach(func() {
+		err := k8sClient.Delete(ctx, release)
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+	})
+
+	It("can create and return a new Reconciler object", func() {
+		Expect(reflect.TypeOf(reconciler)).To(Equal(reflect.TypeOf(&Reconciler{})))
+	})
+
+	It("should reconcile using the ReconcileHandler", func() {
+		adapter := NewAdapter(release, ctrl.Log, k8sClient, ctx)
+		result, err := reconciler.ReconcileHandler(adapter)
+		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(reconcile.Result{})))
+		Expect(err).To(BeNil())
+	})
+
+	It("can fail when Reconcile fails to prepare the adapter when release is not found", func() {
+		Expect(k8sClient.Delete(ctx, release)).Should(Succeed())
+		Eventually(func() error {
+			_, err := reconciler.Reconcile(ctx, req)
+			return err
+		}).Should(BeNil())
+	})
+
+	It("can Reconcile function prepare the adapter and return the result of the reconcile handling operation", func() {
+		result, err := reconciler.Reconcile(ctx, req)
+		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(reconcile.Result{})))
+		Expect(err).To(BeNil())
+	})
+
+	It("can setup the cache by adding a new index field to search for ReleaseLinks", func() {
+		err := setupCache(manager)
+		Expect(err).ToNot(HaveOccurred())
+	})
+
+	It("can setup a new controller manager with the given reconciler", func() {
+		err := setupControllerWithManager(manager, reconciler)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("can setup a new Controller manager and start it", func() {
+		err := SetupController(manager, &ctrl.Log)
+		Expect(err).To(BeNil())
+		go func() {
+			defer GinkgoRecover()
+			err = manager.Start(ctx)
+			Expect(err).NotTo(HaveOccurred())
+		}()
+	})
+})
--- a/controllers/release/release_suite_test.go
+++ b/controllers/release/release_suite_test.go
@@ -1,0 +1,110 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package release
+
+import (
+	"context"
+	"go/build"
+	"path/filepath"
+	"testing"
+
+	"k8s.io/client-go/rest"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
+	appstudiov1alpha1 "github.com/redhat-appstudio/release-service/api/v1alpha1"
+	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+)
+
+const (
+	testApiVersion = "appstudio.redhat.com/v1alpha1"
+	testNamespace  = "default"
+)
+
+var (
+	cfg       *rest.Config
+	k8sClient client.Client
+	testEnv   *envtest.Environment
+	ctx       context.Context
+	cancel    context.CancelFunc
+)
+
+func TestControllerRelease(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Release Controller Test Suite")
+}
+
+var _ = BeforeSuite(func() {
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+	ctx, cancel = context.WithCancel(context.TODO())
+
+	// adding required CRDs, including tekton for PipelineRun Kind
+	testEnv = &envtest.Environment{
+		CRDDirectoryPaths: []string{
+			filepath.Join("..", "..", "config", "crd", "bases"),
+			filepath.Join(
+				build.Default.GOPATH,
+				"pkg", "mod", "github.com", "tektoncd",
+				"pipeline@v0.32.2", "config",
+			),
+			filepath.Join(
+				build.Default.GOPATH,
+				"pkg", "mod", "github.com", "redhat-appstudio", "managed-gitops",
+				"appstudio-shared@v0.0.0-20220603115212-1fb4d804a8c2", "config", "crd", "bases",
+			),
+		},
+		ErrorIfCRDPathMissing: true,
+	}
+
+	var err error
+	cfg, err = testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	Expect(appstudiov1alpha1.AddToScheme(clientsetscheme.Scheme)).To(Succeed())
+	Expect(tektonv1beta1.AddToScheme(clientsetscheme.Scheme)).To(Succeed())
+	Expect(appstudioshared.AddToScheme(clientsetscheme.Scheme)).To(Succeed())
+
+	k8sManager, _ := ctrl.NewManager(cfg, ctrl.Options{
+		Scheme:             clientsetscheme.Scheme,
+		MetricsBindAddress: "0", // this disables metrics
+		LeaderElection:     false,
+	})
+
+	k8sClient = k8sManager.GetClient()
+	go func() {
+		defer GinkgoRecover()
+		Expect(setupCache(k8sManager)).To(Succeed())
+		Expect(k8sManager.Start(ctx)).To(Succeed())
+	}()
+})
+
+var _ = AfterSuite(func() {
+	cancel()
+	By("tearing down the test environment")
+	err := testEnv.Stop()
+	Expect(err).NotTo(HaveOccurred())
+})
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,8 @@ go 1.17
 require (
 	github.com/go-logr/logr v1.2.2
 	github.com/onsi/ginkgo v1.16.5
-	github.com/onsi/gomega v1.17.0
+	github.com/onsi/ginkgo/v2 v2.1.4
+	github.com/onsi/gomega v1.19.0
 	github.com/operator-framework/operator-lib v0.10.0
 	github.com/redhat-appstudio/application-service v0.0.0-20220310134421-5864c77a3d17
 	github.com/stretchr/testify v1.7.1
@@ -20,6 +21,7 @@ require (
 require (
 	github.com/kcp-dev/apimachinery v0.0.0-20220621200107-3d03cbbc3770 // indirect
 	github.com/kcp-dev/logicalcluster v1.0.0 // indirect
+	golang.org/x/mod v0.6.0-dev.0.20220106191415-9b9b3d81d5e3 // indirect
 )
 
 require (
@@ -84,14 +86,13 @@ require (
 	go.uber.org/multierr v1.7.0 // indirect
 	go.uber.org/zap v1.19.1 // indirect
 	golang.org/x/crypto v0.0.0-20210921155107-089bfa567519 // indirect
-	golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
+	golang.org/x/net v0.0.0-20220225172249-27dd8689420f // indirect
 	golang.org/x/oauth2 v0.0.0-20220223155221-ee480838109b // indirect
 	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c // indirect
 	golang.org/x/sys v0.0.0-20220422013727-9388b58f7150 // indirect
 	golang.org/x/term v0.0.0-20210927222741-03fcf44c2211 // indirect
 	golang.org/x/text v0.3.7 // indirect
 	golang.org/x/time v0.0.0-20220210224613-90d013bbcef8 // indirect
-	golang.org/x/tools v0.1.10 // indirect
 	golang.org/x/xerrors v0.0.0-20220411194840-2f41105eb62f // indirect
 	gomodules.xyz/jsonpatch/v2 v2.2.0 // indirect
 	google.golang.org/api v0.63.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -721,6 +721,7 @@ github.com/google/pprof v0.0.0-20201023163331-3e6fc7fc9c4c/go.mod h1:kpwsk12EmLe
 github.com/google/pprof v0.0.0-20201203190320-1bf35d6f28c2/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/pprof v0.0.0-20210122040257-d980be63207e/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/pprof v0.0.0-20210226084205-cbba55b83ad5/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
+github.com/google/pprof v0.0.0-20210407192527-94a9f03dee38/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/pprof v0.0.0-20210601050228-01bbb1931b22/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/pprof v0.0.0-20210609004039-a478d1d731e9/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
 github.com/google/pprof v0.0.0-20210720184732-4bb14d4b1be1/go.mod h1:kpwsk12EmLew5upagYY7GY0pfYCcupk39gWOCRROcvE=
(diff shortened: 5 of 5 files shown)


controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		operations := []ReconcileOperation{
+			cacheAdapter.EnsureFinalizersAreCalled,
+			cacheAdapter.EnsureReleasePipelineRunExists,
+			cacheAdapter.EnsureFinalizerIsAdded,
+			cacheAdapter.EnsureReleasePipelineStatusIsTracked,
+		}
+		for _, operation := range operations {
+			result, err := operation()
+			Expect(!result.CancelRequest && err == nil).To(BeTrue())
+		}
```
davidmogar, line 270: I´d say that you only need an eventually block to check that the finalizer is there and then do your stuff.

davidmogar (commented):
First pass to first half of first file. I added some comments that affect multiple tests, so maybe you could address those and take another look while I keep reviewing.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+			},
+		}
+		Expect(noCacheClient.Create(ctx, releaseLink)).Should(Succeed())
+
+		applicationSnapshot = &appstudioshared.ApplicationSnapshot{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "test-snapshot-",
+				Namespace:    "default",
+			},
+			TypeMeta: metav1.TypeMeta{
```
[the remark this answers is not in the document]
theflockers: It requires the TypeMeta to `client.Create`. The field is gone for some bug I commented on in some old PR. I will look up for it to refresh my mind.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		releaseLink         *appstudiov1alpha1.ReleaseLink
+		applicationSnapshot *appstudioshared.ApplicationSnapshot
+		noCacheClient       client.Client
+
+		//kclient             client.Client
+	)
+
+	BeforeEach(func() {
+		ctx := context.Background()
+		noCacheClient, _ = client.New(cfg, client.Options{Scheme: clientsetscheme.Scheme})
```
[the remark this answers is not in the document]
theflockers: By default when you declare `something *client.Client` it does not come with the cache. I discovered this when trying to make the `Ensure...` functions to work properly. I had to use `manager.GetClient()` to have a client with cache.

That said, I guess I can just keep one. This is kind of a leftover from changes due the tons of problems I had to test the adapter.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+		// some functions needs the cache to be up to work
+		cacheAdapter = NewAdapter(release, ctrl.Log, cacheClient, ctx)
+		Expect(reflect.TypeOf(cacheAdapter)).
+			To(Equal(reflect.TypeOf(&Adapter{})))
+
+	})
+	AfterEach(func() {
+		ctx := context.Background()
+		noCacheClient.Delete(ctx, releaseLink)
```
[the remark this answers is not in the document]
theflockers: That is because In some tests I am deleting the release or other of these objects during the tests then the `AfterEach` breaks the full suite.

Do you know a way to avoid that behavior only for some specific cases?

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		noCacheClient.Delete(ctx, applicationSnapshot)
+		noCacheClient.Delete(ctx, releaseStrategy)
+		noCacheClient.Delete(ctx, release)
+	})
+
+	It("can create a new Adapter instance", func() {
+		Expect(reflect.TypeOf(NewAdapter(release, ctrl.Log, noCacheClient, ctx))).
```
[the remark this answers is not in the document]
theflockers: I will remove the top one. This is actually testing the `NewAdapter` function.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+			To(Equal(reflect.TypeOf(&Adapter{})))
+	})
+
+	It("can make sure Finalizer is added to a Release object", func() {
```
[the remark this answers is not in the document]
theflockers: I still struggle to find proper phrases for these `It`s.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		// unf. the return type and error are the same so we can't
+		// differ the result and error from the previous call
+		result, err = adapter.EnsureFinalizerIsAdded()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+	})
+
+	It("can make sure a PipelineRun object exists for the instanced adapter", func() {
+		// some racing condition in EnvTest makes it throw an error even if the CancelRequest is false
+		// so I am adding a function to wait for it to succeed before continuing.
+		Expect(reflect.TypeOf(cacheAdapter)).
```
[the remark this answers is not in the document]
theflockers: No. Due to having no Idea about what to test. This cleared up after some time on it. Some refactoring is required. Nice to have reviewers!

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+			To(Equal(reflect.TypeOf(&Adapter{})))
+		Eventually(func() bool {
```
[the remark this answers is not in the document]
theflockers: So far yes.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(cacheClient.Delete(ctx, pipelineRun)).Should(Succeed())
+	})
+
+	It("can make sure that the status of a Release Pipeline is tracked", func() {
+		// same as above
+		Eventually(func() bool {
+			result, err := cacheAdapter.EnsureReleasePipelineRunExists()
+			return (!result.CancelRequest && err == nil)
+		}).Should(BeTrue())
+		pipelineRun, _ := cacheAdapter.getReleasePipelineRun()
```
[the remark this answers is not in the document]
theflockers: I did not get you here.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		// release is running but has no pipelineRun matching the release
+		release.MarkRunning()
+		pipelineRun.ObjectMeta.Labels = nil
+		cacheClient.Update(ctx, pipelineRun)
+		release.Status.Conditions = []metav1.Condition{}
+		result, err = cacheAdapter.EnsureReleasePipelineStatusIsTracked()
+
+		// err should not be nil here but cacheAdapter.getReleasePipelineRun does not advise
```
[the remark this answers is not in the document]
theflockers: I added a comment on Slack explaining it better: https://coreos.slack.com/archives/C031USXS2FJ/p1658939031225129

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		// there is no pipelineRun for the release
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		// local cleanup
+		cacheClient.Delete(ctx, pipelineRun)
```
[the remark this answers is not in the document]
theflockers: yes and no. Some others I need to delete the PipelineRun within the test. This makes the test suite fail if there is no object to delete (unless I don't use any `Expect` as I did for the `release*`).

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+	})
+
+	It("can make sure Finalizers are called for a Release object", func() {
+		operations := []ReconcileOperation{
+			cacheAdapter.EnsureFinalizersAreCalled,
+			cacheAdapter.EnsureReleasePipelineRunExists,
+			cacheAdapter.EnsureReleasePipelineStatusIsTracked,
+		}
+
+		for _, operation := range operations {
```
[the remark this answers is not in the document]
theflockers: I done it like this beforehand. Did not like it. That's the only reason.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+			result, err := operation()
+			Expect(!result.CancelRequest && err == nil).To(BeTrue())
+		}
+		release.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
+
+		// should Requeue if there are no Finalizers
+		result, err := cacheAdapter.EnsureFinalizersAreCalled()
```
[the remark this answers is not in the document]
theflockers: I still don't proper understand how this works. I was only actually searching a way to get the expected result. If you this test is wrong I will need some guidance to fix it.

controllers/release/release_adapter_test.go
theflockers: Ok. Now I got what you meant  in the other comment regarding this kind of block.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		err = cacheAdapter.finalizeRelease()
+		Expect(err).Should(Succeed())
+
+		pipelineRun, _ := cacheAdapter.getReleasePipelineRun()
+		Expect(cacheClient.Delete(ctx, pipelineRun)).ShouldNot(Succeed())
+	})
+
+	It("can return the ApplicationSnapshot from the release on the given adapter", func() {
```
davidmogar, line 316: Maybe:
```suggestion
	It("can get an existing ApplicationSnapshot", func() {
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+		// shold err when releaseLink does not exist
+		noCacheClient.Delete(ctx, releaseLink)
+		_, err = adapter.getReleaseLink()
+		Expect(err).Should(HaveOccurred())
+	})
+
+	It("can return the pipelineRun from the release within the give adapter", func() {
+		Eventually(func() bool {
+			result, err := cacheAdapter.EnsureReleasePipelineRunExists()
```
davidmogar, line 342: Not sure you even need this. You create a Release in the `BeforeEach` and that is going to create a PipelineRun already. I'd say that you can add the `getReleasePipelineRun` call here and wait for it to succeed.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: Just tried to remove the `noCacheClient` and all broke.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(release.Status.ReleasePipelineRun, pipelineRun.GetName())
+		Expect(release.Status.Conditions[0].Status == "Unknown")
+		Expect(release.Status.Conditions[0].Reason == "Running")
+	})
+
+	It("can update the Status of a release within the given adapter", func() {
+		applicationSnapshot.TypeMeta.Kind = "ApplicationSnapshot"
+
+		pipelineRun := tekton.NewReleasePipelineRun("test-pipeline", testNamespace).AsPipelineRun()
+		pipelineRun.SetName("test")
```
davidmogar, line 405: In a previous test you use `pipelineRun.Name`. Choose one so we can be consistent ;)

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+		adapter.release.Status.ReleasePipelineRun = fmt.Sprintf("%s%c%s",
+			pipelineRun.GetNamespace(), types.Separator, pipelineRun.GetName())
+
+		Expect(adapter.updateStatus()).Should(Succeed())
+		Expect(adapter.release.Status.ReleasePipelineRun).
+			To(MatchRegexp(pipelineRun.GetName()))
```
davidmogar, line 412: Not a regex isn't it? Should this just be equal?

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
johnbieren: I remember you mentioning this bug, but I am confused as to why we don't have it for the release strategy created on line 59. Also, can we keep the ordering consistent? By that I mean the release link defines the type before the object meta and here we do object meta first

controllers/release/release_adapter_test.go
theflockers: Got it. I thought beforehand that I would need to explicitly call the `Reconcile()` but indeed it is called when the object is handled by the cluster. I have lots of this. I will remove.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(!result.CancelRequest && err != nil).To(BeTrue())
+	})
+
+	It("can create a ReleasePipelineRun within the given adapter", func() {
+		applicationSnapshot.TypeMeta.Kind = "applicationSnapshot"
+		Expect(noCacheClient.Update(ctx, applicationSnapshot)).Should(Succeed())
+		pipelineRun, err := adapter.createReleasePipelineRun(
+			releaseStrategy,
+			applicationSnapshot)
+		Expect(pipelineRun != nil && err == nil).To(BeTrue())
```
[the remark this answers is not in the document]
davidmogar: In that file you would test that the functions work properly. Here you would test that a set of them get called (the set that makes a PipelineRun to be a ReleasePipelineRun). In other words, in one place you test that the machine that produces tires work and the one producing doors do what is expected. In the other place you would test that  a car comes with for wheels and 5 doors.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(reflect.TypeOf(pipelineRun)).
+			To(Equal(reflect.TypeOf(&tektonv1beta1.PipelineRun{})))
+		Expect(pipelineRun).NotTo(BeNil())
+
+		// clean up
+		Expect(cacheClient.Delete(ctx, pipelineRun)).Should(Succeed())
+	})
+
+	It("can make sure that the status of a Release Pipeline is tracked", func() {
+		// same as above
```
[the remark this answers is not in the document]
davidmogar: Voting for removing it.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+	})
+
+	It("can return the releaseStrategy from a given releaseLink to the given adapter", func() {
+		strategy, err := adapter.getReleaseStrategy(releaseLink)
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(strategy)).
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseStrategy{})))
+	})
+
+	It("can return the releaseStrategy from a given adapter", func() {
```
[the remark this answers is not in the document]
davidmogar: Totally. Also, the name should be the one referenced in the ReleaseLink.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		strategy, _ := adapter.getReleaseStrategyFromRelease()
+		Expect(reflect.TypeOf(strategy)).
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseStrategy{})))
+	})
+
+	It("can return the target ReleaseLink from a given ReleaseLink within a given adapter", func() {
```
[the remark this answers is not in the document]
davidmogar: Yep, all conditions should be checked so we don't screw up in the future.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		link, _ := adapter.getTargetReleaseLink()
+		Expect(reflect.TypeOf(link)).
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseLink{})))
+	})
+
+	It("can mark the status for a given Release according to the pipelineRun status", func() {
+		release.MarkRunning()
+		pipelineRun := tekton.NewReleasePipelineRun("test-pipeline", testNamespace).AsPipelineRun()
+
+		pipelineRun.Status.MarkSucceeded("suceedeed", "set it to succeeded")
```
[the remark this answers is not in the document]
davidmogar: Good eye legolas. In any case this wouldn't affect the test. That's just the reason and the message of the condition. The important is the type which passed within the `MarkSucceeded` function.

controllers/release/release_controller_test.go
```diff
@@ -0,0 +1,144 @@
...
+	})
+
+	It("can ReconcileHandler receive an adapter and return the result for the handling operation", func() {
+		adapter := NewAdapter(release, ctrl.Log, cacheClient, ctx)
+		result, err := reconciler.ReconcileHandler(adapter)
+		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(reconcile.Result{})))
+		Expect(err).To(BeNil())
+	})
+
+	It("can Reconcile function fail to prepare the adapter and return the result of the reconcile handling operation", func() {
```
davidmogar, line 110: I don't really understand this test. What is it testing?

controllers/release/release_controller_test.go
```diff
@@ -0,0 +1,144 @@
...
+	})
+
+	It("can Reconcile function prepare the adapter and return the result of the reconcile handling operation", func() {
+		result, err := reconciler.Reconcile(ctx, req)
+		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(reconcile.Result{})))
+		Expect(err).To(BeNil())
+	})
+
+	It("can setup the cache by adding a new index field to search for ReleaseLinks", func() {
+		err := setupCache(k8sManager)
```
davidmogar, line 123: The cache has a `Get` function. Maybe you can inspect if the expected fields are catched.

controllers/release/release_controller_test.go
```diff
@@ -0,0 +1,144 @@
...
+	})
+
+	It("can setup a new controller manager with the given reconciler", func() {
+		err := setupControllerWithManager(k8sManager, reconciler)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("can setup a new Controller manager and start it", func() {
+		// the ctrl.Complete() ignores the object returned by ctrl.Build()
+		// and returns `nil` in case of success. It returns an error otherwise.
```
davidmogar, line 134: I think it's ok to remove this comment.

controllers/release/release_suite_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
+	appstudiov1alpha1 "github.com/redhat-appstudio/release-service/api/v1alpha1"
+	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
```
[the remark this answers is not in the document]
davidmogar: This is already package wide. Constants, even if not exported, are publicly accessible within the package.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+			},
+			Spec: appstudioshared.ApplicationSnapshotSpec{
+				Application: "testapplication",
+				DisplayName: "Test application",
+				Components:  []appstudioshared.ApplicationSnapshotComponent{},
+			},
+		}
+		Expect(noCacheClient.Create(ctx, applicationSnapshot)).Should(Succeed())
+
+		// adding the Type Metas as it loses it after creation
```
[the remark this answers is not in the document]
johnbieren: Yeah, I think it would be nice to do it like you did here 
https://github.com/redhat-appstudio/release-service/blob/main/tekton/pipeline_run_test.go#L121

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
johnbieren: ack

theflockers:
Depends on https://github.com/redhat-appstudio/release-service/pull/76

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: It was due to some test requirement. Fixed the test and the TypeMeta is not required anymore. Removed all of them.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: You fixed it :D

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		// err is ignored.
+		result, _ = cacheAdapter.EnsureFinalizerIsAdded()
+		Expect(!result.CancelRequest).To(BeTrue())
+
+		cacheClient.Delete(ctx, release)
+
+		result, err = cacheAdapter.EnsureFinalizersAreCalled()
+		// err should occur as it is not possible to update the immutable
+		// DeletionTimestamp of the object, so it will throw an error, but
+		// this is ok for the
```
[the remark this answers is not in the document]
theflockers: same.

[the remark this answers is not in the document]
theflockers: comments deleted.

controllers/release/release_adapter_test.go
theflockers: Fixed.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
davidmogar: This sounds better :wink:

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,423 @@
...
+	})
+
+	It("can make sure that the status of a Release Pipeline is tracked", func() {
+		// same as above
+		Eventually(func() bool {
+			result, err := adapter.EnsureReleasePipelineRunExists()
+			return (!result.CancelRequest && err == nil)
+		}).Should(BeTrue())
+		pipelineRun, _ := adapter.getReleasePipelineRun()
+		Expect(pipelineRun).NotTo(BeNil())
```
johnbieren, line 176: Got lost with the new commit, but the comments here https://github.com/redhat-appstudio/release-service/pull/71#discussion_r934476585 still apply to this line

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,423 @@
...
+		}, time.Second*5).Should(BeTrue())
+	})
+
+	It("can return the releaseLink from the release within the given adapter", func() {
+		link, err := adapter.getReleaseLink()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(link)).
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseLink{})))
+
+		// shold err when releaseLink does not exist
```
johnbieren, line 411: `/shold/should/`

controllers/release/release_adapter_test.go
theflockers: Changed to `ContainSubstring` as suggested.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+				GenerateName: "test-snapshot-",
+				Namespace:    "default",
+			},
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: testApiVersion,
+				Kind:       "ApplicationSnapshot",
+			},
+			Spec: appstudioshared.ApplicationSnapshotSpec{
+				Application: "testapplication",
+				DisplayName: "Test application",
```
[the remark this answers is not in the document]
theflockers: removed.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+				Policy:                "test-policy",
+				PersistentVolumeClaim: "test-pvc",
+				ServiceAccount:        "test-account",
+			},
+		}
+		Expect(noCacheClient.Create(ctx, releaseStrategy)).Should(Succeed())
+
+		releaseLink = &appstudiov1alpha1.ReleaseLink{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: "appstudio.redhat.com/v1alpha1",
```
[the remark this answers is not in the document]
theflockers: I removed this `TypeMeta` declaration.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: Removed.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(reflect.TypeOf(link)).
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseLink{})))
+
+		// shold err when releaseLink does not exist
+		noCacheClient.Delete(ctx, releaseLink)
+		_, err = adapter.getReleaseLink()
+		Expect(err).Should(HaveOccurred())
+	})
+
+	It("can return the pipelineRun from the release within the give adapter", func() {
```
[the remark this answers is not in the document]
theflockers: removed all of them.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: Ok, added: 
```		Expect(strategy.Namespace).Should(ContainSubstring(link.Spec.Target))
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		err = adapter.registerReleaseStatusData(pipelineRun, releaseStrategy)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(release.Status.ReleasePipelineRun, pipelineRun.GetName())
+		Expect(release.Status.Conditions[0].Status == "Unknown")
+		Expect(release.Status.Conditions[0].Reason == "Running")
+	})
+
+	It("can update the Status of a release within the given adapter", func() {
+		applicationSnapshot.TypeMeta.Kind = "ApplicationSnapshot"
```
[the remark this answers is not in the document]
theflockers: Answered at https://github.com/redhat-appstudio/release-service/pull/71#discussion_r937800744

controllers/release/release_suite_test.go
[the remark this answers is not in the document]
theflockers: Maybe I done something wrong? Removing it breaks right away.
```
controllers/release/release_controller_test.go:89:49: undefined: logf
```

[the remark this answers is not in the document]
theflockers: Found. Should be called as `log` instead. We also need change in some other test files (the new ones and some old ones).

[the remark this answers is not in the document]
theflockers: No. It keeps importing the log. I suggest we treat this in a separate PR to clean up this `logf` imports.

controllers/release/release_suite_test.go
```diff
@@ -0,0 +1,127 @@
...
+	err = appstudiov1alpha1.AddToScheme(clientsetscheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = tektonv1beta1.AddToScheme(clientsetscheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	err = appstudioshared.AddToScheme(clientsetscheme.Scheme)
+	Expect(err).NotTo(HaveOccurred())
+
+	//webhookInstallOptions := &testEnv.WebhookInstallOptions
```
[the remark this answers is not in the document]
theflockers: removed.

controllers/release/release_suite_test.go
```diff
@@ -0,0 +1,127 @@
+/*
```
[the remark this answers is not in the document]
theflockers: It is basically the test bootstrap. Ginkgo searches for this file to know what test suite to run for a given directory. Not sure if you could add all of the code existing on it to the test itself but it is the way they advise to do. I did not try to write it differently to be honest to see if there is any difference.

controllers/release/release_controller_test.go
```diff
@@ -0,0 +1,144 @@
...
+		req = ctrl.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: release.Namespace,
+				Name:      release.Name,
+			},
+		}
+
+		webhookInstallOptions := &testEnv.WebhookInstallOptions
+
+		k8sManager, _ = ctrl.NewManager(cfg, ctrl.Options{
```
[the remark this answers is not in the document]
theflockers: Yes but this one is required for the test in line #127 as I need an alternative manager to be able to start after setup a controller and it is not used elsewhere.

controllers/release/release_controller_test.go
```diff
@@ -0,0 +1,144 @@
...
+			Scheme:             clientsetscheme.Scheme,
+			Host:               webhookInstallOptions.LocalServingHost,
+			Port:               webhookInstallOptions.LocalServingPort,
+			CertDir:            webhookInstallOptions.LocalServingCertDir,
+			MetricsBindAddress: "0", // this disables metrics
+			LeaderElection:     false,
+		})
+		err := (&appstudiov1alpha1.Release{}).SetupWebhookWithManager(k8sManager)
```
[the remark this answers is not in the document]
theflockers: I guess you removed this line.

controllers/release/release_controller_test.go
theflockers: I expect a failure when the release does not exist, hence can't create the adapter. It exists only to fall into the error condition of the `Reconcile` function and increase the coverage.

controllers/release/release_controller_test.go
theflockers: removed.

controllers/release/release_adapter_test.go
theflockers: Still working on it. I am commenting before pushing it. Too long review.

controllers/release/release_adapter_test.go
theflockers: Same as the comment above.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		pipelineRun, _ := cacheAdapter.getReleasePipelineRun()
+		Expect(cacheClient.Delete(ctx, pipelineRun)).Should(Succeed())
+		Expect(cacheClient.Delete(ctx, release)).Should(Succeed())
+
+		result, err := cacheAdapter.EnsureFinalizersAreCalled()
+		Expect(!result.CancelRequest && err != nil).To(BeTrue())
+	})
+
+	It("can create a ReleasePipelineRun within the given adapter", func() {
+		applicationSnapshot.TypeMeta.Kind = "applicationSnapshot"
```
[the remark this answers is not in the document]
theflockers: done.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: changed to test the `ServiceAccoutName` and the policy defined in the strategy matches. Also this test was moved up a bit to line #273.

controllers/release/release_adapter_test.go
theflockers: changed.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+			cacheAdapter.EnsureFinalizerIsAdded,
+			cacheAdapter.EnsureReleasePipelineStatusIsTracked,
+		}
+		for _, operation := range operations {
+			result, err := operation()
+			Expect(!result.CancelRequest && err == nil).To(BeTrue())
+		}
+		release.SetDeletionTimestamp(&metav1.Time{Time: time.Now()})
+
+		// delete the pipelineRun before the release so it should thrown an
```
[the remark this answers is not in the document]
theflockers: changed.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(err).Should(HaveOccurred())
+	})
+
+	It("can return the releaseLink from the release within the given adapter", func() {
+		link, err := adapter.getReleaseLink()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(link)).
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseLink{})))
+
+		// shold err when releaseLink does not exist
```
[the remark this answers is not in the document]
theflockers: fixed.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: fixed as suggested as well.

controllers/release/release_controller_test.go
```diff
@@ -0,0 +1,144 @@
...
+	AfterEach(func() {
+		ctx := context.Background()
+		cacheClient.Delete(ctx, release)
+	})
+
+	It("can create and return a new Reconciler object", func() {
+		Expect(reflect.TypeOf(reconciler)).To(Equal(reflect.TypeOf(&Reconciler{})))
+	})
+
+	It("can ReconcileHandler receive an adapter and return the result for the handling operation", func() {
```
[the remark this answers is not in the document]
theflockers: fixed.

controllers/release/release_controller_test.go
theflockers: I need to take a look on this still.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
johnbieren: I think the added check is good, but looking again I actually think it makes more sense being in the test before `can return the releaseStrategy from a given releaseLink` since you already have the releaselink there. Can you just move this line to that test instead?

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
johnbieren: Looks good, can you also add `Expect(link.Namespace).Should(Equal(releaseLink.Spec.Target))` as that is the last condition we care about regarding the target release link?

controllers/release/release_suite_test.go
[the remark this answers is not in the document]
johnbieren: I meant the constants, not the imports. Namely `testApiVersion` and `testNamespace`

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+			},
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "test-releaselink-",
+				Namespace:    testNamespace,
+				Labels: map[string]string{
+					"release.appstudio.openshift.io/auto-release": "false",
+				},
+			},
+			Spec: appstudiov1alpha1.ReleaseLinkSpec{
+				DisplayName:     "Test ReleaseLink",
```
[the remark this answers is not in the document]
theflockers: Removed.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		Expect(reflect.TypeOf(adapter)).
+			To(Equal(reflect.TypeOf(&Adapter{})))
+
+		// some functions needs the cache to be up to work
+		cacheAdapter = NewAdapter(release, ctrl.Log, cacheClient, ctx)
+		Expect(reflect.TypeOf(cacheAdapter)).
+			To(Equal(reflect.TypeOf(&Adapter{})))
+
+	})
+	AfterEach(func() {
```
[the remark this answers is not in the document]
theflockers: It seems it was fixed.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+
+		// return success also when the finalizer is present already
+		//
+		// unf. the return type and error are the same so we can't
+		// differ the result and error from the previous call
+		result, err = adapter.EnsureFinalizerIsAdded()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+	})
+
+	It("can make sure a PipelineRun object exists for the instanced adapter", func() {
```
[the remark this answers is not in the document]
theflockers: I also prefer it like this. I don't know why in my mind I was locked into the `can` stuff.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: you changed to 10 and I added a few more with 10 secs as well.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+	It("can make sure Finalizer is added to a Release object", func() {
+		result, err := adapter.EnsureFinalizerIsAdded()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		Expect(release.ObjectMeta.GetFinalizers()[0]).
+			To(ContainSubstring("release-finalizer"))
+
+		// return success also when the finalizer is present already
+		//
+		// unf. the return type and error are the same so we can't
```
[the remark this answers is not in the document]
theflockers: removed. Thanks for the capital letter tip. I forget now I am an actual developer and should be strict with some grammar rules :D

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: I think this one should be fine now. May you crosscheck?

controllers/release/release_suite_test.go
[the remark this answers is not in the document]
theflockers: ah, right. Now I noticed where the comment was placed. I was a bit dizzy already when I got here :D. Taking care of this now.

controllers/release/release_adapter.go
```diff
@@ -19,9 +19,11 @@ package release
 import (
```
theflockers, line 19: This file will be gone from this PR when the required one is merged.

controllers/release/release_adapter_test.go
johnbieren: At time of writing, lines 158 and 170 are the ones I don't think we need. `Expect(pipelineRun).NotTo(BeNil())` we already check for it to be not nil in `getReleasePipelineRun()`

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+
+		// should return results.ContinueProcessing() in case the release is done
+		release.MarkSucceeded()
+		result, err = adapter.EnsureReleasePipelineStatusIsTracked()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		// release is running but has no pipelineRun matching the release
+		release.MarkRunning()
+		pipelineRun.ObjectMeta.Labels = nil
+		k8sClient.Update(ctx, pipelineRun)
```
davidmogar, line 183: You have a race condition here.

The update is not immediate so when `EnsureReleasePipelineStatusIsTracked` is calling `getReleasePipelineRun`, it's probably getting an old version. In fact, the `Expect` you have is valid for the case in which the pipelineRun is tracked or when no pipelineRun is found.

You need to add an `Eventually` block getting the pipelineRun and waiting for the function to return `nil, nil`.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		release.Status.Conditions = []metav1.Condition{}
+		result, err = adapter.EnsureReleasePipelineStatusIsTracked()
+
+		// err should not be nil here but adapter.getReleasePipelineRun does not advise
+		// there is no pipelineRun for the release
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
```
davidmogar, line 189: The check should probably be:
```
Expect(!result.RequeueRequest && pipelineRun == nil).To(BeTrue())
```
I'd also remove the comment.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Expect(pipelineRun).NotTo(BeNil())
+
+		// clean up
+		Expect(k8sClient.Delete(ctx, pipelineRun)).Should(Succeed())
+	})
+
+	It("ensures the Release Pipeline is tracked", func() {
+		Eventually(func() bool {
+			result, err := adapter.EnsureReleasePipelineRunExists()
+			return (!result.CancelRequest && err == nil)
```
davidmogar, line 167: Redundant parenthesis:
```suggestion
			return !result.CancelRequest && err == nil
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		pipelineRun.ObjectMeta.Labels = nil
+		k8sClient.Update(ctx, pipelineRun)
+		release.Status.Conditions = []metav1.Condition{}
+		result, err = adapter.EnsureReleasePipelineStatusIsTracked()
+
+		// err should not be nil here but adapter.getReleasePipelineRun does not advise
+		// there is no pipelineRun for the release
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		// local cleanup
```
davidmogar, line 191:
```suggestion
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			result, err := adapter.EnsureReleasePipelineRunExists()
+			return (!result.CancelRequest && err == nil)
+		}, time.Second*10).Should(BeTrue())
+
+		pipelineRun, _ := adapter.getReleasePipelineRun()
+		Expect(reflect.TypeOf(pipelineRun)).
+			To(Equal(reflect.TypeOf(&tektonv1beta1.PipelineRun{})))
+		Expect(pipelineRun).NotTo(BeNil())
+
+		// clean up
```
davidmogar, line 160:
```suggestion
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+
+		// err should not be nil here but adapter.getReleasePipelineRun does not advise
+		// there is no pipelineRun for the release
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+
+		// local cleanup
+		k8sClient.Delete(ctx, pipelineRun)
+	})
+
+	It("ensures Finalizers are called for a Release object", func() {
```
davidmogar, line 195:
```suggestion
	It("ensures finalizers are called when a release is deleted", func() {
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		// Call to EnsureFinalizersAreCalled should succeed if the Release was not set to be deleted
+		result, err := adapter.EnsureFinalizersAreCalled()
+		Expect(result.CancelRequest).To(BeFalse())
```
davidmogar, line 198: This check also works if the deletion stamp is set but the release contains no finalizer. The proper test should be:
```
Expect(!result.CancelRequest && !result.RequeueRequest).To(BeTrue())
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Eventually(func() error {
+			return k8sClient.Get(ctx, types.NamespacedName{
+				Name:      release.Name,
+				Namespace: release.Namespace,
+			}, release)
+		}, time.Second*10).Should(HaveOccurred())
+
+		Eventually(func() bool {
+			pipelineRun, _ := adapter.getReleasePipelineRun()
+			return (pipelineRun == nil)
```
davidmogar, line 233: Redundant parenthesis:
```suggestion
			return pipelineRun == nil
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		}, time.Second*10).Should(BeTrue())
+	})
+
+	It("ensures Finalizers are called and fails for a Release object when the PipelineRun was already finalized", func() {
```
davidmogar, line 237: Ok, this tests should not exists because that sounds like a bug in code. A release should be properly deleted and not fail if the PipelineRun was deleted previously. Could you create a bug ticket and assign it to me? The proper test would be checking that the release gets deleted even if the PipelineRun doesn't exist, but that would fail without the fix.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		result, err = adapter.EnsureFinalizersAreCalled()
+		Expect(!result.CancelRequest && err != nil).To(BeTrue())
+	})
+
+	It("can create a ReleasePipelineRun", func() {
+		pipelineRun, err := adapter.createReleasePipelineRun(
+			releaseStrategy,
+			applicationSnapshot)
+		Expect(err).ShouldNot(HaveOccurred())
+		Expect(pipelineRun.Spec.ServiceAccountName).Should(ContainSubstring("test-account"))
```
davidmogar, line 274: Missing checks:

- Owner.
- Release labels.
- Params for the snapshot and for params in the ReleaseStrategy.
- Workspace.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,414 @@
...
+		applicationSnapshot.TypeMeta.Kind = "applicationSnapshot"
+		Expect(noCacheClient.Update(ctx, applicationSnapshot)).Should(Succeed())
+		pipelineRun, err := adapter.createReleasePipelineRun(
+			releaseStrategy,
+			applicationSnapshot)
+		Expect(pipelineRun != nil && err == nil).To(BeTrue())
+		Expect(noCacheClient.Delete(ctx, pipelineRun)).Should(Succeed())
+	})
+
+	It("can finalize (delete) the releasePipelineRun just processed", func() {
```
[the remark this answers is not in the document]
davidmogar: Happy to rewrite this code as part of the bug fix. Remove it and I'll make sure I add it in my PR.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Expect(err).Should(Succeed())
+
+		Eventually(func() bool {
+			pipelineRun, _ := adapter.getReleasePipelineRun()
+			return (pipelineRun == nil)
+		}, time.Second*10).Should(BeTrue())
+	})
+
+	It("can return the pipelineRun from the release", func() {
+		Eventually(func() bool {
```
davidmogar, line 305: No race condition AFAIK.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			result, err := adapter.EnsureReleasePipelineRunExists()
+			return (!result.CancelRequest && err == nil)
+		}, time.Second*10).Should(BeTrue())
+		pipelineRun, err := adapter.getReleasePipelineRun()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(pipelineRun)).
```
davidmogar, line 311: One-liner.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			To(Equal(reflect.TypeOf(&tektonv1beta1.PipelineRun{})))
+	})
+
+	It("can return the releaseStrategy from a given releaseLink", func() {
```
davidmogar, line 315:
```suggestion
	It("can return the releaseStrategy referenced in a given releaseLink", func() {
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		strategy, err := adapter.getReleaseStrategy(releaseLink)
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(strategy)).
```
davidmogar, line 318: One-liner.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseStrategy{})))
+	})
+
+	It("can return the releaseStrategy", func() {
+		strategy, _ := adapter.getReleaseStrategyFromRelease()
+		link, _ := adapter.getReleaseLink()
+
+		Expect(reflect.TypeOf(strategy)).
```
davidmogar, line 326: One-liner.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseStrategy{})))
+		Expect(strategy.Namespace).Should(ContainSubstring(link.Spec.Target))
```
davidmogar, line 328: It should be equal.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+	})
+
+	It("can return the target ReleaseLink from a given ReleaseLink", func() {
+		link, _ := adapter.getTargetReleaseLink()
+		Expect(reflect.TypeOf(link)).
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseLink{})))
+		Expect(link.Spec.Application).Should(ContainSubstring("test-app"))
```
davidmogar, line 335: Equal.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Expect(link.Spec.ReleaseStrategy).Should(ContainSubstring(releaseStrategy.Name))
```
davidmogar, line 336: Equal.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+	})
+
+	It("can mark the status for a given Release according to the pipelineRun status", func() {
+		pipelineRun := tekton.NewReleasePipelineRun("test-pipeline", testNamespace).AsPipelineRun()
+		pipelineRun.Name = "foo" // PipelineRun needs a name before registering the Release status
+		err := adapter.registerReleaseStatusData(pipelineRun, releaseStrategy)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(adapter.release.Status.StartTime).To(Not(BeNil()))
```
davidmogar, line 344:
```suggestion
		Expect(adapter.release.Status.StartTime).NotTo(BeNil())
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		pipelineRun.Status.MarkSucceeded("succeeded", "set it to succeeded")
+		Expect(pipelineRun.IsDone()).To(BeTrue())
+
+		err = adapter.registerReleasePipelineRunStatus(pipelineRun)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(release.HasSucceeded()).To(BeTrue())
+
+		Expect(adapter.release.Status.StartTime).ToNot(BeNil()) // Double check that we still have a valid StartTime
+
+		release.Status.Conditions = []metav1.Condition{} // Clear up previous condition so isDone returns false
```
davidmogar, line 354: Funny because I did it, but comments in this tests should be on top of the line and not to the right. It make it easier to follow the code.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		pipelineRun.Status.MarkFailed("failed", "set it to failed")
+		err = adapter.registerReleasePipelineRunStatus(pipelineRun)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(release.HasSucceeded()).To(BeFalse())
+	})
+
+	It("can add the release information to its status and mark it as Running", func() {
+		// err should be nil if releaseStrategy or pipelineRun is nil
+		err := adapter.registerReleaseStatusData(nil, nil)
```
davidmogar, line 363:
```suggestion
		Expecto(adapter.registerReleaseStatusData(nil, nil)).To(Succeed())
```
Removing the check in the next line.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+
+		release.Status.Conditions = []metav1.Condition{} // Clear up previous condition so isDone returns false
+		pipelineRun.Status.MarkFailed("failed", "set it to failed")
+		err = adapter.registerReleasePipelineRunStatus(pipelineRun)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(release.HasSucceeded()).To(BeFalse())
+	})
+
+	It("can add the release information to its status and mark it as Running", func() {
+		// err should be nil if releaseStrategy or pipelineRun is nil
```
davidmogar, line 362:
```suggestion
		// The call should succeed if releaseStrategy or pipelineRun are nil
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		err := adapter.registerReleaseStatusData(nil, nil)
+		Expect(err).To(BeNil())
+
+		pipelineRun := tekton.NewReleasePipelineRun("test-pipeline", testNamespace).AsPipelineRun()
+		pipelineRun.Name = "new-name"
+		err = adapter.registerReleaseStatusData(pipelineRun, releaseStrategy)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(release.Status.ReleasePipelineRun, pipelineRun.GetName())
+		Expect(release.Status.Conditions[0].Status == "Unknown")
```
davidmogar, line 372:
```suggestion
		Expect(release.Status.Conditions[0].Status == metav1.ConditionUnknown)
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Expect(release.Status.Conditions[0].Reason == "Running")
```
davidmogar, line 373:
```suggestion
		Expect(release.Status.Conditions[0].Reason == string(appstudiov1alpha1.ReleaseReasonRunning))
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		// err should be nil if releaseStrategy or pipelineRun is nil
+		err := adapter.registerReleaseStatusData(nil, nil)
+		Expect(err).To(BeNil())
+
+		pipelineRun := tekton.NewReleasePipelineRun("test-pipeline", testNamespace).AsPipelineRun()
+		pipelineRun.Name = "new-name"
+		err = adapter.registerReleaseStatusData(pipelineRun, releaseStrategy)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(release.Status.ReleasePipelineRun, pipelineRun.GetName())
```
davidmogar, line 371: Does that work and properly test the value or should look like:
```suggestion
		Expect(release.Status.ReleasePipelineRun).To(Equal(pipelineRun.GetName()))
```
?

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Expect(release.Status.Conditions[0].Status == "Unknown")
+		Expect(release.Status.Conditions[0].Reason == "Running")
+	})
+
+	It("can update the Status of a release", func() {
+		patch := client.MergeFrom(adapter.release.DeepCopy())
+		adapter.release.Status.ReleasePipelineRun = "foo/test"
+
+		Expect(k8sClient.Status().Patch(ctx, release, patch)).Should(Succeed())
+		Expect(release.Status.ReleasePipelineRun).
```
davidmogar, line 381: You should wrap a Get in an eventually block and check that the value reached the cluster.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			To(Equal("foo/test"))
+	})
+	It("can get an existing ApplicationSnapshot", func() {
```
davidmogar, line 384: Could you add an empty line on top?

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		snapshot, err := adapter.getApplicationSnapshot()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(snapshot)).
```
davidmogar, line 387: One-liner.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			To(Equal(reflect.TypeOf(&appstudioshared.ApplicationSnapshot{})))
+
+		Expect(k8sClient.Delete(ctx, snapshot)).Should(Succeed())
+
+		k8sClient.Get(ctx, types.NamespacedName{
+			Name:      snapshot.Name,
+			Namespace: testNamespace,
+		}, applicationSnapshot)
+
+		Eventually(func() bool {
```
davidmogar, line 397: It's a nice test actually. It's not done in other `get*` functions, but it could be done. Up to you ;)

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			_, err = adapter.getApplicationSnapshot()
+			return errors.IsNotFound(err)
+		}, time.Second*10).Should(BeTrue())
+	})
+
+	It("can return the releaseLink from the release", func() {
+		link, err := adapter.getReleaseLink()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(link)).
```
davidmogar, line 406: One-liner.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+			To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleaseLink{})))
+
+		// should err when releaseLink does not exist
```
davidmogar, line 409: Comment consistency:
```suggestion
		// Should err when releaseLink does not exist
```

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Expect(k8sClient.Delete(ctx, releaseLink)).Should(Succeed())
+		k8sClient.Get(ctx, types.NamespacedName{
+			Name:      releaseLink.Name,
+			Namespace: testNamespace,
+		}, applicationSnapshot)
```
davidmogar, line 414: What is this `Get` for? Not checked and not used.

controllers/release/release_adapter_test.go
theflockers: I changed this to another Eventually block getting the `pipelineRun` and error is nil.

controllers/release/release_adapter_test.go
theflockers: removed.

controllers/release/release_adapter_test.go
theflockers: Ok. fixed.

controllers/release/release_adapter_test.go
theflockers: fixed

controllers/release/release_adapter_test.go
theflockers: Sure, I will create the ticket.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: Please! I want to keep the good coverage :joy:!

controllers/release/release_adapter_test.go
theflockers: I don't like having to use Eventually blocks. If you see them is because it was required by the test. But I am happy to crosscheck.

controllers/release/release_adapter_test.go
theflockers: Changed to `ToNot()` instead as I have it around as well.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: It should contain the namespace: 
```
  Expected
      <string>: default/new-name
  to equal
      <string>: new-name
```
I changed to  `Expect(release.Status.ReleasePipelineRun).To(Equal(testNamespace + "/" + pipelineRun.Name))`

Ok with that or should I use fmt to concat?

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed

controllers/release/release_adapter_test.go
theflockers: You hate so much the missing `\n` that did not see the last required "one-liner". Fixed both!

controllers/release/release_adapter_test.go
theflockers: Looks like copy&paste left over from the previous test. Removed.

controllers/release/release_adapter_test.go
theflockers: Fixed. Added `It should`.

controllers/release/release_adapter_test.go
theflockers: removed.

controllers/release/release_adapter_test.go
theflockers: fixed.

controllers/release/release_adapter_test.go
theflockers: fixed

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+
+		err = k8sClient.Delete(ctx, applicationSnapshot)
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+
+		Expect(k8sClient.Delete(ctx, releaseStrategy)).Should(Succeed())
+	})
+
+	It("can create a new Adapter instance", func() {
+		Expect(reflect.TypeOf(NewAdapter(release, ctrl.Log, k8sClient, ctx))).
```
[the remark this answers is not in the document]
davidmogar: If only we had bigger monitors than in the 90s... :sweat_smile:

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+		Expect(reflect.TypeOf(pipelineRun)).
+			To(Equal(reflect.TypeOf(&tektonv1beta1.PipelineRun{})))
+		Expect(pipelineRun).NotTo(BeNil())
+
+		// clean up
+		Expect(k8sClient.Delete(ctx, pipelineRun)).Should(Succeed())
+	})
+
+	It("ensures the Release Pipeline is tracked", func() {
+		Eventually(func() bool {
```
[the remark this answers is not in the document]
davidmogar: I removed the `Eventually` and I got no errors. Maybe the problem was caused by removing those resources before each tests.

controllers/release/release_adapter_test.go
davidmogar: Also, catch all errors.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,344 @@
...
+		pipelineRun.ObjectMeta.Labels = nil
+		k8sClient.Update(ctx, pipelineRun)
+		release.Status.Conditions = []metav1.Condition{}
+		result, err = adapter.EnsureReleasePipelineStatusIsTracked()
+		Eventually(func() bool {
+			pipelineRun, err := adapter.getReleasePipelineRun()
+			return pipelineRun == nil && err == nil
+
+		}, time.Second*10).Should(BeTrue())
+		k8sClient.Delete(ctx, pipelineRun)
```
davidmogar, line 180: Catch the error with an Expect.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,344 @@
...
+	})
+
+	It("can get an existing ApplicationSnapshot", func() {
+		snapshot, err := adapter.getApplicationSnapshot()
+		Expect(err).Should(Succeed())
+		Expect(reflect.TypeOf(snapshot)).To(Equal(reflect.TypeOf(&appstudioshared.ApplicationSnapshot{})))
+
+		Expect(k8sClient.Delete(ctx, snapshot)).Should(Succeed())
+
+		k8sClient.Get(ctx, types.NamespacedName{
```
davidmogar, line 320: I thought I commented about this one. What for? It does nothing in the test.

controllers/release/release_controller_test.go
```diff
@@ -0,0 +1,137 @@
...
+		err := setupControllerWithManager(manager, reconciler)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("can setup a new Controller manager and start it", func() {
+		err := SetupController(manager, &ctrl.Log)
+		Expect(err).To(BeNil())
+		go func() {
+			defer GinkgoRecover()
+			// can it start the controller?
```
davidmogar, line 132:
```suggestion
```

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: It makes sense. I will give it a try.

controllers/release/release_adapter_test.go
theflockers: Just wondering here if I do have to test all of this here or should I rewrite the test for `tekton.NewReleasePipelineRun` testing these things. it something goes wrong with them it should be something wrong on the tekton package, not the adapter (in this case, I guess).

It seems more appropriate as here my Goal is only test the Status. Are these attributes part of the Status as well?

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: Seems to be working without them. So fixed.

controllers/release/release_adapter_test.go
theflockers: removed.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: Moved. Hope I did move the right one as I changed the description :D last week.

controllers/release/release_adapter_test.go
davidmogar: Here. This test is about creating a ReleasePipelineRun, and a ReleasePipelineRun is basically a PipelineRun, with an owner, release labels and all the other stuff.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: done.

controllers/release/release_suite_test.go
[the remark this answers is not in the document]
theflockers: The way  done it is valid for all tests on the `release/` directory as they are part of the package `release`. Is there at action pending for this one?

controllers/release/release_adapter_test.go
theflockers: true. I was looking into another one. I guess due to me moving tests around I thought your comments where about the  `can add the release information to its status and mark it as Running`

controllers/release/release_adapter_test.go
davidmogar: Even the IDE's complain about it ;)

controllers/release/release_adapter_test.go
davidmogar: That's fine ;)

controllers/release/release_adapter_test.go
theflockers: I will add these changes, although I still think this should be assured in the `tekton.NewReleasePipeline()` test, not here. BTW, it is not done there and should be fixed (I will do when I finish here).

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
davidmogar: This was already removed.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
davidmogar: Completely outdated.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
davidmogar: Exactly, and we return if it's already there. That test could have been done in a different way :shrug:

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
davidmogar: I will bring it back. Just make a note in the ticket ;)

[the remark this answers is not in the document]
davidmogar: Also, it's not good coverage. It's a REALLY GOOD coverage.

controllers/release/release_adapter_test.go
```diff
@@ -0,0 +1,421 @@
...
+
+		Expect(release.ObjectMeta.GetFinalizers()[0]).
+			To(ContainSubstring("release-finalizer"))
+
+		result, err = adapter.EnsureFinalizerIsAdded()
+		Expect(!result.CancelRequest && err == nil).To(BeTrue())
+	})
+
+	It("ensures a PipelineRun object exists", func() {
+		// some racing condition in EnvTest makes it throw an error even if the CancelRequest is false
```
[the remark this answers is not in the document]
davidmogar: Oudated I think. They can be removed AFAIK.

controllers/release/release_adapter_test.go
davidmogar: Nope. It shouldn't be done there. NewReleasePipeline is an entry point to create a ReleasePipelineRun. It creates the bare minimum. Here, we compose the ReleasePipelineRun adding everything that is needed. `createReleasePipelineRun` should contain calls to all those functions to generate a valid ReleasePipelineRun. If in the future any of those calls are removed by mistake, we want to know we screwed up, because that's a big bug.

controllers/release/release_adapter_test.go
davidmogar: Oudated. Not needed.

controllers/release/release_adapter_test.go
theflockers: I will skip this for now.

controllers/release/release_adapter_test.go
davidmogar: This is missing yet.

controllers/release/release_adapter_test.go
theflockers: Ticket created:  https://issues.redhat.com/browse/HACBS-987

PR #73, opened 2022-07-20
feat(HACBS-695): add support for KCP virtual workspaces

johnbieren:


johnbieren:
@davidmogar this PR is very WIP and I have done 0 testing. You will also see `TODO`s I have left in the code for things I still need to figure out. But, I wanted to open it so you can get an idea of what I plan to do for this and shout early if you don't agree with the direction I am taking

api/v1alpha1/releaselink_types.go
```diff
@@ -34,7 +47,7 @@ type ReleaseLinkSpec struct {
 	// Target is a reference to the workspace to establish a link with
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +required
-	Target string `json:"target"`
+	Target Target `json:"target"`
```
davidmogar, line 50: This should be renamed to `TargetReleaseLink` so it's more clear. This will be done as part of a future PR to split the ReleaseLink in two.

api/v1alpha1/releaselink_types.go
```diff
@@ -20,6 +20,19 @@ import (
...
+type Target struct {
+	// Namespace is a reference to the namespace to establish a link with
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	Namespace string `json:"namespace"`
+
+	// Workspace is a reference to the workspace to establish a link with in a KCP setup
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +optional
+	Workspace string `json:"workspace,omitempty"` // TODO - make sure when empty, it equals the empty string and can still be passed along to functions
```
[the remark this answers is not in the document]
davidmogar: Also, we should probably target a single ReleaseLink. That's it, adds the ReleaseLink name here. Let's do this as part of a future PR to split the ReleaseLink in two.

controllers/release/release_adapter.go
```diff
@@ -332,7 +345,7 @@ func (a *Adapter) registerReleaseStatusData(releasePipelineRun *v1beta1.Pipeline
 		releasePipelineRun.Namespace, types.Separator, releasePipelineRun.Name)
 	a.release.Status.ReleaseStrategy = fmt.Sprintf("%s%c%s",
 		releaseStrategy.Namespace, types.Separator, releaseStrategy.Name)
-	a.release.Status.TargetWorkspace = releasePipelineRun.Namespace
+	a.release.Status.TargetWorkspace = releasePipelineRun.Namespace // TODO - do we want to have both workspace and namespace here?
```
davidmogar, line 348: Probably.

controllers/release/release_controller.go
```diff
@@ -69,7 +70,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
-	adapter := NewAdapter(release, logger, r.Client, ctx)
+	adapter := NewAdapter(release, logger, r.Client, ctx, req.ClusterName)
```
davidmogar, line 73: Ok, not sure this is quite right. I think we can improve the code by not passing the clusterName here. We can replace this line with the following:
```
if req.ClusterName != "" {
  ctx = logicalcluster.WithCluster(context, logicalcluster.New(cluster))
}
adapter := NewAdapter(release, logger, r.Client, ctx)
```

Then, all the instances that were using the context beforehand, can keep the same code.

For those places in which you actually need to get the name of the cluster, you can use the following:
```
clusterName, ok := logicalcluster.ClusterFromContext(a.context)
if ok { // We are running on KCP
  // whatever
}
```

I haven't tested any of this though, so give it a go yourself.

api/v1alpha1/releaselink_types.go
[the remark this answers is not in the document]
johnbieren: I don't follow. Instead of targeting a namespace and workspace we want to target a ReleaseLink name?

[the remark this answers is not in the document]
davidmogar: Please ignore. We need to keep it as it is, you are right.

controllers/release/release_controller.go
davidmogar: As discussed only, the context passed here would target the `dev` workspace always as we are reconciling a Release that is always created in the `dev` workspace. For the `managed` one, we should add another context to the adapter that is set once the ReleaseLink is loaded and we know the name of the workspace. If the Workspace field is not set, we should reference the dev context.

api/v1alpha1/releaselink_types.go
johnbieren: I believe we said offline not to rename here. Feel free to unresolve if that is not the case

controllers/release/release_adapter.go
johnbieren: Done

controllers/release/release_controller.go
johnbieren: Done

api/v1alpha1/releaselink_types.go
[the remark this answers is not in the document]
johnbieren: Done

go.mod
```diff
@@ -114,3 +112,5 @@ require (
 )
 
 replace sigs.k8s.io/controller-runtime => github.com/kcp-dev/controller-runtime v0.11.3-0.20220624161137-f6e5a2f56683
+
+replace github.com/kcp-dev/logicalcluster => github.com/johnbieren/logicalcluster v1.1.1
```
johnbieren, line 116: This is a temporary line for testing that I don't plan to merge with

controllers/release/release_adapter.go
```diff
@@ -155,19 +159,37 @@ func (a *Adapter) EnsureReleasePipelineStatusIsTracked() (results.OperationResul
...
+	releaseLink, err := a.getReleaseLink()
+	if err != nil {
+		return results.RequeueWithError(err)
+	}
+
+	if a.targetContext == nil {
+		if releaseLink.Spec.Target.Workspace == "" {
+			a.targetContext = a.context
+		} else {
+			a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(releaseLink.Spec.Target.Workspace))
```
johnbieren, line 174: reminder to myself to test what happens if I run a get operation using a context for a cluster that doesn't exist

api/v1alpha1/releaselink_types.go
davidmogar: Yeah, that's what my comment here says :stuck_out_tongue_closed_eyes:

controllers/release/release_adapter.go
davidmogar: That's a good test. I'm expecting it to simply fail and those errors are already catched. In any case, I agree on this 
 being a good thing to test.

johnbieren:
Will reopen after rebasing and testing is all done

johnbieren:
I force pushed while it was closed so GitHub won't let me reopen it. New PR for this is https://github.com/redhat-appstudio/release-service/pull/88

PR #75, opened 2022-08-01
feat(HACBS-930): check the managed ReleaseLink's auto-release label

johnbieren:
This commit changes the functionality of the release controller to
only continue processing a release if the ReleaseLink in the managed
namespace is setup for auto-releases. This is done via the a label
as described in HACBS-555.


controllers/release/release_adapter.go
```diff
@@ -210,6 +214,45 @@ func (a *Adapter) getApplicationSnapshot() (*appstudioshared.ApplicationSnapshot
...
+		return nil, err
+	}
+
+	releaseLinks := &v1alpha1.ReleaseLinkList{}
+	labelSelector := labels.NewSelector()
+	labelRequirement, err := labels.NewRequirement("release.appstudio.openshift.io/auto-release", selection.NotIn, []string{"false"})
+	if err != nil {
+		return nil, err
+	}
+	labelSelector = labelSelector.Add(*labelRequirement)
```
davidmogar, line 233: Let's not create a variable. It's only used once, so do the `Add` in line 238.

controllers/release/release_adapter.go
```diff
@@ -210,6 +214,45 @@ func (a *Adapter) getApplicationSnapshot() (*appstudioshared.ApplicationSnapshot
...
+		LabelSelector: labelSelector,
+	}
+
+	err = a.client.List(a.context, releaseLinks, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, foundReleaseLink := range releaseLinks.Items {
+		if foundReleaseLink.Spec.Application == releaseLink.Spec.Application {
```
davidmogar, line 247: It seems that the `integration-service` [is able to list using 2 fields](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/integration/integration_adapter.go#L176). In the past we opted for this code because we couldn't make it work. Should we take the chance to improve this code or do you prefer to do it in a different PR?

controllers/release/release_adapter.go
```diff
@@ -210,6 +214,45 @@ func (a *Adapter) getApplicationSnapshot() (*appstudioshared.ApplicationSnapshot
...
+			return &foundReleaseLink, nil
+		}
+	}
+
+	return nil, fmt.Errorf("no ReleaseLink found in target workspace '%s' with target '%s' and application '%s'",
```
davidmogar, line 252:
```suggestion
	return nil, fmt.Errorf("no active ReleaseLink found in target workspace '%s' with target '%s' and application '%s'",
```

controllers/release/release_adapter.go
```diff
@@ -210,6 +214,45 @@ func (a *Adapter) getApplicationSnapshot() (*appstudioshared.ApplicationSnapshot
 	return applicationSnapshot, nil
 }
 
+// getEnabledTargetReleaseLink returns the ReleaseLink targeted by ReleaseLink in the Release being processed.
+// Only ReleaseLinks with the auto-release label set to true (or missing the label, which we treat the same as
```
davidmogar, line 218:
```suggestion
// Only ReleaseLinks with the auto-release label set to true (or missing the label, which is treated the same as
```

controllers/release/release_adapter.go
```diff
@@ -210,6 +214,45 @@ func (a *Adapter) getApplicationSnapshot() (*appstudioshared.ApplicationSnapshot
...
+// having the label and it being set to true) will be searched for. If a matching ReleaseLink is not found or
+// the List operation fails, an error will be returned.
+func (a *Adapter) getEnabledTargetReleaseLink() (*v1alpha1.ReleaseLink, error) {
```
davidmogar, line 221: Maybe change it to `getActiveTargetReleaseLink`? For some reason `enabled` sounds weird to me. Leaving the decision to you as you are the native speaker.

controllers/release/release_adapter.go
johnbieren: Done

controllers/release/release_adapter.go
johnbieren: Done

controllers/release/release_adapter.go
johnbieren: Done

controllers/release/release_adapter.go
johnbieren: I think English wise they mean about the same so if active sounds better to you, I am fine with that. Changed

controllers/release/release_adapter.go
johnbieren: Can you elaborate? Are you saying incorporate this into the fieldselector and no longer require this loop?

davidmogar: Exactly. If it's possible (and it seems it is) the code would be much cleaner.

johnbieren: Sounds good, I'll test it out and comment back when I have it working (so no need to look at changes until then)

davidmogar: I converted this PR to a draft to better track the in-progress status.

johnbieren:
Tested and moved from draft to ready. I am fine on waiting for https://github.com/redhat-appstudio/release-service/pull/71 to merge before we merge this, though

controllers/release/release_adapter_test.go
```diff
@@ -311,11 +312,22 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
 		Expect(link.Namespace).Should(Equal(releaseLink.Spec.Target))
+
+		// It should return nil if the ReleaseLink has auto-release set to false
+		patch := client.MergeFrom(link.DeepCopy())
+		link.SetLabels(map[string]string{
+			"release.appstudio.openshift.io/auto-release": "false",
+		})
+		Expect(k8sClient.Patch(ctx, link, patch)).Should(Succeed())
+		Eventually(func() bool {
+			_, err := adapter.getActiveTargetReleaseLink()
```
theflockers, line 328: I would also catch the result of `_`. If the return value of the function changes in the future to not be `nil`  (for any reason) I guess keeping this would cause a wrong `success`.

theflockers (changes requested):

controllers/release/release_adapter_test.go
johnbieren: Fixed I think, let me know if not

theflockers (approved):

controllers/release/release_adapter.go
```diff
@@ -194,6 +195,48 @@ func (a *Adapter) finalizeRelease() error {
...
+
+		}
+	}
+
+	if !containsValidReleaseLink {
+		return nil, fmt.Errorf("no ReleaseLink found in target workspace '%s' with target '%s' and application '%s'",
+			releaseLink.Spec.Target, releaseLink.Namespace, releaseLink.Spec.Application)
+	}
+
+	return &releaseLinks.Items[0], nil
```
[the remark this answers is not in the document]
johnbieren: Agreed but resolving this because it sounds we are in agreement to leave it as is for now

controllers/release/release_adapter_test.go
```diff
@@ -311,11 +312,34 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+		})
+		Expect(k8sClient.Patch(ctx, link, patch)).Should(Succeed())
+		Eventually(func() bool {
+			targetReleaseLink, err := adapter.getActiveTargetReleaseLink()
+			return targetReleaseLink == nil && err != nil &&
+				strings.Contains(err.Error(), "with auto-release label set to false")
+		}, time.Second*10).Should(BeTrue())
+	})
+
+	It("can fail to find the target ReleaseLink when target does not match", func() {
```
davidmogar, line 334:
```suggestion
	It("fails to find the target ReleaseLink when target does not match", func() {
```

PR #77, opened 2022-08-09
fix(HACBS-987): fix finalization when no PR is present

davidmogar:
There was an edge case in which the finalization could fail when the PipelineRun was not present.

### DIFF
--- a/controllers/release/release_adapter_test.go
+++ b/controllers/release/release_adapter_test.go
@@ -236,6 +236,21 @@ var _ = Describe("Release Adapter", Ordered, func() {
 		}, time.Second*10).Should(BeTrue())
 	})
 
+	It("can finalize a Release", func() {
+		// Release should be properly finalized if there is no ReleasePipelineRun
+		Expect(adapter.finalizeRelease()).To(Succeed())
+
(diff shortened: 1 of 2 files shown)


controllers/release/release_adapter_test.go
```diff
@@ -236,6 +236,17 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
 
+	It("can finalize a Release", func() {
+		// Release should be properly finalized if there is no ReleasePipelineRun
+		Expect(adapter.finalizeRelease()).To(Succeed())
+
+		// It should finalize properly as well when there's a ReleasePipelineRun
+		result, err := adapter.EnsureReleasePipelineRunExists()
+		Expect(result.CancelRequest).To(BeFalse())
+		Expect(err).NotTo(HaveOccurred())
+		Expect(adapter.finalizeRelease()).To(Succeed())
```
theflockers, line 247: maybe adding an `Eventually` block with Get here to make sure it was really gone?

davidmogar: Absolutely right. My bad.

theflockers (approved):

PR #80, opened 2022-08-09
test(HACBS-913): increase coverage in controllers/results/ directory

theflockers:
- adds test to `results/` directory

### DIFF
--- a/controllers/results/operation_results_test.go
+++ b/controllers/results/operation_results_test.go
@@ -1,0 +1,131 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package results
+
+import (
+	"errors"
+	"reflect"
+	"testing"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestOperationResult(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Operation Results Test Suite")
+}
+
+var _ = BeforeSuite(func() {
+	By("tearing down the test environment")
+})
+
+var _ = Describe("Operation Results", func() {
+
+	var result *OperationResult
+	var testErr = errors.New("Test")
+
+	BeforeEach(func() {
+		result = &OperationResult{}
+		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(&OperationResult{})))
+	})
+
+	Context("When calling ContinueProcessing", func() {
+		It("returns a tuple instructing the reconcile loop to continue processing operations", func() {
+			result, err := ContinueProcessing()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).Should(And(
+				HaveField("RequeueDelay", time.Duration(0)),
+				HaveField("RequeueRequest", false),
+				HaveField("CancelRequest", false),
+			))
+		})
+	})
+
+	Context("When calling Requeue", func() {
+		It("returns a tuple instructing the reconcile loop to requeue the object", func() {
+			result, err := Requeue()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).Should(And(
+				HaveField("RequeueDelay", time.Duration(0)),
+				HaveField("RequeueRequest", true),
+				HaveField("CancelRequest", false),
+			))
+		})
+	})
+
+	Context("When calling RequeueWithError", func() {
+		It("returns a tuple instructing the reconcile loop to requeue the object with the given reconcile error", func() {
+			result, err := RequeueWithError(testErr)
+			Expect(err.Error()).To(Equal(testErr.Error()))
+			Expect(result).Should(And(
+				HaveField("RequeueDelay", time.Duration(0)),
+				HaveField("RequeueRequest", true),
+				HaveField("CancelRequest", false),
+			))
+		})
+	})
+
+	Context("When calling RequeueAfter", func() {
+		It("returns a tuple instructing the reconcile loop to requeue the object after the given delay", func() {
+			result, err := RequeueAfter(time.Duration(1), testErr)
+			Expect(err.Error()).To(Equal(testErr.Error()))
+			Expect(result).Should(And(
+				HaveField("RequeueDelay", time.Duration(1)),
+				HaveField("RequeueRequest", true),
+				HaveField("CancelRequest", false),
+			))
+		})
+	})
+
+	Context("When calling RequeueOnErrorOrContinue", func() {
+		It("returns a tuple instructing the reconcile loop to requeue the object in case of an error or to continue processing operations", func() {
+			result, err := RequeueOnErrorOrContinue(testErr)
+			Expect(err.Error()).To(Equal(testErr.Error()))
+			Expect(result).Should(And(
+				HaveField("RequeueDelay", time.Duration(0)),
+				HaveField("RequeueRequest", false),
+				HaveField("CancelRequest", false),
+			))
+		})
+	})
+
+	Context("When calling RequeueOnErrorOrStop", func() {
+		It("returns a tuple instructing the reconcile loop to requeue the object in case of an error or to stop processing operations", func() {
+			result, err := RequeueOnErrorOrStop(testErr)
+			Expect(err.Error()).To(Equal(testErr.Error()))
+			Expect(result).Should(And(
+				HaveField("RequeueDelay", time.Duration(0)),
+				HaveField("RequeueRequest", false),
+				HaveField("CancelRequest", true),
+			))
+		})
+	})
+
+	Context("When calling StopProcessing", func() {
+		It("returns a tuple instructing the reconcile loop to stop processing operations", func() {
+			result, err := StopProcessing()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(result).To(And(
+				HaveField("RequeueDelay", time.Duration(0)),
+				HaveField("RequeueRequest", false),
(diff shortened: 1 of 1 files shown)


controllers/results/results_suite_test.go
```diff
@@ -0,0 +1,33 @@
+/*
```
theflockers, line 1: wondering if this file is required although using it is pointed as "Best practices". Thoughts?

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+var _ = Describe("Operation Results", func() {
+	var result *OperationResult
+
+	BeforeEach(func() {
+		result = &OperationResult{}
+		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(&OperationResult{})))
+	})
+
+	It("returns an tuple instructing the reconcile loop to continue processing operations", func() {
```
johnbieren, line 37: `s/returns an/returns a/`

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+		Expect(ContinueProcessing()).Should(And(
+			HaveField("RequeueDelay", time.Duration(0)),
+			HaveField("RequeueRequest", false),
+			HaveField("CancelRequest", false),
+		))
+	})
+
+	It("// Requeue returns an (OperationResult, error) tuple instructing the reconcile loop to requeue the object", func() {
```
johnbieren, line 45: what is going on with this string? at least the `//` is not intended, right?

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+var _ = Describe("Operation Results", func() {
+	var result *OperationResult
+
+	BeforeEach(func() {
+		result = &OperationResult{}
+		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(&OperationResult{})))
+	})
+
+	It("returns an tuple instructing the reconcile loop to continue processing operations", func() {
+		Expect(ContinueProcessing()).Should(And(
```
johnbieren, line 38: should we also check that the error is `nil`?

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+			HaveField("RequeueDelay", time.Duration(0)),
+			HaveField("RequeueRequest", false),
+			HaveField("CancelRequest", false),
+		))
+	})
+
+	It("// Requeue returns an (OperationResult, error) tuple instructing the reconcile loop to requeue the object", func() {
+		Expect(Requeue()).Should(And(
```
johnbieren, line 46: should we also check that the error is `nil`?

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+			HaveField("RequeueDelay", time.Duration(0)),
+			HaveField("RequeueRequest", true),
+			HaveField("CancelRequest", false),
+		))
+	})
+
+	It("returns an tuple instructing the reconcile loop to requeue the object with the given reconcile error", func() {
```
johnbieren, line 53: `s/an tuple/a tuple/`

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+		result, err := RequeueWithError(errors.New("Test"))
+		Expect(fmt.Sprint(err)).To(Equal("Test"))
```
johnbieren, line 55: a variable for the error string instead of having to type it all over the place may be nice, but I don't feel strongly about this. up to you

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+		Expect(result).Should(And(
+			HaveField("RequeueDelay", time.Duration(0)),
+			HaveField("RequeueRequest", true),
+			HaveField("CancelRequest", false),
+		))
+	})
+
+	It("returns an tuple instructing the reconcile loop to requeue the object after the given delay", func() {
```
johnbieren, line 63: `s/an tuple/a tuple/`

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+		result, err := RequeueAfter(time.Duration(10), errors.New("Test"))
```
johnbieren, line 64: I can't really tell the unit of duration from the `operation_results.go` code. Is this seconds? if so maybe we should use like 1 instead of 10

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+		Expect(fmt.Sprint(err)).To(Equal("Test"))
+		Expect(result).Should(And(
+			HaveField("RequeueDelay", time.Duration(10)),
+			HaveField("RequeueRequest", true),
+			HaveField("CancelRequest", false),
+		))
+	})
+
+	It("returns an tuple instructing the reconcile loop to requeue the object in case of an error or to continue processing operations", func() {
```
johnbieren, line 73: `s/an tuple/a tuple/`

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+		result, err := RequeueOnErrorOrContinue(errors.New("Test"))
+		Expect(fmt.Sprint(err)).To(Equal("Test"))
+		Expect(result).Should(And(
+			HaveField("RequeueDelay", time.Duration(0)),
+			HaveField("RequeueRequest", false),
+			HaveField("CancelRequest", false),
+		))
+	})
+
+	It("returns an tuple instructing the reconcile loop to requeue the object in case of an error or to stop processing operations", func() {
```
johnbieren, line 83: `s/an tuple/a tuple/`

controllers/results/operation_results_test.go
```diff
@@ -0,0 +1,100 @@
...
+		result, err := RequeueOnErrorOrStop(errors.New("Test"))
+		Expect(fmt.Sprint(err)).To(Equal("Test"))
+		Expect(result).Should(And(
+			HaveField("RequeueDelay", time.Duration(0)),
+			HaveField("RequeueRequest", false),
+			HaveField("CancelRequest", true),
+		))
+	})
+
+	It("returns an tuple instructing the reconcile loop to stop processing operations", func() {
```
johnbieren, line 93: `s/an tuple/a tuple/`

johnbieren (changes requested):

controllers/results/operation_results_test.go
theflockers: Copy & Paste failure.

controllers/results/operation_results_test.go
theflockers: The check for `nil` is implicit. If it return an error it will break.

controllers/results/operation_results_test.go
theflockers: copy & paste issue again - got them from the `operation_results.go`  file. It needs fixes there as well.

controllers/results/operation_results_test.go
theflockers: I have no opinion on it. Just chose an arbitrary number to check. I am fine changing it to 1.

controllers/results/operation_results_test.go
theflockers: I am not that bad in grammar. Just did not check what I was copying :joy:

controllers/results/results_suite_test.go
theflockers: file removed.

controllers/results/operation_results_test.go
theflockers: fixed

controllers/results/operation_results_test.go
theflockers: fixed.

controllers/results/operation_results_test.go
theflockers: fixed.

controllers/results/operation_results_test.go
theflockers: Fixed. I made it explicit as well.

controllers/results/operation_results_test.go
theflockers: fixed.

controllers/results/operation_results_test.go
theflockers: fixed.

controllers/results/operation_results_test.go
theflockers: changed.

controllers/results/operation_results_test.go
theflockers: fixed.

controllers/results/operation_results_test.go
theflockers: fixed.

controllers/results/operation_results_test.go
theflockers: fixed.

controllers/results/operation_results_test.go
johnbieren: it is more clear this way imo

controllers/results/operation_results_test.go
johnbieren: I think the former because the first word you say after a/an starts with a vowel in that case. like I would read that aloud as "returns an operation result error tuple"

johnbieren (approved):

davidmogar (commented):
You already have an approval so you don't need mine. Also, what I'm going to comment about is not required and it's only my opinion (which is *always* the best one :stuck_out_tongue_winking_eye:), so you have two reasons to merge this change ignoring my comment.

Thing is that reading through the tests is confusing to me. For example:
```
It("returns a tuple instructing the reconcile loop to continue processing operations"...
```
Who? What? I have to go to the code to know that.

In this case the `Describe` is very generic and in a file with just global functions, I don't think it's good enough. I would probably have a `Context` for each test. The previous example would become something like the following:
```
var _ = Describe("Operation Results", func() {

	var result *OperationResult
	var testErr = errors.New("Test")

	BeforeEach(func() {
		result = &OperationResult{}
		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(&OperationResult{})))
	})

        Context("When calling ContinueProcessing", func() {
	        It("returns a tuple instructing the reconcile loop to continue processing operations", func() { ... }
	}
```
Not sure if you have to wrap it with another `Describe` (I don't think so), but you get the idea. It becomes easier to understand.

Apart from this, all good.

davidmogar (approved):

PR #82, opened 2022-08-11
chore: changes codecov settings

theflockers:
changes codecov settings

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,2 +1,16 @@
+# Documentation: https://docs.codecov.io/docs/codecov-yaml
 ignore:
   - "api/v1alpha1/zz_generated.deepcopy.go" # generated file
+
+coverage:
+    status:
+        # Allows coverage to drop by a 3% when compared against the base commit.
+        project:
+            default:
+                target: auto
+                threshold: 3%
+        # Sets the expected status for `codecov/patch` check.
+        # We set this to be only informational hence it does not cause the check to fail.
+        patch:
+            default:
+                informational: true


codecov.yml
```diff
@@ -1,2 +1,23 @@
...
+
+# general codecov settings
+codecov:
+    notify:
+        require_ci_to_pass: false
+coverage:
+    status:
+        # sets the expected status for `codecov/project` check.
+        # `target: auto` means it will use the current coverage to calculate against
+        #          the give PR/commit.
```
johnbieren, line 13: given?

codecov.yml
```diff
@@ -1,2 +1,23 @@
+# Documentation: https://docs.codecov.io/docs/codecov-yaml
 ignore:
   - "api/v1alpha1/zz_generated.deepcopy.go" # generated file
+
+# general codecov settings
+codecov:
+    notify:
+        require_ci_to_pass: false
```
johnbieren, line 8: what does this mean?

codecov.yml
```diff
@@ -1,2 +1,23 @@
...
+coverage:
+    status:
+        # sets the expected status for `codecov/project` check.
+        # `target: auto` means it will use the current coverage to calculate against
+        #          the give PR/commit.
+        # theshould is the max delta allowed when compared with the current coverage
```
johnbieren, line 14: threshold?

codecov.yml
```diff
@@ -1,2 +1,23 @@
...
+        project:
+            default:
+                target:  auto
+                threshold: 3%
```
johnbieren, line 18: so what does this mean? we fail the check unless we increase the code coverage by 3%?

codecov.yml
```diff
@@ -1,2 +1,23 @@
...
+        # sets the expected status for `codecov/patch` check
+        # we set this to be only informatinal hence it does not cause the check to fail.
```
johnbieren, line 20: informational

codecov.yml
```diff
@@ -1,2 +1,23 @@
...
+        patch:
+            default:
+                informational: true
```
johnbieren, line 23: this sounds redundant with line 8

codecov.yml
theflockers: _Should Codecov wait for all other statuses to pass before sending its status._

It is only about notifications.

codecov.yml
theflockers: This is indeed confusing.
The target is the number that defines the coverage we want. _Auto_ means codecov will calculate it by itself (the default). The threshold is the amount in percent that it can miss from the target. If the target is 80%, it will pass starting from 77%. It happens that at each push (because of codecov characteristics) the coverage result is flaky. This is basically to avoid this.

codecov.yml
theflockers: I guess this was clarified in that comment. This makes specifically `codecov/patch` as informational only.

codecov.yml
johnbieren: So the statuses are what, patch and project? This just means don't wait for the project result to add the patch result (and vice versa)?

codecov.yml
```diff
@@ -1,2 +1,23 @@
...
+# general codecov settings
+codecov:
+    notify:
+        require_ci_to_pass: false
+coverage:
+    status:
+        # sets the expected status for `codecov/project` check.
+        # `target: auto` means it will use the current coverage to calculate against
+        #          the given PR/commit.
+        # theshold is the max delta allowed when compared with the current coverage
```
johnbieren, line 14: `s/theshold/threshold/`. Also the line 12-13 description ends in a period and this one does not

codecov.yml
theflockers: that's right.

davidmogar: According to the docs, `require_ci_to_pass` is used to `Show the Codecov [status without waiting](https://docs.codecov.com/docs/codecovyml-reference#codecovrequire_ci_to_pass) for other status checks to complete or pass.
```
So I'm not getting why it works. We currently have a CI step that never completes or passes but we still get the notification. What am I missing?

File: codecov.yml [hunk 8]
```diff
@@ -1,2 +1,23 @@
...
+        require_ci_to_pass: false
+coverage:
+    status:
+        # sets the expected status for `codecov/project` check.
+        # `target: auto` means it will use the current coverage to calculate against
+        #          the given PR/commit.
+        # threshold is the max delta allowed when compared with the current coverage.
+        project:
+            default:
+                target:  auto
```
davidmogar, line 17:
```suggestion
                target: auto
```

johnbieren (approved):

codecov.yml
theflockers: I understood it only regards to notifications.

theflockers: thinking now this line is useless. I will remove it.

codecov.yml
```diff
@@ -1,2 +1,19 @@
...
+    status:
+        # sets the expected status for `codecov/project` check.
+        # `target: auto` means it will use the current coverage to calculate against
+        #          the given PR/commit.
+        # threshold is the max delta allowed when compared with the current coverage.
+        project:
+            default:
+                target: auto
+                threshold: 3%
+        # sets the expected status for `codecov/patch` check.
```
davidmogar, line 15: I don't think this is needed.
```suggestion
```

codecov.yml
```diff
@@ -1,2 +1,19 @@
...
+        # we set this to be only informational hence it does not cause the check to fail.
```
davidmogar, line 16:
```suggestion
        # We set patch status to be informational hence it does not cause the check to fail.
```

codecov.yml
```diff
@@ -1,2 +1,19 @@
+# Documentation: https://docs.codecov.io/docs/codecov-yaml
 ignore:
   - "api/v1alpha1/zz_generated.deepcopy.go" # generated file
+
+coverage:
+    status:
+        # sets the expected status for `codecov/project` check.
```
davidmogar, line 7: I don't think this is needed.
```suggestion
```

codecov.yml
```diff
@@ -1,2 +1,19 @@
...
+        # `target: auto` means it will use the current coverage to calculate against
+        #          the given PR/commit.
+        # threshold is the max delta allowed when compared with the current coverage.
```
davidmogar, line 10: I would change the whole comment to:
```
Allow coverage to drop by a 3% when compared against the base commit.
```

davidmogar (approved):

PR #84, opened 2022-08-18
test(HACBS-798): increase test coverage on metrics/ directory

theflockers:
- adds tests for `metrics/` directory

### DIFF
--- a/metrics/release_suite_test.go
+++ b/metrics/release_suite_test.go
@@ -1,0 +1,101 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"context"
+	"fmt"
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type inputHeader struct {
+	Name string
+	Help string
+}
+
+func TestMetricsRelease(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Metrics Release Test Suite")
+}
+
+var (
+	testEnv *envtest.Environment
+	ctx     context.Context
+	cancel  context.CancelFunc
+)
+
+var _ = BeforeSuite(func() {
+	ctx, cancel = context.WithCancel(context.TODO())
+	testEnv = &envtest.Environment{}
+	cfg, err := testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	k8sManager, _ := ctrl.NewManager(cfg, ctrl.Options{
+		MetricsBindAddress: ":8081",
+		LeaderElection:     false,
+	})
+
+	go func() {
+		defer GinkgoRecover()
+		Expect(k8sManager.Start(ctx)).To(Succeed())
+	}()
+})
+
+var _ = AfterSuite(func() {
+	cancel()
+	By("tearing down the test environment")
+	err := testEnv.Stop()
+	Expect(err).NotTo(HaveOccurred())
+})
+
+// createCounterReader creates a prometheus counter type string with the given parameters to be used as input data
+// for 'strings.NewReader' in Prometheus 'client_golang' function 'testutil.CollectAndCompare'.
+func createCounterReader(header inputHeader, labels string, renderHeader bool, count int) string {
+	readerData := ""
+	if renderHeader {
+		readerData = fmt.Sprintf("# HELP %s %s\n# TYPE %s counter\n", header.Name, header.Help, header.Name)
+	}
+	readerData += fmt.Sprintf("%s{%s} %d\n", header.Name, labels, count)
+
+	return readerData
+}
+
+// createHistogramReader creates a prometheus histogram type string with the given parameters to be used as input data
+// for 'strings.NewReader' in Prometheus 'client_golang' function 'testutil.CollectAndCompare'.
+func createHistogramReader(header inputHeader, timeBuckets []string, bucketsData []int, labels string, sum float64, count int) string {
+	readerData := fmt.Sprintf("# HELP %s %s\n# TYPE %s histogram\n", header.Name, header.Help, header.Name)
+	for i, bucket := range timeBuckets {
+		readerData += fmt.Sprintf("%s_bucket{%sle=\"%s\"} %d\n", header.Name, labels, bucket, bucketsData[i])
+	}
+
+	if labels != "" {
+		readerData += fmt.Sprintf("%s_sum{%s} %f\n\n", header.Name, labels, sum)
+		readerData += fmt.Sprintf("%s_count{%s} %d\n", header.Name, labels, count)
+	} else {
+		readerData += fmt.Sprintf("%s_sum %f\n%s_count %d\n", header.Name, sum, header.Name, count)
+	}
+
+	return readerData
+}
--- a/metrics/release_test.go
+++ b/metrics/release_test.go
@@ -1,0 +1,204 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/redhat-appstudio/release-service/kcp"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/testutil"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("Metrics Release", Ordered, func() {
+	BeforeAll(func() {
+		// We need to unregister in advance otherwise it breaks with 'AlreadyRegisteredError'
+		metrics.Registry.Unregister(ReleaseAttemptRunningSeconds)
+		metrics.Registry.Unregister(ReleaseAttemptConcurrentTotal)
+		metrics.Registry.Unregister(ReleaseAttemptDurationSeconds)
+	})
+
+	var (
+		AttemptRunningSecondsHeader = inputHeader{
+			Name: "release_attempt_running_seconds",
+			Help: "Release durations from the moment the release resource was created til the release is marked as running",
+		}
+		AttemptDurationSecondsHeader = inputHeader{
+			Name: "release_attempt_duration_seconds",
+			Help: "Release durations from the moment the release PipelineRun was created til the release is marked as finished",
+		}
+		AttemptTotalHeader = inputHeader{
+			Name: "release_attempt_total",
+			Help: "Total number of releases processed by the operator",
+		}
+	)
+
+	const (
+		validReleaseReason   = "valid_release_reason"
+		invalidReleaseReason = "invalid_release_reason"
+		strategy             = "nostrategy"
+	)
+
+	var defaultNamespace = kcp.NamespaceReference{
+		Namespace: "default",
+		Workspace: "default",
+	}
+
+	Context("When RegisterNewRelease is called", func() {
+		// As we need to share metrics within the Context, we need to use "per Context" '(Before|After)All'
+		BeforeAll(func() {
+			// Mocking metrics to be able to resent data with each tests. Otherwise, we would have to take previous tests into account.
+			//
+			// 'Help' can't be overridden due to 'https://github.com/prometheus/client_golang/blob/83d56b1144a0c2eb10d399e7abbae3333bebc463/prometheus/registry.go#L314'
+			ReleaseAttemptRunningSeconds = prometheus.NewHistogram(
+				prometheus.HistogramOpts{
+					Name:    "release_attempt_running_seconds",
+					Help:    "Release durations from the moment the release resource was created til the release is marked as running",
+					Buckets: []float64{1, 5, 10, 30},
+				},
+			)
+			ReleaseAttemptConcurrentTotal = prometheus.NewGauge(
+				prometheus.GaugeOpts{
+					Name: "release_attempt_concurrent_requests",
+					Help: "Total number of concurrent release attempts",
+				},
+			)
+			metrics.Registry.MustRegister(ReleaseAttemptRunningSeconds, ReleaseAttemptConcurrentTotal)
+		})
+
+		AfterAll(func() {
+			metrics.Registry.Unregister(ReleaseAttemptRunningSeconds)
+			metrics.Registry.Unregister(ReleaseAttemptConcurrentTotal)
+		})
+
+		// Input seconds for duration of operations less or equal to the following buckets of 1, 5, 10 and 30 seconds
+		inputSeconds := []float64{1, 3, 8, 15}
+		elapsedSeconds := 0.0
+
+		It("increments the 'release_attempt_concurrent_total'.", func() {
+			creationTime := metav1.Time{}
+			for _, seconds := range inputSeconds {
+				startTime := metav1.NewTime(creationTime.Add(time.Second * time.Duration(seconds)))
+				elapsedSeconds += seconds
+				RegisterNewRelease(creationTime, &startTime)
+			}
+			Expect(testutil.ToFloat64(ReleaseAttemptConcurrentTotal)).To(Equal(float64(len(inputSeconds))))
+		})
+
+		It("registers a new observation for 'release_attempt_running_seconds' with the elapsed time from the moment"+
+			"the Release was created to when it started (Release marked as 'Running').", func() {
+			// Defined buckets for ReleaseAttemptRunningSeconds
+			timeBuckets := []string{"1", "5", "10", "30"}
+			data := []int{1, 2, 3, 4}
+			readerData := createHistogramReader(AttemptRunningSecondsHeader, timeBuckets, data, "", elapsedSeconds, len(inputSeconds))
+			Expect(testutil.CollectAndCompare(ReleaseAttemptRunningSeconds, strings.NewReader(readerData))).To(Succeed())
+		})
+	})
+
+	Context("When RegisterCompletedRelease is called", func() {
+		BeforeAll(func() {
+			ReleaseAttemptRunningSeconds = prometheus.NewHistogram(
+				prometheus.HistogramOpts{
+					Name:    "release_attempt_running_seconds",
+					Help:    "Release durations from the moment the release resource was created til the release is marked as running",
+					Buckets: []float64{1, 5, 10, 30},
+				},
+			)
+			ReleaseAttemptDurationSeconds = prometheus.NewHistogramVec(
+				prometheus.HistogramOpts{
+					Name:    "release_attempt_duration_seconds",
+					Help:    "Release durations from the moment the release PipelineRun was created til the release is marked as finished",
+					Buckets: []float64{60, 600, 1800, 3600},
+				},
+				[]string{"reason", "strategy", "succeeded", "target_namespace", "target_workspace"},
+			)
+
+			ReleaseAttemptConcurrentTotal = prometheus.NewGauge(
+				prometheus.GaugeOpts{
+					Name: "release_attempt_concurrent_requests",
+					Help: "Total number of concurrent release attempts",
+				},
+			)
+			metrics.Registry.MustRegister(ReleaseAttemptRunningSeconds, ReleaseAttemptDurationSeconds, ReleaseAttemptConcurrentTotal)
+		})
+
+		AfterAll(func() {
+			metrics.Registry.Unregister(ReleaseAttemptRunningSeconds)
+			metrics.Registry.Unregister(ReleaseAttemptDurationSeconds)
+			metrics.Registry.Unregister(ReleaseAttemptConcurrentTotal)
+		})
+
+		// Input seconds for duration of operations less or equal to the following buckets of 60, 600, 1800 and 3600 seconds
+		inputSeconds := []float64{30, 500, 1500, 3000}
+		elapsedSeconds := 0.0
+		labels := fmt.Sprintf(`reason="%s", strategy="%s", succeeded="true", target_namespace="%s", target_workspace="%s",`,
+			validReleaseReason, strategy, defaultNamespace.Workspace, defaultNamespace.Workspace)
+
+		It("increments 'ReleaseAttemptConcurrentTotal' so we can decrement it to a non-negative number in the next test", func() {
+			creationTime := metav1.Time{}
+			for _, seconds := range inputSeconds {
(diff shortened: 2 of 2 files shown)


metrics/test_utils.go
```diff
@@ -0,0 +1,108 @@
...
+	parsedMetrics := &testCompleteReleaseMetrics{}
+	ParseReleaseMetrics(string(entry), parsedMetrics)
+
+	parts := strings.Split(entry, " ")
+	if len(parts) < 2 {
+		return errors.New("required metric was not found")
+	}
+
+	releaseMetrics.Name = ""
+	// Here we wabt only the metric name, not the labels
```
johnbieren, line 67: `s/wabt/want/` (most important part of the review of course)

davidmogar (commented):
Ok, I stopped half way through as I don't think this is the proper way to do it as I commented during a meeting. Prometheus already supplies a testutil package to query metrics.

For example, the first Context block should be addressed in a similar way to this IMHO:
```
	Context("When RegisterNewRelease is called", func() {
		It("increases ReleaseAttemptConcurrentTotal", func() {
			for i := 1; i <= 5; i++ {
				RegisterNewRelease(metav1.Time{}, &metav1.Time{})
				Expect(testutil.ToFloat64(ReleaseAttemptConcurrentTotal)).To(Equal(float64(i)))
			}
			Expect(testutil.CollectAndCount(ReleaseAttemptConcurrentTotal)).To(Equal(1))
		})

		It("increments the observations of ReleaseAttemptRunningSeconds", func() {
			deltas := []int{0, 3500, 7500, 15500}
			sum := 0.0
			creationTime := metav1.Time{}
			for _, delta := range deltas {
				sum += float64(delta) / 1000
				startTime := metav1.NewTime(creationTime.Add(time.Millisecond * time.Duration(delta)))
				RegisterNewRelease(creationTime, &startTime)
			}

			// Increase buckets' data and total count by 5. This has to be done as the previous test has already called
			// the RegisterNewRelease function 5 times.
			reader := createHistogramReader(
				"release_attempt_running_seconds",
				"Release durations from the moment the release resource was created til the release is marked as running",
				[]string{"0.5", "1", "2", "3", "4", "5", "6", "7", "10", "15", "30", "+Inf"},
				[]int{6, 6, 6, 6, 7, 7, 7, 7, 8, 8, 9, 9},
				sum,
				len(deltas)+5,
			)
			Expect(testutil.CollectAndCompare(ReleaseAttemptRunningSeconds, reader)).To(Succeed())
		})
	})
```
This uses the following function, declared in `release_test.go` (`test_utils.go` file in that path is confusing to me):
```
func createHistogramReader(metricName, metricHelp string, buckets []string, bucketsData []int, totalSum float64, totalCount int) *strings.Reader {
	data := fmt.Sprintf("# HELP %s %s\n# TYPE %s histogram", metricName, metricHelp, metricName)
	for i, bucket := range buckets {
		data += "\n" + fmt.Sprintf("%s{le=\"%s\"} %d", metricName, bucket, bucketsData[i])
	}
	data += fmt.Sprintf("\n%s_sum %f\n%s_count %d\n", metricName, totalSum, metricName, totalCount)

	return strings.NewReader(data)
}
```
This has a drawback. Previous calls to the functions using the metrics will impact multiple tests. This could probably be fixed by mocking the metrics in the `metrics.go` file. I would redeclare them in a `BeforeEach` so values are cleared. At the end we are testing the functions exclusively.

theflockers:
I resolved all old comments as they don't make sense anymore as I fully refactored the code.

metrics/release_test.go
```diff
@@ -0,0 +1,118 @@
...
+
+	Context("When RegisterCompletedRelease is called", func() {
+		It(`decrements the 'release_attempt_concurrent_total' metric, increments 'release_attempt_total' and registers a new observation for 
+		   'release_attempt_duration_seconds' with the elapsed time from the moment the Release attempt started (Release marked as 'Running').`, func() {
+			// buckets for operation duration less or equal to 1, 10, 30 and 60 minutes
+			buckets := []float64{60, 600, 1800, 3600}
+			elapsedSeconds := 0.0
+			creationTime := metav1.Time{}
+			for _, bucket := range buckets {
+				startTime := metav1.NewTime(creationTime.Add(time.Millisecond * time.Duration(bucket*1000)))
```
davidmogar, line 85:
```suggestion
				startTime := metav1.NewTime(creationTime.Add(time.Millisecond * time.Duration(bucket)))
```

metrics/release_test.go
```diff
@@ -0,0 +1,118 @@
...
+			Help: "Release durations from the moment the release PipelineRun was created til the release is marked as finished",
+		}
+		AttemptTotalHeader = inputHeader{
+			Name: "release_attempt_total",
+			Help: "Total number of releases processed by the operator",
+		}
+	)
+
+	const (
+		validReleaseReason   = "valid_release_reason"
```
davidmogar, line 50: You can remove these variables and user `v1alpha1.ReleaseReason...` variables.

davidmogar (changes requested):
There's some stuff that could be improved but this is totally the way to go. Good work so far!

metrics/release_test.go
theflockers: Short comment about this one:
```
package github.com/redhat-appstudio/release-service/metrics
	imports github.com/redhat-appstudio/release-service/api/v1alpha1: import cycle not allowed in test
FAIL	github.com/redhat-appstudio/release-service/metrics [setup failed]
FAIL
```
Metrics is imported by v1alpha1. I kept the old way so far. If you have suggestions, pls come!

metrics/release_suite_test.go
```diff
@@ -0,0 +1,127 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	TYPE_GAUGE   = 1
```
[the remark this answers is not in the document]
theflockers: Constants removed.

metrics/release_suite_test.go
```diff
@@ -0,0 +1,127 @@
...
+	TYPE_COUNTER = 2
+)
+
+type inputHeader struct {
+	Name string
+	Help string
+}
+
+func createNewCounterOrGaugeReader(histHeader inputHeader, labels string, metricType int, renderHeader bool, count int) string {
```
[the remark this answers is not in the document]
theflockers: moved

metrics/release_suite_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+func createNewHistogramReader(histHeader inputHeader, timeBuckets []string, bucketsData []int, labels string, sum float64, count int) string {
+	metadata := "# HELP %s %s\n# TYPE %s histogram\n" // metadata is required to query
+	histData := "%s_bucket{le=\"%s\"} %d\n"
+	histDataWithLabels := "%s_bucket{%sle=\"%s\"} %d\n"
+	totalData := "%s_sum %f\n%s_count %d\n"
+	sumWithLabels := "%s_sum{%s} %f\n\n"
+	countWithLabels := "%s_count{%s} %d\n"
+
+	// prepares the metadata header
```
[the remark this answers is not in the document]
theflockers: I think `Prepares` or `Preparing` is better right? But thumbs up for the uppercase in the paragraph.

metrics/release_test.go
theflockers: The problem here was the following:
```
// For instance:
time.Duration(60)  == 60ns 
time.Duration(60) * time.Millisecond == 60ms 
So we still need to multiply by 1000. Although, I changed it to `time.Second * time.Duration(bucket)` instead and now it looks (and works) better.
``

File: metrics/release_test.go [hunk 7]
```diff
@@ -0,0 +1,118 @@
...
+		invalidReleaseReason = "invalid_release_reason"
+		strategy             = "nostrategy"
+	)
+
+	Context("When RegisterNewRelease is called", func() {
+		It(`increments the 'release_attempt_concurrent_total' and registers a new observation for 'release_attempt_running_seconds'
```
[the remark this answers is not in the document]
theflockers: done.

metrics/release_suite_test.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+var (
+	testEnv *envtest.Environment
+	ctx     context.Context
+	cancel  context.CancelFunc
+)
+
+var _ = BeforeSuite(func() {
+
```
davidmogar, line 49:
```suggestion
```

metrics/release_test.go
davidmogar: Right... I actually don't like this but I'm not sure we want to move those constants away from the type itself (which would be the obvious solution). Let's keep it as it is for now.

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+		strategy             = "nostrategy"
+	)
+
+	var defaultNamespace = kcp.NamespaceReference{
+		Namespace: "default",
+		Workspace: "default",
+	}
+
+	Context("When RegisterNewRelease is called", func() {
+
```
davidmogar, line 71:
```suggestion
```

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+			4 releases started 4 seconds ago
+			*/
+			data := []int{1, 2, 3, 4}
+			readerData := createNewHistogramReader(AttemptRunningSecondsHeader, timeBuckets, data, "", elapsedSeconds, len(buckets))
+			Expect(testutil.CollectAndCompare(ReleaseAttemptRunningSeconds, strings.NewReader(readerData))).To(Succeed())
+		})
+	})
+
+	Context("When RegisterCompletedRelease is called", func() {
+
```
davidmogar, line 134:
```suggestion
```

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+			timeBuckets := []string{"60", "600", "1800", "3600"}
+			// For each time bucket how many Releases completed below 4 seconds
+			data := []int{1, 2, 3, 4}
+			readerData := createNewHistogramReader(AttemptDurationSecondsHeader, timeBuckets, data, labels, elapsedSeconds, len(buckets))
+			Expect(testutil.CollectAndCompare(ReleaseAttemptDurationSeconds, strings.NewReader(readerData))).To(Succeed())
+		})
+	})
+
+	Context("When RegisterInvalidRelease", func() {
+
```
davidmogar, line 206:
```suggestion
```

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+	var defaultNamespace = kcp.NamespaceReference{
+		Namespace: "default",
+		Workspace: "default",
+	}
+
+	Context("When RegisterNewRelease is called", func() {
+
+		// As we need to share metrics within the Context, we need to use "per Context" `Before`|`After`All()
+		BeforeAll(func() {
+			// Overriding the metric definitions to simplify the tests
```
davidmogar, line 74:
```suggestion
			// Mocking metrics to be able to resent data with each tests. Otherwise, we would have to take into account previous tests.
```

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+			//
+			// `Help` can't be overriden due to `https://github.com/prometheus/client_golang/blob/83d56b1144a0c2eb10d399e7abbae3333bebc463/prometheus/registry.go#L314`
```
davidmogar, line 76: It's fine to remove this and then set Help to something like `"help"`.

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+					Buckets: []float64{1, 5, 10, 30},
+				},
+			)
+			ReleaseAttemptConcurrentTotal = prometheus.NewGauge(
+				prometheus.GaugeOpts{
+					Name: "release_attempt_concurrent_requests",
+					Help: "Total number of concurrent release attempts",
+				},
+			)
+			metrics.Registry.MustRegister(ReleaseAttemptRunningSeconds,
```
davidmogar, line 90: Same line as it's not long. Otherwise, I would write it as:
```
metrics.Registry.MustRegister(ReleaseAttemptRunningSeconds,
 
Add a suggestion, <Ctrl+g>
Add heading textAdd bold text, <Ctrl+b>Add italic text, <Ctrl+i>
Add a quote, <Ctrl+Shift+.>Add code, <Ctrl+e>Add a link, <Ctrl+k>
Add a bulleted list, <Ctrl+Shift+8>Add a numbered list, <Ctrl+Shift+7>Add a task list, <Ctrl+Shift+l>
Directly mention a user or team
Reference an issue, pull request, or discussion
Add saved reply
Same line as it's not long. Otherwise, I would write it as:
```

```
metrics.Registry.MustRegister(
        ReleaseAttemptRunningSeconds,
        ReleaseAttemptConcurrentTotal
)
```

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+				ReleaseAttemptConcurrentTotal)
+
+		})
+
+		AfterAll(func() {
+			metrics.Registry.Unregister(ReleaseAttemptRunningSeconds)
+			metrics.Registry.Unregister(ReleaseAttemptConcurrentTotal)
+
```
davidmogar, line 98:
```suggestion
```

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+		})
+
+		// Buckets for operation duration less or equal to 1, 5, 10 and 30 seconds
+		buckets := []float64{1, 5, 10, 30}
```
davidmogar, line 102: I would declare every single variable in each tests to ensure they don't get modified and their values carried over.

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+		elapsedSeconds := 0.0
+
+		It(`increments the 'release_attempt_concurrent_total'.`, func() {
```
davidmogar, line 105: It feels weird to me that you use backticks. The proper way to split long lines would be to use:
```
"fooooooooooooooooooooooooooooooooo" +
"bar"
```

This is the default behavior in IntelliJ for example.

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	Context("When RegisterNewRelease is called", func() {
+
+		// As we need to share metrics within the Context, we need to use "per Context" `Before`|`After`All()
+		BeforeAll(func() {
+			// Overriding the metric definitions to simplify the tests
+			//
+			// `Help` can't be overriden due to `https://github.com/prometheus/client_golang/blob/83d56b1144a0c2eb10d399e7abbae3333bebc463/prometheus/registry.go#L314`
+			ReleaseAttemptRunningSeconds = prometheus.NewHistogram(
```
davidmogar, line 77: Why to have them per Context and not in a global BeforeEach/AfterEach? That way every single tests could use the same metrics but without carrying over previous data.

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+			creationTime := metav1.Time{}
+			for _, bucket := range buckets {
+				startTime := metav1.NewTime(creationTime.Add(time.Second * time.Duration(bucket)))
+				elapsedSeconds += bucket
+				RegisterNewRelease(creationTime, &startTime)
+			}
+			Expect(testutil.ToFloat64(ReleaseAttemptConcurrentTotal)).To(Equal(float64(len(buckets))))
+		})
+
+		It(`registers a new observation for 'release_attempt_running_seconds' with the elapsed time from the moment
```
davidmogar, line 115: I think that as you didn't use a global BeforeEach, you fell in the same trap. Imagine that I want to run this tests in an isolated fashion. I can't. It's not even calling `RegisterNewRelease`. You are relying in the previous test.

metrics/release_test.go
```diff
@@ -0,0 +1,221 @@
...
+
+		})
+
+		// Buckets for operation duration less or equal to 1, 5, 10 and 30 seconds
+		buckets := []float64{1, 5, 10, 30}
+		elapsedSeconds := 0.0
+
+		It(`increments the 'release_attempt_concurrent_total'.`, func() {
+			creationTime := metav1.Time{}
+			for _, bucket := range buckets {
```
davidmogar, line 107: These are not buckets right? This is the actual information you pass to the function. Also, as they are so similar to the buckets defined there, I'm afraid we could get sporadic mistakes depending on how fast the execution is. I would use different values.

metrics/release_suite_test.go
```diff
@@ -0,0 +1,115 @@
...
+	if renderHeader {
+		readerData = fmt.Sprintf(metadata, histHeader.Name, histHeader.Help, histHeader.Name)
+	}
+	readerData += fmt.Sprintf(totalData, histHeader.Name, labels, count)
+
+	return readerData
+}
+
+func createNewHistogramReader(histHeader inputHeader, timeBuckets []string, bucketsData []int, labels string, sum float64, count int) string {
```
davidmogar, line 87: You can drop the `hist` prefix here and in the variables as this function is referencing histograms already. Also, the name of the function should be `createHistogramReader`. It's frequent to see `New` in Go functions when they limit themselves to create an instance of a struct.

metrics/release_suite_test.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	// Prepares the sum and count
+	if labels != "" {
+		readerData += fmt.Sprintf(sumWithLabels, histHeader.Name, labels, sum)
+		readerData += fmt.Sprintf(countWithLabels, histHeader.Name, labels, count)
+	} else {
+		readerData += fmt.Sprintf(totalData, histHeader.Name, sum, histHeader.Name, count)
+	}
+	return readerData
```
davidmogar, line 114:
```suggestion

	return readerData
```

metrics/release_suite_test.go
```diff
@@ -0,0 +1,115 @@
...
+	if renderHeader {
+		readerData = fmt.Sprintf(metadata, histHeader.Name, histHeader.Help, histHeader.Name)
+	}
+	readerData += fmt.Sprintf(totalData, histHeader.Name, labels, count)
+
+	return readerData
+}
+
+func createNewHistogramReader(histHeader inputHeader, timeBuckets []string, bucketsData []int, labels string, sum float64, count int) string {
+	metadata := "# HELP %s %s\n# TYPE %s histogram\n" // metadata is required to query
```
davidmogar, line 88: I wouldn't declare variables that are used only once. Makes the code more verbose and difficult to review in my opinion.

metrics/release_suite_test.go
```diff
@@ -0,0 +1,115 @@
...
+	// Prepares the metadata header
+	readerData := fmt.Sprintf(metadata, histHeader.Name, histHeader.Help, histHeader.Name)
+
+	// Prepares the histogram data
+	for i, bucket := range timeBuckets {
+		if labels != "" {
+			readerData += fmt.Sprintf(histDataWithLabels, histHeader.Name, labels, bucket, bucketsData[i])
+		} else {
+			readerData += fmt.Sprintf(histData, histHeader.Name, bucket, bucketsData[i])
+		}
```
davidmogar, line 104: As `histDataWithLabels` value (`"%s_bucket{%sle=\"%s\"} %d\n"`) is mostly the same to the one of `histData`, I would remove both and keep a variable defined like:
```
bucketsData := "%s_bucket{%sle=\"%s\"} %d\n"
```
Then, you just need to pass empty string for the labels part when you don't have labels.

davidmogar (commented):
I stopped the review half through as I think some of my comments apply to multiple places already.

metrics/release_test.go
theflockers: It complains that the data defined can't be different from the previous one, even if you deregister and registers again. Once you define a big string for Help, you have to stick to it. 

I guess it is because it only resets the counters but does not delete the metric.

metrics/release_test.go
theflockers: I guess the idea here was doing it per context. I cant recall now the real reason after one week (I will look into it further to recall)  but so far it looks IMHO more organized.

metrics/release_test.go
theflockers: I guess something messed up with your paste. Can you edit?

metrics/release_test.go
theflockers: This one actually does not need them. Left over probably. But I will take a look on the others.

metrics/release_suite_test.go
theflockers: variables removed.

metrics/release_suite_test.go
theflockers: done

metrics/release_test.go
theflockers: ack

metrics/release_suite_test.go
theflockers: done

metrics/release_test.go
theflockers: done

metrics/release_test.go
theflockers: done

metrics/release_test.go
theflockers: done

metrics/release_suite_test.go
theflockers: done

metrics/release_suite_test.go
theflockers: removed all.

metrics/release_test.go
theflockers: Ok, now I remember. The thing is I wanted to test the actual flow for "Registering a release"  and then "Finalizing that release" so they it ends up being dependent to each other. I am dropping this approach as I remembered we talking about isolating the tests and It makes more sense.

theflockers: @davidmogar Do you think this is really required? Asking because changing it to BeforeEach instead of the `BeforeAll` within the context made it behave totally unexpectedly. I.e.:  even after deregister and register again some metrics, it gets what is defined in the *metrics* package. If you think this way is acceptable, I would prefer to leave it as is to not have to invest more time on it.

metrics/release_suite_test.go
```diff
@@ -0,0 +1,105 @@
...
+})
+
+var _ = AfterSuite(func() {
+	cancel()
+	By("tearing down the test environment")
+	err := testEnv.Stop()
+	Expect(err).NotTo(HaveOccurred())
+})
+
+// createCounterReader creates a prometheus counter type string with the given parameteres to be used as input data
```
johnbieren, line 73: `s/parameteres/parameters/`

metrics/release_suite_test.go
```diff
@@ -0,0 +1,105 @@
...
+	readerData := ""
+	if renderHeader {
+		readerData = fmt.Sprintf("# HELP %s %s\n# TYPE %s counter\n", header.Name, header.Help, header.Name)
+	}
+	readerData += fmt.Sprintf("%s{%s} %d\n", header.Name, labels, count)
+
+	return readerData
+}
+
+// createHistogramReader creates a prometheus histogram type string with the given parameteres to be used as input data
```
johnbieren, line 85: `s/parameteres/parameters/`

metrics/release_test.go
johnbieren: either way, it is overridden not overriden

johnbieren (commented):
You two know the metrics code infinitely better than me, so when y'all agree it is good to merge, +1 from me

metrics/release_test.go
davidmogar: Let's leave it like this and revisit in the future. I strongly think that the best approach is to define them in a single place (not per context), but I also believe we should merge this already.

metrics/release_test.go
davidmogar: TF xD

I didn't see this...

Put everything in the same like as it's not long at all.
```
metrics.Registry.MustRegister(ReleaseAttemptRunningSeconds, ReleaseAttemptConcurrentTotal)
```

Otherwise, write it like:
```
metrics.Registry.MustRegister(
        ReleaseAttemptRunningSeconds,
        ReleaseAttemptConcurrentTotal
)
```

metrics/release_test.go
```diff
@@ -0,0 +1,219 @@
...
+			// Defined buckets for ReleaseAttemptRunningSeconds
+			timeBuckets := []string{"1", "5", "10", "30"}
+			/**
+			Expected value for each bucket:
+
+			1 release started 1 second ago
+			2 releases started 2 seconds ago
+			3 releases started 3 seconds ago
+			4 releases started 4 seconds ago
+			*/
```
davidmogar, line 125: I don't think this comment is that useful.

metrics/release_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+		})
+
+		AfterAll(func() {
+			metrics.Registry.Unregister(ReleaseAttemptRunningSeconds)
+			metrics.Registry.Unregister(ReleaseAttemptConcurrentTotal)
+
+		})
+
+		// input seconds for duration of operations less or equal to the following buckets of 1, 5, 10 and 30 seconds
```
davidmogar, line 100:
```suggestion
		// Input seconds for duration of operations less or equal to the following buckets of 1, 5, 10 and 30 seconds
```

davidmogar (approved):

PR #85, opened 2022-08-19
feat(HACBS-1007): split ReleaseLink into two CRDs

davidmogar:
This commit splits ReleaseLink into ReleasePlan and ReleasePlanAdmission to reduce the complexity of the added funcionallity now that we are integrating with the Environment API.


davidmogar:
Tested the code and reached the creation of the PipelineRun (I didn't test beyond that as I don't have a working setup), but it should have failed on validation. If it didn't, it was able to obtain the release strategy from the ReleaseAdmissionPlan successfully. That's also covered by the unit tests.

api/v1alpha1/release_types.go
```diff
@@ -60,6 +60,11 @@ func (rr ReleaseReason) String() string {
 	return string(rr)
 }
 
+const (
```
johnbieren, line 63: I think this makes more sense in releaseplan. A release itself will never be marked with the `auto-release` label

api/v1alpha1/releaseadmissionplan_types.go
```diff
@@ -0,0 +1,83 @@
...
+
+package v1alpha1
+
+import (
+	"github.com/redhat-appstudio/release-service/kcp"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ReleaseAdmissionPlanSpec defines the desired state of ReleaseAdmissionPlan.
+type ReleaseAdmissionPlanSpec struct {
```
johnbieren, line 25: This is not the agreed upon name. It is ReleasePlanAdmission

api/v1alpha1/releaseadmissionplan_types.go
```diff
@@ -0,0 +1,83 @@
...
+	// DisplayName is the long name of the ReleaseAdmissionPlan
+	// +optional
+	DisplayName string `json:"displayName"`
+
+	// Application is a reference to the application to be released in the managed workspace
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	Application string `json:"application"`
+
+	// Origin references where in the cluster to find the information required for a release
```
johnbieren, line 35: This description doesn't help me figure out what goes in here at all. The comment from the ticket is a much better description imo

api/v1alpha1/release_types.go
```diff
@@ -60,6 +60,11 @@ func (rr ReleaseReason) String() string {
...
+	// AutoReleaseLabel is the label name for the auto-release setting
+	AutoReleaseLabel = "release.appstudio.openshift.io/auto-release"
+)
+
 // ReleaseStatus defines the observed state of Release
```
theflockers, line 68: Any reason this one does not have the "."?

api/v1alpha1/release_types.go
```diff
@@ -23,20 +23,20 @@ import (
...
-// ReleaseSpec defines the desired state of Release
+// ReleaseSpec defines the desired state of Release.
 type ReleaseSpec struct {
 	// ApplicationSnapshot to be released
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +required
 	ApplicationSnapshot string `json:"applicationSnapshot"`
 
-	// ReleaseLink referencing the workspace where the snapshot will be released
+	// ReleasePlan to use for this particular Release
```
theflockers, line 33: Maybe _ReleasePlan used_ instead or _to use_? Not sure the correct grammar here, but this sentence sounded confusing to me.

api/v1alpha1/releaseplan_types.go
```diff
@@ -17,60 +17,55 @@ limitations under the License.
...
 	DisplayName string `json:"displayName"`
 
 	// Application is a reference to the application to be released in the managed workspace
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +required
 	Application string `json:"application"`
 
-	// Target is a reference to the workspace to establish a link with
-	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// Target references where in the cluster to find the information required for a release
```
johnbieren, line 35: Much like origin, this comment is not useful to me at all

api/v1alpha1/releaseplan_types.go
```diff
@@ -17,60 +17,55 @@ limitations under the License.
...
+type ReleasePlanStatus struct {
 }
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Display Name",type=string,priority=1,JSONPath=`.spec.displayName`
 // +kubebuilder:printcolumn:name="Application",type=string,JSONPath=`.spec.application`
-// +kubebuilder:printcolumn:name="Target",type=string,JSONPath=`.spec.target`
-// +kubebuilder:printcolumn:name="Release Strategy",type=string,JSONPath=`.spec.releaseStrategy`
+// +kubebuilder:printcolumn:name="NamespaceReference Namespace",type=string,priority=1,JSONPath=`.spec.target.namespace`
```
johnbieren, line 48: the name being Target Namespace makes way more sense to me

api/v1alpha1/releaseplan_types.go
```diff
@@ -17,60 +17,55 @@ limitations under the License.
...
+// +kubebuilder:printcolumn:name="NamespaceReference Workspace",type=string,priority=1,JSONPath=`.spec.target.workspace`
```
johnbieren, line 49: the name being Target Workspace makes way more sense to me

api/v1alpha1/releaseplan_webhook.go
```diff
@@ -0,0 +1,74 @@
+/*
```
johnbieren, line 1: I think there should be a test for circular still.
```
if workspace == ""
target.namespace cannot be releaseplan.namespace

if workspace != ""
if target.namespace == releaseplan.namespace && target.workspace == releaseplan.labels["kcp-dev"] ; throw err
```
That is not the real label name but you get the idea

api/v1alpha1/releaseadmissionplan_webhook.go
```diff
@@ -0,0 +1,74 @@
...
+func (rp *ReleaseAdmissionPlan) ValidateUpdate(old runtime.Object) error {
+	return rp.validateAutoReleaseLabel()
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type.
+func (rp *ReleaseAdmissionPlan) ValidateDelete() error {
+	return nil
+}
+
+// validateAutoReleaseLabel throws an error if the auto-release label value is set to anything besides true and false.
```
theflockers, line 66: _true *or* false_ sounds better I guess.

kcp/types.go
```diff
@@ -0,0 +1,24 @@
+package kcp
+
+import (
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// NamespaceReference holds a reference to a specific namespace within a cluster using KCP
```
johnbieren, line 7: you added periods to the other comments like this why not here?

api/v1alpha1/releaseadmissionplan_webhook_test.go
```diff
@@ -0,0 +1,129 @@
...
+			}, timeout).Should(BeTrue())
+		})
+	})
+
+	Context("When a ReleaseAdmissionPlan is created with an invalid auto-release label value", func() {
+		It("should get rejected until the value is valid", func() {
+			releaseAdmissionPlan.Labels = map[string]string{AutoReleaseLabel: "foo"}
+			err := k8sClient.Create(ctx, releaseAdmissionPlan)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("'%s' label can only be set to true or false", AutoReleaseLabel))
```
theflockers, line 80: The _It_ block says _should get rejected until the value is valid_. Does it make sense then also add an `Eventually` block as well?

api/v1alpha1/releaseadmissionplan_types.go
```diff
@@ -0,0 +1,83 @@
...
+	// +required
+	Origin kcp.NamespaceReference `json:"origin"`
+
+	// Environment defines which Environment will be used to release the application
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
```
johnbieren, line 41: do we know that this is required? We don't do anything with it. Why have it be required already?

controllers/release/release_adapter.go
```diff
@@ -197,45 +197,46 @@ func (a *Adapter) finalizeRelease() error {
...
 
-// getActiveTargetReleaseLink returns the ReleaseLink targeted by ReleaseLink in the Release being processed.
-// Only ReleaseLinks with the auto-release label set to true (or missing the label, which is treated the same as
-// having the label and it being set to true) will be searched for. If a matching ReleaseLink is not found or
-// the List operation fails, an error will be returned.
-func (a *Adapter) getActiveTargetReleaseLink() (*v1alpha1.ReleaseLink, error) {
-	releaseLink, err := a.getReleaseLink()
+// getActiveReleaseAdmissionPlan returns the ReleaseAdmissionPlan targeted by the ReleasePlan in the Release being
+// processed. Only ReleaseAdmissionPlans with the 'auto-release' label set to true (or missing the label, which is
+// treated the same as having the label and it being set to true) will be searched for.
```
johnbieren, line 202: It is weird having this line so short. Surely some of the next line can fit on it

controllers/release/release_controller.go
```diff
@@ -113,14 +113,14 @@ func SetupController(manager ctrl.Manager, log *logr.Logger) error {
 	return setupControllerWithManager(manager, NewReleaseReconciler(manager.GetClient(), log, manager.GetScheme()))
 }
 
-// setupCache adds a new index field to be able to search ReleaseLinks by target.
+// setupCache adds a new index field to be able to search ReleaseAdmissionPlans by origin.
```
johnbieren, line 116: by origin namespace.

kcp/types.go
```diff
@@ -0,0 +1,24 @@
...
+	// +required
+	Namespace string `json:"namespace"`
+
+	// Workspace references a KCP workspace within the cluster
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])(":"[a-z0-9]([-a-z0-9]*[a-z0-9]))*$
+	// +optional
+	Workspace string `json:"workspace,omitempty"`
+}
+
+type NamespacedName struct {
```
johnbieren, line 20: wait what? A NamespacedName is a NamespacedName + a Workspace? That's recursive I am not following. I also only see NamespaceReference used in the code, not NamespacedName

controllers/release/release_adapter_test.go
```diff
@@ -43,10 +44,11 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
 
-		release             *appstudiov1alpha1.Release
-		releaseStrategy     *appstudiov1alpha1.ReleaseStrategy
-		releaseLink         *appstudiov1alpha1.ReleaseLink
-		applicationSnapshot *appstudioshared.ApplicationSnapshot
+		release              *appstudiov1alpha1.Release
+		releaseStrategy      *appstudiov1alpha1.ReleaseStrategy
+		releasePlan          *appstudiov1alpha1.ReleasePlan
+		releaseAdmissionPlan *appstudiov1alpha1.ReleaseAdmissionPlan
+		applicationSnapshot  *appstudioshared.ApplicationSnapshot
```
johnbieren, line 51: no longer anal about alphabetical order huh?

api/v1alpha1/releaseadmissionplan_webhook_test.go
```diff
@@ -0,0 +1,129 @@
...
+		})
+	})
+
+	Context("When a ReleaseAdmissionPlan is created with a valid auto-release label value", func() {
+		It("shouldn't be modified", func() {
+			// Using value "true"
```
theflockers, line 86: I got to know today that ginkgo has the `By("Using value \"true\"")` definition to help to document and identify where a failure occurred in big `It` blocks. It might be useful in this case as you left this and the next comment.

controllers/release/release_adapter_test.go
```diff
@@ -299,46 +324,45 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			return activeReleaseAdmissionPlan == nil && err != nil &&
 				strings.Contains(err.Error(), "with auto-release label set to false")
 		}, time.Second*10).Should(BeTrue())
 	})
 
-	It("fails to find the target ReleaseLink when target does not match", func() {
-		patch := client.MergeFrom(releaseLink.DeepCopy())
-		releaseLink.Spec.Target = "foo"
-		Expect(k8sClient.Patch(ctx, releaseLink, patch)).Should(Succeed())
+	It("fails to find the target ReleaseAdmissionPlan when target does not match", func() {
```
johnbieren, line 358: `s/target/target namespace/`

controllers/release/release_adapter_test.go
```diff
@@ -404,16 +428,16 @@ var _ = Describe("Release Adapter", Ordered, func() {
 		}, time.Second*10).Should(BeTrue())
 	})
 
-	It("can return the releaseLink from the release", func() {
-		link, err := adapter.getReleaseLink()
+	It("can return the getReleasePlan from the release", func() {
```
johnbieren, line 431: `s/getReleasePlan/releasePlan/`

johnbieren (changes requested):

api/v1alpha1/releaseplan_webhook_test.go
```diff
@@ -0,0 +1,127 @@
...
+			}, timeout).Should(BeTrue())
+		})
+	})
+
+	Context("When a ReleasePlan is created with an invalid auto-release label value", func() {
+		It("should get rejected until the value is valid", func() {
+			releasePlan.Labels = map[string]string{AutoReleaseLabel: "foo"}
+			err := k8sClient.Create(ctx, releasePlan)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("'%s' label can only be set to true or false", AutoReleaseLabel))
```
theflockers, line 78: Also says _until_. Sounds like an `Eventually` block is required.

theflockers (commented):
I will keep looking at it. So far I left some comments.

api/v1alpha1/release_types.go
davidmogar: It's something that belongs to the release service. Although I agree, I don't want to duplicate the declaration (actually you cannot as they share the package) and I don't want only one of the two new CRDs to have it defined in their file. Alternatives?

api/v1alpha1/releaseadmissionplan_types.go
davidmogar: I'll fix it.

api/v1alpha1/releaseplan_types.go
davidmogar: Changed to `// Target references where to send the release requests`.

api/v1alpha1/releaseplan_types.go
davidmogar: Agree.

api/v1alpha1/releaseplan_types.go
davidmogar: Agree.

api/v1alpha1/releaseplan_webhook.go
davidmogar: I don't see any reason to add it. The reason why we did add it in the pass is that it could create a circle dependency in code (use the same object for both things). That cannot happen now.

kcp/types.go
davidmogar: I'll fix it.

api/v1alpha1/releaseadmissionplan_types.go
davidmogar: It will be introduced just after this change, but maybe you are right. There could be cases in which we don't want a deployment I guess...

controllers/release/release_adapter.go
davidmogar: Fixed.

controllers/release/release_controller.go
davidmogar: Right.

kcp/types.go
davidmogar: It was a goody for you to use in the future if needed. That way you can add more information. I'll remove it and add it back once is needed.

controllers/release/release_adapter_test.go
davidmogar: Not an order I added but an order I fixed :wink:

controllers/release/release_adapter_test.go
davidmogar: Right.

controllers/release/release_adapter_test.go
davidmogar: Changed to `ReleasePlan`.

api/v1alpha1/release_types.go
davidmogar: No. I'll fix it.

api/v1alpha1/release_types.go
davidmogar: Nothing used when you define a Release. It get's used later on. I think it's ok.

api/v1alpha1/releaseadmissionplan_webhook.go
davidmogar: Agree.

api/v1alpha1/releaseadmissionplan_webhook_test.go
davidmogar: Don't think so. That `until` would be a manual user action.

api/v1alpha1/releaseplan_webhook_test.go
davidmogar: Same as before. Manual user action.

api/v1alpha1/release_types.go
johnbieren: What do you mean by duplicate the declaration? Why can't we just move lines 63-66 to the releaseplan_types.go file?

api/v1alpha1/releaseadmissionplan_types.go
johnbieren: I like it a lot better now, but maybe reads better in English as "Origin references where the release requests should come from"

api/v1alpha1/releaseplan_types.go
johnbieren: I like that a lot better. I don't *love* it but I have no better suggestions

api/v1alpha1/releaseplan_webhook.go
johnbieren: We only did it before because of a circular dependency in code? I thought we wanted to prevent users from targeting themselves on a conceptual level. If we don't care about the users doing that, sure we can leave it out but I didn't think that

kcp/types.go
```diff
@@ -0,0 +1,14 @@
...
+
+// NamespaceReference holds a reference to a specific namespace within a cluster using KCP.
+type NamespaceReference struct {
+	// Namespace references a namespace within the cluster
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	Namespace string `json:"namespace"`
+
+	// Workspace references a KCP workspace within the cluster
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])(":"[a-z0-9]([-a-z0-9]*[a-z0-9]))*$
```
johnbieren, line 11: This needs to be updated I think. I can do it in my PR if you want. But we seem to need to be able to have more than one colon

johnbieren: Here is an example workspace value `root:rh-sso-6340056:jbieren-q6jqm:jbieren-managed`

davidmogar: That's actually right, but it seems that the colon doesn't need quotes. I fixed this: https://regex101.com/r/2ePsX7/1

api/v1alpha1/releaseplan_webhook.go
davidmogar: Yep, only circular dependency as you could reach to a ReleaseLink with no strategy.

api/v1alpha1/releaseadmissionplan_types.go
davidmogar: Agree.

api/v1alpha1/release_types.go
davidmogar: Because you need them in the releaseplanadmission as well.

johnbieren: Oh I follow now. I can't think of any nice solution. I agree it isn't great to have it in just one of the two new CRDs, but I think having it in the release crd is even worse

johnbieren (approved):

PR #86, opened 2022-08-19
refactor: rewrite kcp script

davidmogar:
The script was buggy and was causing problems to our CI. In addition, it was difficult to read and was using timestamps for the name prefixes instead of something better like md5 hashes.

This commit rewrites the script and updates some file names.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,7 +102,7 @@ help: ## Display this help.
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 ifeq ($(CONTROL_PLANE), kcp)
-	hack/generate-kcp-api.sh ## Generate KCP APIResourceSchema from CRDs
+	hack/generate-kcp-api.sh ## Generate KCP APIExport and APIResourceSchemas from CRDs
 endif
 
 .PHONY: generate
--- a/config/kcp/apiexport_release.yaml
+++ b/config/kcp/apiexport_release.yaml
@@ -1,13 +1,10 @@
-# This file is generated from CRDs by ./hack/generate-kcp-api.sh script.
-# Please do not modify!
-
 apiVersion: apis.kcp.dev/v1alpha1
 kind: APIExport
 metadata:
   name: release
 spec:
   latestResourceSchemas:
-    - v202208301555.releaseplanadmissions.appstudio.redhat.com
-    - v202208301555.releaseplans.appstudio.redhat.com
-    - v202208301555.releases.appstudio.redhat.com
-    - v202208301555.releasestrategies.appstudio.redhat.com
+    - md5-06f1498b5778196e58c2bc1fc88108b8.releaseplanadmissions.appstudio.redhat.com
+    - md5-9959e22acaaf6d162a03980c75b93dfe.releaseplans.appstudio.redhat.com
+    - md5-e3f55fb6efb4bca5c1b68c9813fe9a90.releases.appstudio.redhat.com
+    - md5-a1a202cb4819f0c40034030c40f8526f.releasestrategies.appstudio.redhat.com
--- a/config/kcp/apiresourceschema_release.yaml
+++ b/config/kcp/apiresourceschema_release.yaml
@@ -1,11 +1,8 @@
-# This file is generated from CRDs by ./hack/generate-kcp-api.sh script.
-# Please do not modify!
-
 apiVersion: apis.kcp.dev/v1alpha1
 kind: APIResourceSchema
 metadata:
   creationTimestamp: null
-  name: v202208301555.releaseplanadmissions.appstudio.redhat.com
+  name: md5-06f1498b5778196e58c2bc1fc88108b8.releaseplanadmissions.appstudio.redhat.com
 spec:
   group: appstudio.redhat.com
   names:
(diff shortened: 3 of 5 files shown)


davidmogar:
The files produced are not tested. Leaving that to @johnbieren :wink:

johnbieren:
@davidmogar can you rebase this (specifically because of the CRD split)? I can test this out after my KCP meeting this afternoon

davidmogar:
> @davidmogar can you rebase this (specifically because of the CRD split)? I can test this out after my KCP meeting this afternoon

It's rebased now.

theflockers (changes requested):

theflockers:
Just a very ignorant question: Shouldn't this be a job at some point instead of a script?

davidmogar:
> Just a very ignorant question: Shouldn't this be a job at some point instead of a script?

Johnny created a target in the Makefile for this.

johnbieren:
> Just a very ignorant question: Shouldn't this be a job at some point instead of a script?

what do you mean by a job?

config/kcp/api_export.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apis.kcp.dev/v1alpha1
+kind: APIExport
+metadata:
+  name: release
+spec:
+  latestResourceSchemas:
```
johnbieren, line 6: `"9959e22acaaf6d162a03980c75b93dfe.releaseplans.appstudio.redhat.com" is forbidden: [metadata.name: Invalid value: "9959e22acaaf6d162a03980c75b93dfe.releaseplans.appstudio.redhat.com": must match ^[a-z]([-a-z0-9]*[a-z0-9])?$ in front of .releaseplans.appstudio.redhat.com]`

config/kcp/api_export.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: apis.kcp.dev/v1alpha1
```
johnbieren, line 1: Can we not rename this file? Every other stream calls theirs `apiexport_<stream name>.yaml`. I don't see any benefit from renaming. Also, if KCP wants to just grab all of our apiexports, ours will at best be unclear based on the filename and at worst, conflict with others if they decide to copy our style

config/kcp/api_resource_schemas.yaml
```diff
@@ -1,11 +1,8 @@
-# This file is generated from CRDs by ./hack/generate-kcp-api.sh script.
```
johnbieren, line 1: Same rename comment

hack/generate-kcp-resources.sh
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env bash
```
johnbieren, line 1: You aren't generating all kcp resources. You are generating some kcp apis. Why the rename?

hack/generate-kcp-resources.sh
```diff
@@ -0,0 +1,52 @@
...
+set -e
+
+ROOT=$(dirname "$(dirname "$(readlink -fm "$0")")")
+CRD_DIRECTORY=$(realpath "$ROOT"/config/crd/bases)
+KCP_API_EXPORT_FILE=$(realpath "$ROOT"/config/kcp/api_export.yaml)
```
johnbieren, line 6: Same rename comment

hack/generate-kcp-resources.sh
```diff
@@ -0,0 +1,52 @@
...
+KCP_API_EXPORT_HEADER="$(cat << EOF
+apiVersion: apis.kcp.dev/v1alpha1
+kind: APIExport
+metadata:
+  name: release
+spec:
+  latestResourceSchemas:
+EOF
+)"
+KCP_API_SCHEMA_FILE=$(realpath "$ROOT"/config/kcp/api_resource_schemas.yaml)
```
johnbieren, line 16: Same rename comment

johnbieren (changes requested):

config/kcp/api_export.yaml
davidmogar: We should update our patterns. Names can start with a number https://kubernetes.io/docs/concepts/overview/working-with-objects/names.

config/kcp/api_export.yaml
davidmogar: Having the release prefix here when this is a resource that will be just deployed in the cluster under a different name doesn't make sense to me at all, but yes, can be named back to `apiexport_release.yaml`.

hack/generate-kcp-resources.sh
davidmogar: You are not generating apis. You are generating resources of two different kinds (`API*`).

johnbieren: Still not convinced changing the name that all of appstudio is using for their script that does this task is necessary

davidmogar: I think those names are not appropriate and this is a file managed by us. We keep copying stuff blindly just because that's how others did it. In any case I don't want to spend much time on this. I'll rename it back.

config/kcp/api_export.yaml
davidmogar: They would be assuming too much and writing crappy code. In any case, renaming it back.

config/kcp/api_export.yaml
davidmogar: I created a ticket for this: https://issues.redhat.com/browse/HACBS-1090

johnbieren (approved):

theflockers:
> > Just a very ignorant question: Shouldn't this be a job at some point instead of a script?
> 
> what do you mean by a job?

Something like a "fire and forget" container.

theflockers (approved):
LGTM

PR #87, opened 2022-08-22
fix(HACBS-1027): use a Red Hat provided image

davidmogar:
The current image was causing problems while building the operator as buildah doesn't use Docker Hub by default.


johnbieren (approved):

johnbieren (approved):
this should probably be something all of hacbs is synced on

davidmogar:
> this should probably be something all of hacbs is synced on

I created a PR for the integration service. The build service was already doing this (leaving apart versions).

PR #90, opened 2022-08-26
feat(HACBS-980): refactor Release status errors

johnbieren:
This commit adds a new type of Release status error called a
TargetDisabledError. This error is used to let the user know that the
connection with the target is unsuccessful. It does not make sense to
call this a ReleaseValidationError as is currently the case, since
there is nothing invalid about the Release itself.

### DIFF
--- a/controllers/release/release_adapter.go
+++ b/controllers/release/release_adapter.go
@@ -19,10 +19,12 @@ package release
 import (
 	"context"
 	"fmt"
+	"strings"
+	"time"
+
 	hasv1alpha1 "github.com/redhat-appstudio/application-service/api/v1alpha1"
 	"github.com/redhat-appstudio/release-service/gitops"
 	"github.com/redhat-appstudio/release-service/syncer"
-	"time"
 
 	"k8s.io/apimachinery/pkg/api/errors"
 
@@ -110,6 +112,18 @@ func (a *Adapter) EnsureFinalizerIsAdded() (results.OperationResult, error) {
 	return results.ContinueProcessing()
 }
 
+// EnsureReleasePlanAdmissionEnabled is an operation that will ensure that the ReleasePlanAdmission is enabled.
+// If it is not, no further operations will occur for this Release.
+func (a *Adapter) EnsureReleasePlanAdmissionEnabled() (results.OperationResult, error) {
+	_, err := a.getActiveReleasePlanAdmission()
+	if err != nil && strings.Contains(err.Error(), "auto-release label set to false") {
+		patch := client.MergeFrom(a.release.DeepCopy())
+		a.release.MarkInvalid(v1alpha1.ReleaseReasonTargetDisabledError, err.Error())
+		return results.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.release, patch))
+	}
+	return results.ContinueProcessing()
+}
+
 // EnsureReleasePipelineRunExists is an operation that will ensure that a release PipelineRun associated to the Release
 // being processed exists. Otherwise, it will create a new release PipelineRun.
 func (a *Adapter) EnsureReleasePipelineRunExists() (results.OperationResult, error) {
@@ -118,10 +132,19 @@ func (a *Adapter) EnsureReleasePipelineRunExists() (results.OperationResult, err
 		return results.RequeueWithError(err)
 	}
 
-	var releaseStrategy *v1alpha1.ReleaseStrategy
+	var (
+		releasePlanAdmission *v1alpha1.ReleasePlanAdmission
+		releaseStrategy      *v1alpha1.ReleaseStrategy
+	)
 
 	if pipelineRun == nil {
-		releaseStrategy, err = a.getReleaseStrategyFromRelease()
+		releasePlanAdmission, err = a.getActiveReleasePlanAdmission()
+		if err != nil {
+			patch := client.MergeFrom(a.release.DeepCopy())
+			a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, err.Error())
+			return results.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.release, patch))
+		}
+		releaseStrategy, err = a.getReleaseStrategy(releasePlanAdmission)
 		if err != nil {
 			patch := client.MergeFrom(a.release.DeepCopy())
 			a.release.MarkInvalid(v1alpha1.ReleaseReasonValidationError, err.Error())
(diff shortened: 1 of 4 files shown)


controllers/release/release_adapter.go
```diff
@@ -113,10 +113,19 @@ func (a *Adapter) EnsureReleasePipelineRunExists() (results.OperationResult, err
...
+		releasePlanAdmission *v1alpha1.ReleasePlanAdmission
+		releaseStrategy      *v1alpha1.ReleaseStrategy
+	)
 
 	if pipelineRun == nil {
-		releaseStrategy, err = a.getReleaseStrategyFromRelease()
+		releasePlanAdmission, err = a.getActiveReleasePlanAdmission()
+		if err != nil {
+			patch := client.MergeFrom(a.release.DeepCopy())
+			a.release.MarkInvalid(v1alpha1.ReleaseReasonTargetDisabledError, err.Error())
```
davidmogar, line 125: Didn't we discuss that we would add a new operation to rule out this case before even trying to create the Pipeline? It also required changes in the `getActiveReleasePlanAdmission` function. With this code we are falling in the same thing. We are treating two errors (missing plan or disabled release) as the same one.

davidmogar (changes requested):
We can postpone this until we discuss what changes we need in the release-service.

controllers/release/release_adapter.go
johnbieren: No we are not because the error message is different. And we don't try to create the release pipeline until later on in this function so if the dev and managed workspace aren't properly linked, we never try to create a release pipeline

johnbieren:
> We can postpone this until we discuss what changes we need in the release-service.

what? We will create changes in the release-service for the foreseeable future.. it is an active project. Not sure what you mean

controllers/release/release_adapter.go
davidmogar: One of the errors is about the ReleasePlanAdmission not being there (ReleaseReasonValidationError) while the other error is about the ReleasePlanAdmission containing a label which disables releases (ReleaseReasonTargetDisabledError).

davidmogar:
> > We can postpone this until we discuss what changes we need in the release-service.
> 
> what? We will create changes in the release-service for the foreseeable future.. it is an active project. Not sure what you mean

This PR has conflicts caused by the code introduced for the demo. What I'm saying is that we can park it to the side until we agree on what changes we want/require for that code and the merge it all together. If you want to go ahead and fix the conflicts now, that's ok for me as well.

controllers/release/release_adapter.go
johnbieren: You want to make the release as invalid if one character is off in the application name of the releaseplan? I thought the entire point of this was all the releaseplan related errors should be a new type and not be marking the release as invalid, not just the auto-release label is disabled

johnbieren: From the ticket
```
Acceptance Criteria
Releases marked with some new error types:
The managed ReleaseLink's target is NOT the dev ReleaseLink's namespace
The managed ReleaseLink and the dev ReleaseLink do not have the same application
The managed ReleaseLink is set to "auto-release": "false"
```

davidmogar: Yeah, you wrote this particular ticket. In any case, you have a comment from me there mentioning that this should be a separated operation as we discussed during our talk discussing all the changes needed. Happy to go through that again if needed, but the summary is that a missing plan is a validation error while a plan disabling release is not an error.

johnbieren: I created the ticket, then we groomed it and Scott added the acceptance criteria. So now your comment takes precedence over the team-agreed-upon acceptance criteria?

johnbieren:
@davidmogar can you tell me again why we want a new operation for this? I moved this to a draft and didn't add tests for the new operation yet because I wanted to get your thoughts on how I did it. I personally don't really like how it is done now. The cleanest way I can think to do it is to check here https://github.com/johnbieren/release-service/blob/release_status_errors/controllers/release/release_adapter.go#L145 if the error string contains something about disabled. If so, set the `TargetDisabledError`. Otherwise, we have to do the same list call twice.. once for the label and once for when we need the ReleasePlanAdmission (as is the case with this PR now)

johnbieren:
> @davidmogar can you tell me again why we want a new operation for this? I moved this to a draft and didn't add tests for the new operation yet because I wanted to get your thoughts on how I did it. I personally don't really like how it is done now. The cleanest way I can think to do it is to check here https://github.com/johnbieren/release-service/blob/release_status_errors/controllers/release/release_adapter.go#L145 if the error string contains something about disabled. If so, set the `TargetDisabledError`. Otherwise, we have to do the same list call twice.. once for the label and once for when we need the ReleasePlanAdmission (as is the case with this PR now)

@davidmogar ping

controllers/release/release_adapter_test.go
```diff
@@ -179,6 +179,31 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+		Expect(!result.RequeueRequest && !result.CancelRequest && err == nil).To(BeTrue())
+	})
+
+	It("ensures processing stops if the ReleasePlanAdmission is found to be disabled", func() {
+		// Disable the ReleasePlanAdmission
+		patch := client.MergeFrom(releasePlanAdmission.DeepCopy())
+		releasePlanAdmission.SetLabels(map[string]string{
+			appstudiov1alpha1.AutoReleaseLabel: "false",
+		})
+		Expect(k8sClient.Patch(ctx, releasePlanAdmission, patch)).Should(Succeed())
```
davidmogar, line 193: I think you need an Eventually block after this. These actions don't take effect immediately sometimes.

controllers/release/release_adapter.go
```diff
@@ -110,6 +112,20 @@ func (a *Adapter) EnsureFinalizerIsAdded() (results.OperationResult, error) {
 	return results.ContinueProcessing()
 }
 
+// EnsureReleasePlanAdmissionEnabled is an operation that will ensure that the ReleasePlanAdmission is enabled.
+// If it is not, no further operations will occur for this Release.
+func (a *Adapter) EnsureReleasePlanAdmissionEnabled() (results.OperationResult, error) {
+	_, err := a.getActiveReleasePlanAdmission()
+	if err != nil {
+		if strings.Contains(err.Error(), "auto-release label set to false") {
```
davidmogar, line 120: Concatenate both `if` with `&&`.

davidmogar (changes requested):

davidmogar (approved):

PR #91, opened 2022-08-29
feat(HACBS-970): add label appstudio.openshift.io/application: <application-name>

theflockers:
- adds the new label `appstudio.openshift.io/application: <application-name>`

### DIFF
--- a/controllers/release/release_adapter.go
+++ b/controllers/release/release_adapter.go
@@ -251,7 +251,7 @@ func (a *Adapter) createOrUpdateSnapshotEnvironmentBinding(releasePlanAdmission 
 func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStrategy, applicationSnapshot *appstudioshared.ApplicationSnapshot) (*v1beta1.PipelineRun, error) {
 	pipelineRun := tekton.NewReleasePipelineRun("release-pipelinerun", releaseStrategy.Namespace).
 		WithOwner(a.release).
-		WithReleaseLabels(a.release.Name, a.release.Namespace, a.release.GetAnnotations()[logicalcluster.AnnotationKey]).
+		WithReleaseAndApplicationLabels(a.release.Name, a.release.Namespace, a.release.GetAnnotations()[logicalcluster.AnnotationKey], applicationSnapshot.Spec.Application).
 		WithReleaseStrategy(releaseStrategy).
 		WithApplicationSnapshot(applicationSnapshot).
 		AsPipelineRun()
--- a/tekton/pipeline_run.go
+++ b/tekton/pipeline_run.go
@@ -35,6 +35,9 @@ import (
 type PipelineType string
 
 const (
+	// appstudioLabelPrefix is the prefix of the application label
+	appstudioLabelPrefix = "appstudio.openshift.io"
+
 	// pipelinesLabelPrefix is the prefix of the pipelines label
 	pipelinesLabelPrefix = "pipelines.appstudio.openshift.io"
 
@@ -46,6 +49,9 @@ const (
 )
 
 var (
+	// ApplicationNameLabel is the label used to specify the application associated with the PipelineRun
+	ApplicationNameLabel = fmt.Sprintf("%s/%s", appstudioLabelPrefix, "application")
+
 	// PipelinesTypeLabel is the label used to describe the type of pipeline
 	PipelinesTypeLabel = fmt.Sprintf("%s/%s", pipelinesLabelPrefix, "type")
 
@@ -122,13 +128,14 @@ func (r *ReleasePipelineRun) WithOwner(release *v1alpha1.Release) *ReleasePipeli
 }
 
 // WithReleaseLabels adds Release name, namespace, and workspace as labels to the release PipelineRun.
-func (r *ReleasePipelineRun) WithReleaseLabels(releaseName, releaseNamespace, releaseWorkspace string) *ReleasePipelineRun {
+func (r *ReleasePipelineRun) WithReleaseAndApplicationLabels(releaseName, releaseNamespace, releaseWorkspace string, applicationName string) *ReleasePipelineRun {
 	r.ObjectMeta.Labels = map[string]string{
 		PipelinesTypeLabel:    PipelineTypeRelease,
 		ReleaseNameLabel:      releaseName,
 		ReleaseNamespaceLabel: releaseNamespace,
 		// PipelineRun does not allow labels with : in the value, which KCP workspaces have
 		ReleaseWorkspaceLabel: strings.ReplaceAll(releaseWorkspace, ":", "__"),
+		ApplicationNameLabel:  applicationName,
 	}
 
 	return r
--- a/tekton/pipeline_run_test.go
+++ b/tekton/pipeline_run_test.go
@@ -51,6 +51,7 @@ var _ = Describe("PipelineRun", func() {
 		persistentVolumeClaim = "test-pvc"
 		serviceAccountName    = "test-service-account"
 		apiVersion            = "appstudio.redhat.com/v1alpha1"
+		applicationName       = "test-application"
 	)
 	var (
 		release                            *v1alpha1.Release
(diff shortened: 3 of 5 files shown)


controllers/release/release_adapter.go
```diff
@@ -185,14 +185,19 @@ func (a *Adapter) EnsureTargetContextIsSet() (results.OperationResult, error) {
 // will be extracted from the given ReleaseStrategy. The Release's ApplicationSnapshot will also be passed to the
 // release PipelineRun.
 func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStrategy, applicationSnapshot *appstudioshared.ApplicationSnapshot) (*v1beta1.PipelineRun, error) {
+	releasePlan, err := a.getReleasePlan()
```
johnbieren, line 188: We already have the application name in the snapshot
https://github.com/redhat-appstudio/managed-gitops/blob/main/appstudio-shared/apis/appstudio.redhat.com/v1alpha1/applicationsnapshot_types.go#L31
Why not pull it from there instead of having to retrieve the release plan?

theflockers: I don't have an actual reason for have used the one contained in the Release Plan. When I saw it, it looked  to me more appropriate than the one in the snapshot. In your opinion, why the snapshot would be better (besides it is already passed as a parameter)?

johnbieren: That is my only reason - so we don't have to pass more parameters to get information already present in other parameters

davidmogar: I agree with Johnny. We can already get the application name from the snapshot, so no need to fetch the ReleasePlan.

tekton/pipeline_run.go
```diff
@@ -123,13 +129,34 @@ func (r *ReleasePipelineRun) WithOwner(release *v1alpha1.Release) *ReleasePipeli
...
+}
+
+// WithApplicationLabel adds the Application name as label to the release PipelineRun.
+func (r *ReleasePipelineRun) WithApplicationLabel(applicationName string) *ReleasePipelineRun {
+	if len(r.ObjectMeta.Labels) == 0 {
+		r.ObjectMeta.Labels = map[string]string{
+			ApplicationNameLabel: applicationName,
+		}
+	} else {
+		r.ObjectMeta.Labels[ApplicationNameLabel] = applicationName
```
davidmogar, line 158: This line should be enough. Just make sure that an empty release already have an empty map for the labels so the other functions work.

tekton/pipeline_run.go
```diff
@@ -123,13 +129,34 @@ func (r *ReleasePipelineRun) WithOwner(release *v1alpha1.Release) *ReleasePipeli
...
 func (r *ReleasePipelineRun) WithReleaseLabels(releaseName, releaseNamespace, releaseWorkspace string) *ReleasePipelineRun {
-	r.ObjectMeta.Labels = map[string]string{
+	releaseLabels := map[string]string{
 		PipelinesTypeLabel:    PipelineTypeRelease,
 		ReleaseNameLabel:      releaseName,
 		ReleaseNamespaceLabel: releaseNamespace,
 		// PipelineRun does not allow labels with : in the value, which KCP workspaces have
 		ReleaseWorkspaceLabel: strings.ReplaceAll(releaseWorkspace, ":", "__"),
 	}
+	// makes sure previously added labels are kept
```
davidmogar, line 139: If you use a loop like you are doing, other values will always be kept. You only override keys, so I don't think you need the if block.

controllers/release/release_adapter.go
```diff
@@ -185,14 +185,19 @@ func (a *Adapter) EnsureTargetContextIsSet() (results.OperationResult, error) {
...
+	if err != nil {
+		return nil, err
+	}
 	pipelineRun := tekton.NewReleasePipelineRun("release-pipelinerun", releaseStrategy.Namespace).
 		WithOwner(a.release).
 		WithReleaseLabels(a.release.Name, a.release.Namespace, a.release.GetAnnotations()[logicalcluster.AnnotationKey]).
+		WithApplicationLabel(releasePlan.Spec.Application).
```
davidmogar, line 195: Wondering if we should rename `WithReleaseLabels` to `WithHabcsLabels` and pass here the release and the ApplicationSnapshot to generate all the labels in there. I think it would be cleaner and we would keep a single function to update labels. We wouldn't need a loop anymore as the code would be almost the same as we have right now but adding the new key/value. What do you think @johnbieren?

tekton/pipeline_run.go
theflockers: @davidmogar It seems I still need the `if` because Go (for some reason) does not allow me to add new fields to the map if it was initialized without any. It panics with `assignment to entry in nil map` as for instance:

```
	pipelineRun.ObjectMeta.Labels = make(map[string]string)
	pipelineRun.ObjectMeta.Labels["foor"] = "bar"  _// panics with assignment to entry in nil map_
```
or
```
	pipelineRun.ObjectMeta.Labels = map[string]string{}
	pipelineRun.ObjectMeta.Labels["foor"] = "bar"  _// panics with assignment to entry in nil map_
```

But it works  if I initialize it with any `key`, `value` item:

```
	pipelineRun.ObjectMeta.Labels = map[string]string{ "foo": "bar"}
	pipelineRun.ObjectMeta.Labels["a"] = "b"  _// works as expected_
```

It would still require to make sure we initialize the `Labels` field properly in  the `With...Labels` functions, hence checking if `...Labels` is _nil_ and then creating the map content with the proper items on it.

So now I agree even more with your idea to consolidate both in just one function, so we avoid that and check this just once.

controllers/release/release_adapter.go
theflockers: Why not just `WithLabels`?

tekton/pipeline_run.go
theflockers: Just an update: It was caused by the `k8sclient.Create` in the `BeforeEach` routine of the test. It seems it resets the empty `Labels` map making necessary to re-initializing it again.

controllers/release/release_adapter.go
johnbieren: If we are never going to have a release without an application snapshot, i am fine with that approach. I do not like the name `WithHacbsLabels`. I think `WithLabels` is a much better name

theflockers: Also, wouldn't be better to instead of having the strict labeling we call the function as many times we want with the labels we need? Thinking in case we need to add more labels we wouldn't need to touch the function again.

davidmogar: I'm ok with `WithLabels`, although is not as descriptive to me. In a function with this name i would expect labels to be passed. What about `WithReleaseAndApplicationLabels`?

davidmogar: I don't think we are going to need more, but yeah, could be the case. Thing is that we would have to generate and pass labels from every place we create a PR and every single PR should be a very specific set of labels, so it makes sense IMHO to define them in a single function.

johnbieren: `WithReleaseAndApplicationLabels` is fine by me. I just don't think `hacbs` is a good string to add in as we don't have it anywhere else in our code and means nothing without more context

controllers/release/release_adapter.go
theflockers: changed to get it from the snapshot.

controllers/release/release_adapter.go
theflockers: change to `WithReleaseAndApplicationLabels`

davidmogar (approved):

tekton/pipeline_run.go
```diff
@@ -35,6 +35,9 @@ import (
 type PipelineType string
 
 const (
+	// appstudioLabelPrefix is the main prefix Label for Release Service
```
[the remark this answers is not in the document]
theflockers: can you take a look again here @johnbieren ?

johnbieren (approved):

PR #94, opened 2022-09-01
feat(hacbs-773): add unit tests for api directory

parasense:
api tests
* Improve test coverage in api/v1alpha1 directory.

### DIFF
--- a/api/v1alpha1/release_types_test.go
+++ b/api/v1alpha1/release_types_test.go
@@ -1,0 +1,357 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type conditionValues struct {
+	status  metav1.ConditionStatus
+	reason  ReleaseReason
+	message string
+}
+
+var _ = Describe("Release type", func() {
+
+	var r *Release
+
+	BeforeEach(func() {
+		r = &Release{}
+		r.Status.StartTime = &metav1.Time{Time: time.Now().Add(-time.Hour * 1)}
+		r.Status.CompletionTime = &metav1.Time{Time: time.Now()}
+		r.Status.Conditions = []metav1.Condition{{}}
+	})
+
+	Context("When ReleaseReason.String method is called", func() {
+		It("should return the string representation", func() {
+			var ReleaseReasonFake ReleaseReason = "Fake"
+			Expect(ReleaseReasonFake.String()).To(Equal("Fake"))
+		})
+
+		It("should return an empty string when the ReleaseReason is empty", func() {
+			var ReleaseEmptyReason ReleaseReason = ""
+			Expect(ReleaseEmptyReason.String()).To(Equal(""))
+		})
+	})
+
+	Context("When HasBeenDeployed method is called", func() {
+		It("should return false when status.SnapshotEnvironmentBinding is nil", func() {
+			Expect(r.HasBeenDeployed()).To(BeFalse())
+		})
+
+		It("should return true when status.SnapshotEnvironmentBinding is not empty", func() {
+			r.Status.SnapshotEnvironmentBinding = "foo"
+			Expect(r.HasBeenDeployed()).To(BeTrue())
+		})
+	})
+
+	Context("When HasStarted method is called", func() {
+		It("should return false when Status.startTime is nil", func() {
+			r.Status.StartTime = nil
+			Expect(r.HasStarted()).To(BeFalse())
+		})
+
+		It("should return false when Status.startTime is zero", func() {
+			r.Status.StartTime = &metav1.Time{}
+			Expect(r.HasStarted()).To(BeFalse())
+		})
+
+		It("should return true when Status.startTime is not nil or zero", func() {
+			Expect(r.HasStarted()).To(BeTrue())
+		})
+	})
+
+	Context("When HasSucceeded method is called", func() {
+		It("should return false when condition.Status is ConditionUnknown", func() {
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:   releaseConditionType,
+				Status: metav1.ConditionUnknown,
+			}
+			Expect(r.HasSucceeded()).To(BeFalse())
+		})
+
+		It("should return true when condition.Status is ConditionTrue", func() {
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:   releaseConditionType,
+				Status: metav1.ConditionTrue,
+			}
+			Expect(r.HasSucceeded()).To(BeTrue())
+		})
+	})
+
+	Context("When IsDone method is called", func() {
+		It("should return false when condition struct is nil", func() {
+			Expect(r.IsDone()).To(BeFalse())
+		})
+
+		It("should return true when Release status is finished", func() {
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:   releaseConditionType,
+				Status: metav1.ConditionTrue,
+			}
+			Expect(r.IsDone()).To(BeTrue())
+		})
+
+		It("should return false when condition.Status is ConditionUnknown", func() {
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:   releaseConditionType,
+				Status: metav1.ConditionUnknown,
+			}
+			Expect(r.IsDone()).To(BeFalse())
+		})
+	})
+
+	Context("When MarkFailed method is called", func() {
+		It("should do nothing if the Release is finished", func() {
+			args := conditionValues{
+				reason:  ReleaseReasonValidationError,
+				message: "cow say m00",
+			}
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:    releaseConditionType,
+				Status:  metav1.ConditionTrue,
+				Reason:  ReleaseReasonTargetDisabledError.String(),
+				Message: "Testcase one message string",
+			}
+			r.MarkFailed(args.reason, args.message)
+			Expect(len(r.Status.Conditions)).To(Equal(1))
+			Expect(r.Status.Conditions[0]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionTrue),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonTargetDisabledError.String()),
+				"Message": Equal("Testcase one message string"),
+			}))
+		})
+
+		It("should register the Failed condition when the Release is not complete", func() {
+			args := conditionValues{
+				reason:  ReleaseReasonValidationError,
+				message: "what does the fox say",
+			}
+			r.Status.CompletionTime = nil
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:    releaseConditionType,
+				Status:  metav1.ConditionUnknown,
+				Reason:  ReleaseReasonTargetDisabledError.String(),
+				Message: "Testcase two message string",
+			}
+			r.MarkFailed(args.reason, args.message)
+			Expect(len(r.Status.Conditions)).To(Equal(1))
+			Expect(r.Status.Conditions[0]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionFalse),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonValidationError.String()),
+				"Message": Equal("what does the fox say"),
+			}))
+		})
+	})
+
+	Context("When MarkInvalid method is called", func() {
+		It("should register the Invalid status when the Release is not finished", func() {
+			args := conditionValues{
+				reason:  ReleaseReasonPipelineFailed,
+				message: "what does the fox say",
+			}
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:    "Fail",
+				Status:  metav1.ConditionUnknown,
+				Reason:  ReleaseReasonReleasePlanValidationError.String(),
+				Message: "message string",
+			}
+			r.MarkInvalid(args.reason, args.message)
+			Expect(len(r.Status.Conditions)).To(Equal(2))
+			Expect(r.Status.Conditions[1]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionFalse),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonPipelineFailed.String()),
+				"Message": Equal("what does the fox say"),
+			}))
+			Expect(r.Status.Conditions[0].Message).To(Equal("message string"))
+		})
+
+		It("should not register the Invalid status when the Release status is already finished", func() {
+			args := conditionValues{
+				reason:  ReleaseReasonTargetDisabledError,
+				message: "how now brown cow",
+			}
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:    releaseConditionType,
+				Status:  metav1.ConditionTrue,
+				Reason:  ReleaseReasonValidationError.String(),
+				Message: "message string",
+			}
+			r.MarkInvalid(args.reason, args.message)
+			Expect(len(r.Status.Conditions)).To(Equal(1))
+			Expect(r.Status.Conditions[0]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionTrue),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonValidationError.String()),
+				"Message": Equal("message string"),
+			}))
+		})
+	})
+
+	Context("When MarkRunning method is called", func() {
+		It("should do nothing when the Release is already running", func() {
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:    "fake type",
+				Status:  metav1.ConditionFalse,
+				Reason:  ReleaseReasonPipelineFailed.String(),
+				Message: "fake message",
+			}
+			r.MarkRunning()
+			Expect(r.Status.StartTime).ToNot(Equal(time.Time{}))
+			Expect(r.Status.StartTime).ToNot(BeNil())
+			Expect(len(r.Status.Conditions)).To(Equal(1))
+			Expect(r.Status.Conditions[0]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionFalse),
+				"Type":    Equal("fake type"),
+				"Reason":  Equal(ReleaseReasonPipelineFailed.String()),
+				"Message": Equal("fake message"),
+			}))
+		})
+
+		It("should register the Running status when the Release is not running", func() {
+			r.Status.StartTime = nil
+			r.Status.StartTime = &metav1.Time{
+				Time: time.Time{},
+			}
+			r.MarkRunning()
+			Expect(r.Status.StartTime.Time).ToNot(Equal(time.Time{}))
+			Expect(r.Status.StartTime).ToNot(BeNil())
+			Expect(len(r.Status.Conditions)).To(Equal(2))
+			Expect(r.Status.Conditions[1]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionUnknown),
+				"Reason":  Equal(ReleaseReasonRunning.String()),
+				"Message": Equal(""),
+			}))
+		})
+
+		It("should register the Running status when the Release is not running", func() {
+			r.Status.StartTime = nil
+			r.MarkRunning()
+			Expect(r.Status.StartTime.Time).ToNot(Equal(time.Time{}))
+			Expect(r.Status.StartTime).ToNot(BeNil())
+			Expect(len(r.Status.Conditions)).To(Equal(2))
+			Expect(r.Status.Conditions[1]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionUnknown),
+				"Reason":  Equal(ReleaseReasonRunning.String()),
+				"Message": Equal(""),
+			}))
+		})
+	})
+
+	Context("When MarkSucceeded method is called", func() {
+		It("should do nothing when when the Release is already successful", func() {
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:    releaseConditionType,
+				Status:  metav1.ConditionFalse,
+				Reason:  ReleaseReasonTargetDisabledError.String(),
+				Message: "lucy in the sky with diamonds",
+			}
+			rr := r.DeepCopy()
+			r.MarkSucceeded()
+			Expect(r.Status.StartTime).ToNot(BeNil())
+			Expect(r.Status.CompletionTime).ToNot(BeNil())
+			Expect(r.Status.CompletionTime.Time).To(BeTemporally(">=", rr.Status.CompletionTime.Time))
+			Expect(r.Status.CompletionTime.Time).To(BeTemporally(">", r.Status.StartTime.Time))
+			Expect(len(r.Status.Conditions)).To(Equal(1))
+			Expect(r.Status.Conditions[0]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionFalse),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonTargetDisabledError.String()),
+				"Message": Equal("lucy in the sky with diamonds"),
+			}))
+			rr = nil
+		})
+
+		It("register the Succeeded status when the Release is completed", func() {
+			r.Status.Conditions[0] = metav1.Condition{
+				Type:    "Fail",
+				Status:  metav1.ConditionUnknown,
+				Reason:  ReleaseReasonValidationError.String(),
+				Message: "all your base belong to us",
+			}
+			rr := r.DeepCopy()
+			r.MarkSucceeded()
+			Expect(r.Status.CompletionTime).ToNot(BeNil())
+			Expect(r.Status.CompletionTime.Time).To(BeTemporally(">=", rr.Status.CompletionTime.Time))
+			Expect(r.Status.CompletionTime.Time).To(BeTemporally(">", r.Status.StartTime.Time))
+			Expect(len(r.Status.Conditions)).To(Equal(2))
+			Expect(r.Status.Conditions[1]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionTrue),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonSucceeded.String()),
+				"Message": Equal(""),
+			}))
+			Expect(r.Status.Conditions[0].Message).To(Equal("all your base belong to us"))
+			rr = nil
+		})
+	})
+
+	Context("When setStatusCondition method is called", func() {
+		It("should update condition with provided arguments, and empty message", func() {
+			args := conditionValues{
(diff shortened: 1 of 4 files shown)


api/v1alpha1/release_types.go
```diff
@@ -173,7 +173,7 @@ func (r *Release) MarkInvalid(reason ReleaseReason, message string) {
 
 // MarkRunning registers the start time and changes the Succeeded condition to Unknown.
 func (r *Release) MarkRunning() {
-	if r.HasStarted() && r.Status.StartTime != nil {
+	if r.HasStarted() {
 		return
 	}
```
parasense, line 178: FYI - This was removed because it appears `r.HasStarted()` already has the same conditional! So I thought it was redundant, but to be honest this could be out of scope for this patch? I'm writing unit tests, not trying to fix bugs or whatever. Was probably feeling frisky when I clobbered that conditional statement, and could be *way* out of line..

Regardless, here is the code for `HasStarted()`
```
// HasStarted checks whether the Release has a valid start time set in its status.
func (r *Release) HasStarted() bool {
	return r.Status.StartTime != nil && !r.Status.StartTime.IsZero()
}
```
You can see it has the conditional for not `nil` and not zero...

HOWEVER!
Upon further review, it appears `time.Time.IsZero()` also checks for nil.
```
// IsZero returns true if the value is nil or time is zero.
func (t *Time) IsZero() bool {
	if t == nil {
		return true
	}
	return t.Time.IsZero()
}
```
For reference: 
https://github.com/kubernetes/apimachinery/blob/v0.24.3/pkg/apis/meta/v1/time.go#L59

I haven't tested if `IsZero()` is sufficient on it's own, but it look like it might? 
We could potentially reduce code, further simplifying the testing. :crossed_fingers:

parasense:
@davidmogar @johnbieren @theflockers @scoheb 
PTAL :smiley:

api/v1alpha1/validatedelete_test.go
```diff
@@ -0,0 +1,34 @@
...
+		}
+		Expect(releaseplan.ValidateDelete()).To(BeNil())
+	})
+
+	It("Should returns nil", func() {
+		releaseplanadmission := &Release{
+			TypeMeta:   metav1.TypeMeta{},
+			ObjectMeta: metav1.ObjectMeta{},
+		}
+		Expect(releaseplanadmission.ValidateDelete()).To(BeNil())
```
johnbieren, line 32: where do these validate functions come from? I don't see them in the *_types.go files

api/v1alpha1/validatedelete_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+	It("Should return nil", func() {
+		releaseplan := &Release{
+			TypeMeta:   metav1.TypeMeta{},
+			ObjectMeta: metav1.ObjectMeta{},
+		}
+		Expect(releaseplan.ValidateDelete()).To(BeNil())
+	})
+
+	It("Should returns nil", func() {
```
johnbieren, line 27: it may be nice to include the CRD name in the It line so it is easy to narrow down which test failed if one ever does

parasense: You mean the string "releaseplan"? 
Also, I guess the tense for "returns" should probably not be plural.

parasense: So I moved these tests into the webhook tests for each struct, and deleted the `api/v1alpha1/validatedelete_test.go`.
Also fixed them to actually test the `ValidateDelete()` for each struct.

api/v1alpha1/validatedelete_test.go
parasense: They were defined in the webhook files, so as mentioned elsewhere these three specific tests were moved into the existing webhook_test files, and updated to reflect the struct/crd they go with.

davidmogar (changes requested):
This PR is missing a rebase. There are changes in the Makefile and in the Release type that I would not expect in this PR.

In addition, as we agreed during our meeting, the `release_types_test.go` needs to be adapted so if follows the conventions used in other tests within this repository. Take a look to the `release_adapter_test.go` for inspiration.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,1043 @@
...
+			}
+
+			r.MarkFailed(args.reason, args.message)
+
+			Expect(len(r.Status.Conditions)).To(
+				Equal(1),
+				"Expected number of []Conditions to equal \"%+v\"\n",
+				1)
+
+			Expect(r.Status.Conditions[0].Status).To(
```
theflockers, line 286: You could use gomega matcher `MatchElements` to join all these `Expect`s

https://onsi.github.io/gomega/#testing-type-codestructcode

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,1043 @@
...
+type commonArgs struct {
+	status  metav1.ConditionStatus
+	reason  ReleaseReason
+	message string
+}
+
+var _ = Describe("release_types.go test collection", func() {
+
+	// ReleaseReason.String()
+	Describe("TestReleaseReason_String", func() {
```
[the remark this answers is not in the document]
davidmogar: Yes, let's use `Context` to group tests that are related with the same function/feature.

api/v1alpha1/release_types_test.go
parasense: That was a great suggestion! :+1: 
Thanks.
PTAL

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,1043 @@
...
+
+		// fake
+		It("Returns a string: fake", func() {
```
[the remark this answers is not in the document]
parasense: Okay, I went through and low-cased the first character within `It()` strings. PTAL

api/v1alpha1/release_types_test.go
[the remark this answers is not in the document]
parasense: Okay, `Describe()`, `Context()`, and `When()` are all the same EXACT thing, so it probably won't change functionality. PTAL

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,1043 @@
...
+
+type commonArgs struct {
+	status  metav1.ConditionStatus
+	reason  ReleaseReason
+	message string
+}
+
+var _ = Describe("release_types.go test collection", func() {
+
+	// ReleaseReason.String()
```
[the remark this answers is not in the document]
parasense: Okay, deleted comments. PTAL

api/v1alpha1/release_types_test.go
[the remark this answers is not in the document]
theflockers: Exactly @parasense, they are the same with different names.

Although the thing is that ginkgo's idea is to be declarative, so you literally could _read_ the tests as in reading a book (ok, exaggerated) - that's why we choose different words for each block.

It is like this example I discussed with @davidmogar some time ago. You can use both examples:
```
Expect(err).ToNot(HaveOccurred())
Expect(err).To(Succeed())
``
Although both mean the same thing, the first one IMHO tells you clearly what is expected.

File: api/v1alpha1/release_webhook_test.go [hunk 8]
```diff
@@ -75,4 +76,14 @@ var _ = Describe("Release validation webhook", func() {
 			Expect(k8sClient.Update(ctx, release)).ShouldNot(HaveOccurred())
 		})
 	})
+
+	Describe("ValidateDelete() unit test.", func() {
+		It("Release.ValidateDelete() should return nil", func() {
```
theflockers, line 81: Sorry @parasense I noticed now my comments on these last files from some reason did not went through.  Same thing about the `It` description starting on lowercase.

api/v1alpha1/releaseplan_webhook_test.go
```diff
@@ -124,4 +124,14 @@ var _ = Describe("ReleasePlan webhook", func() {
 			Expect(err.Error()).To(ContainSubstring("'%s' label can only be set to true or false", AutoReleaseLabel))
 		})
 	})
+
+	Describe("ValidateDelete() unit test.", func() {
+		It("ReleasePlan.ValidateDelete() should return nil", func() {
```
theflockers, line 129: same as above

api/v1alpha1/releaseplanadmission_webhook_test.go
```diff
@@ -126,4 +126,14 @@ var _ = Describe("ReleasePlanAdmission webhook", func() {
 			Expect(err.Error()).To(ContainSubstring("'%s' label can only be set to true or false", AutoReleaseLabel))
 		})
 	})
+
+	Describe("ValidateDelete() unit test.", func() {
+		It("ReleasePlanAdmission.ValidateDelete() should return nil", func() {
```
theflockers, line 131: same as above

theflockers (changes requested):

.gitignore
```diff
@@ -115,4 +115,14 @@ fabric.properties
 
 # Operator-sdk
 bin/*
+
```
davidmogar, line 118: A PR adding tests shouldn't change the gitignore. I haven't seen any artifact while running `make test` but if they are generated and we don't want them, we should have a Jira issue and a PR for it.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	ReleaseReasonFake ReleaseReason = "Fake"
```
davidmogar, line 30: Use the existing Reasons in the Release type, please.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+	ReleaseReasonBar  ReleaseReason = "Bar"
+)
+
+type commonArgs struct {
+	status  metav1.ConditionStatus
+	reason  ReleaseReason
+	message string
+}
+
+var _ = Describe("release_types.go test collection", func() {
```
davidmogar, line 41:
```suggestion
var _ = Describe("Release type", func() {
```

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+
+	Context("ReleaseReason.String()", func() {
```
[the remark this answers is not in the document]
davidmogar: This comment applies to every single `Context` in this PR.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+
+		It("should return an empty string", func() {
+			var ReleaseEmptyReason ReleaseReason = ""
+			Expect(ReleaseEmptyReason.String()).To(Equal(""),
+				"Expected \"%+v\" to equal \"%+v\"", ReleaseEmptyReason.String(), "")
+		})
+	})
+
+	Context("HasBeenDeployed()", func() {
+		It("should return false because status.SnapshotEnvironmentBinding is nil", func() {
```
davidmogar, line 67:
```suggestion
		It("should return false when Status.SnapshotEnvironmentBinding is nil", func() {
```

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+			r := &Release{}
+			Expect(r.HasBeenDeployed()).To(BeFalse(),
+				"Expected to get: %+v ; Actually Got: %+v\n", false, true)
```
davidmogar, line 70: I don't think we need custom errors at all. `Expect(r.HasBeenDeployed()).To(BeFalse()) is more than enough. The error reported by Ginkgo is similar.

This applies to the other `Expects`.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+		})
+
+		It("should return false because status.SnapshotEnvironmentBinding is empty", func() {
```
davidmogar, line 73: This test and the previous one are exactly the same as "" is the nil value for strings. It should be removed.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+			r := &Release{
+				Status: ReleaseStatus{
+					SnapshotEnvironmentBinding: "", // empty
+				},
+			}
+			Expect(r.HasBeenDeployed()).To(BeFalse(),
+				"Expected to get: %+v ; Actually Got: %+v\n", false, true)
+		})
+
+		It("should return true because status.SnapshotEnvironmentBinding is not empty", func() {
```
davidmogar, line 83:
```suggestion
		It("should return true when Status.SnapshotEnvironmentBinding is not empty", func() {
```

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+					SnapshotEnvironmentBinding: "foo",
+				},
+			}
+			Expect(r.HasBeenDeployed()).To(BeTrue(),
+				"Expected to get: %+v ; Actually Got: %+v\n", true, false)
+		})
+	})
+
+	Context("HasStarted()", func() {
+		It("should return false because Status.startTime is nil", func() {
```
davidmogar, line 95:
```suggestion
		It("should return false when Status.startTime is nil", func() {
```

This comment applies to the other tests as well. Basically change `because` by `when`.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+			r := &Release{}
+			Expect(r.HasStarted()).To(BeFalse(),
+				"Expected to get: %+v ; Actually Got: %+v\n", false, true)
+		})
+
+		It("should return false because Status.startTime is zero", func() {
+			r := &Release{
+				Status: ReleaseStatus{
+					StartTime: &metav1.Time{}, // 0001-01-01 00:00:00 +0000 UTC
```
davidmogar, line 104: This comment is not needed in my opinion.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+						},
+					},
+				},
+			}
+			Expect(r.HasStarted()).To(BeFalse(),
+				"Expected to get: %+v ; Actually Got: %+v\n", false, true)
+		})
+
+		It("should return false because Status.startTime is nil", func() {
+			r := &Release{
```
davidmogar, line 121: You could create a single Release in each context and modify the startTime in each `It` block. We don't like to influence other tests passing over the same objects, but within a `Context` is fine. You could even restore the default state at the end of each `It`. This would reduce the number of lines as this file has become quite big.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+				},
+			}
+			Expect(r.HasStarted()).To(BeTrue(),
+				"Expected to get: %+v ; Actually Got: %+v\n", true, false)
+		})
+	})
+
+	Context("HasSucceeded()", func() {
+		It("should return false because condition.Status is ConditionUnknown", func() {
+			r := &Release{
```
davidmogar, line 144: Same comment as before about reusing the same object.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+						},
+					},
+				},
+			}
+			r.MarkFailed(args.reason, args.message)
+			Expect(len(r.Status.Conditions)).To(
+				Equal(1),
+				"Expected number of []Conditions to equal \"%+v\"\n",
+				1)
+			Expect(r.Status.Conditions[0]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
```
davidmogar, line 241: Great change!

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+				"Status":  Equal(metav1.ConditionTrue),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonBar.String()),
+				"Message": Equal("Testcase one message string"),
+			}), "Expect condition struct fields to match expected fields")
+		})
+
+		It("should not NOOP because CompletionTime is nil, and IsDone returns false", func() {
```
davidmogar, line 249: This description sound a bit confusing to me.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+				"Status":  Equal(metav1.ConditionFalse),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonBar.String()),
+				"Message": Equal("what does the fox say"),
+			}), "Expect condition struct fields to match expected fields")
+		})
+	})
+
+	Context("MarkInvalid()", func() {
+		It("should not NOOP because IsDone() returns false", func() {
```
davidmogar, line 285: Not really understanding the description. As a side note, better than putting `because IsDone() returns false` I would put `because is not done`.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+				"Reason":  Equal(ReleaseReasonBar.String()),
+				"Message": Equal("what does the fox say"),
+			}), "Expect condition struct fields to match expected fields")
+			Expect(r.Status.Conditions[0].Message).To(
+				Equal("message string"),
+				"Expected condition.Message to equal \"%+v\".\n",
+				"message string")
+		})
+
+		It("should NOOP because method IsDone() returns true", func() {
```
davidmogar, line 325: Same as before. I'd try to write `NOOP`.

davidmogar (changes requested):
I added a bunch of comments that affect multiple lines. They will reduce the "complexity" of the file, so we can review again once it's done.

Overall the change to use Ginkgo is good. It just need some changes to adapt to how we have been writing the tests.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+			args := commonArgs{
+				reason:  ReleaseReasonBar,
+				message: "how now brown cow",
+			}
+			r := &Release{
```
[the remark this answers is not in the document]
theflockers: Totally. I would use a `BeforeAll` and only change the required parameter/condition inside the test.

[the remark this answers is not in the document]
parasense: PTAL

api/v1alpha1/release_types_test.go
parasense: PTAL

api/v1alpha1/release_types_test.go
parasense: PTAL

.gitignore
parasense: PTAL

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,671 @@
...
+type commonArgs struct {
+	status  metav1.ConditionStatus
+	reason  ReleaseReason
+	message string
+}
+
+var _ = Describe("release_types.go test collection", func() {
+
+	Context("ReleaseReason.String()", func() {
+		It("should return a string: fake", func() {
```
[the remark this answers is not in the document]
parasense: PTAL

api/v1alpha1/release_types_test.go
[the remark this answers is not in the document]
parasense: PTAL

api/v1alpha1/releaseplanadmission_webhook_test.go
parasense: PTAL

api/v1alpha1/releaseplan_webhook_test.go
parasense: PTAL

api/v1alpha1/release_types_test.go
parasense: Okay, I changed the conjunction from "because" to "when", and that probably makes more sense. PTAL

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+type commonArgs struct {
+	status  metav1.ConditionStatus
+	reason  ReleaseReason
+	message string
+}
+
+var _ = Describe("Release type", func() {
+
+	var r *Release
+	JustBeforeEach(func() {
```
theflockers, line 38: If you are not intended to use `BeforeEach` inside the `Context` blocks as well you don't need `JustBeforeEach`. Unless you are planning to do this later on.

api/v1alpha1/release_types_test.go
parasense: This is probably outdated?
The structure is now setup with Ginkgo's `BeforeEach()` routine, and that dramatically reduced the number of lines. Therefore I'm not sure it's worth taking a step further to setup the structure in the context, and ensuring the structure is appropriate prior to each evaluation. However, I could be persuaded, for example if we ever wanted to optimize for a massive stress-test where we run up against garbage collection.

api/v1alpha1/release_types_test.go
parasense: I agree it's a bit wordy. I've tried to clarify a little, and added a truth table help.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+				CompletionTime: &metav1.Time{
+					Time: time.Now(),
+				},
+				Conditions: []metav1.Condition{{}},
+			},
+		}
+	})
+
+	Context("When ReleaseReason.String method is called", func() {
+		It("should return a string: Fake", func() {
```
davidmogar, line 54:
```suggestion
		It("should return the string representation", func() {
```

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+			var ReleaseReasonFake ReleaseReason = "Fake"
+			Expect(ReleaseReasonFake.String()).To(Equal("Fake"))
+		})
+
+		It("should return an empty string", func() {
```
davidmogar, line 59:
```suggestion
		It("should return an empty string when the ReleaseReason is empty", func() {
```

api/v1alpha1/release_types_test.go
davidmogar: This still says `because`.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+	})
+
+	Context("When HasStarted method is called", func() {
+		It("should return false when Status.startTime is nil", func() {
+			r.Status.StartTime = nil
+			Expect(r.HasStarted()).To(BeFalse())
+		})
+
+		It("should return false when Status.startTime is zero", func() {
+			r.Status.StartTime = &metav1.Time{} // 0001-01-01 00:00:00 +0000 UTC
```
davidmogar, line 88: The comment is not needed.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+			|--------------------------------------------------------------------|
+
+			When no NOOP:
+			* expect zero or nil CompletionTime to be set to Now().
+			* expect setStatusConditionWithMessage() to append a new condition.
+			* * expect condition.Status: ConditionFalse
+			* * expect condition.Reason: to match the reason arg
+			* * expect condition.Message: to match the message arg
+		*/
+		It("should NOOP when IsDone() is true, and Status.CompletionTime is not nil", func() {
```
davidmogar, line 159: We don't need that table on the top or any kind of comment. Let's rewrite this with something like:
```suggestion
		It("should do nothing if the Release has finished", func() {
```

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+			Expect(len(r.Status.Conditions)).To(Equal(1))
+			Expect(r.Status.Conditions[0]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionTrue),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonTargetDisabledError.String()),
+				"Message": Equal("Testcase one message string"),
+			}))
+		})
+
+		It("should not NOOP when CompletionTime is nil, and IsDone returns false", func() {
```
davidmogar, line 180:
```suggestion
		It("should register a failed condition if the CompletionTime is nil", func() {
```
Also, we would need a similar test for when IsDone is true so we test both scenarios.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+			| ConditionTrue    |  true    |  yes |
+			|------------------------------------|
+
+			When no NOOP:
+			* expect setStatusConditionWithMessage() to append a new condition.
+			* * expect condition.Status: ConditionFalse
+			* * expect condition.Reason: to match the reason arg
+			* * expect condition.Message: to match the message arg
+		*/
+		It("should not NOOP when IsDone() returns false", func() {
```
davidmogar, line 221: No need for a table. Let's fix this in the same way than the previous ones:
```suggestion
		It("should register an Invalid status if the Release hasn't finished", func() {
```

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+			Expect(r.Status.Conditions[1]).To(MatchFields(IgnoreMissing|IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionFalse),
+				"Type":    Equal(releaseConditionType),
+				"Reason":  Equal(ReleaseReasonPipelineFailed.String()),
+				"Message": Equal("what does the fox say"),
+			}))
+			Expect(r.Status.Conditions[0].Message).To(Equal("message string"))
+		})
+
+		It("should NOOP when method IsDone() returns true", func() {
```
davidmogar, line 243:
```suggestion
		It("should do nothing if the Release has finished", func() {
```

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,446 @@
...
+			|---------------------------------------------|
+
+			When no NOOP:
+			* expect zero or nil StartTime to be set to Now().
+			* expect setStatusCondition() to append a new condition.
+			* * expect condition.Status: ConditionUnknown
+			* * expect condition.Reason: ReleaseReasonRunning
+			* * expect condition.Message: ""
+		*/
+		It("should NOOP when Status.StartTime is not nil and not zero", func() {
```
davidmogar, line 284: Same comments as above.

davidmogar (changes requested):

api/v1alpha1/release_types_test.go
parasense: The decision to use `JustBeforeEach` was on purpose, to imperatively set things as close as possible to `It()` nodes.
It could be changed to `BeforeEach` if you feel strongly against `JustBeforeEach`? There was no plan to use `BeforeEach`.

That being said, I could be persuaded to set `Release` structures at the `Context` scope, and use `JustBeforeEach` to restore the structure fields prior to each test. That way we optimize for memory, in case somebody ever wants to run a benchmark or similar stress test, the runtime wouldn't thrash GC so much. To use a cliche, I'm not sure _the wick is worth the game_, but I'm willing to burn that candle if we want? What do you think?

api/v1alpha1/release_types_test.go
parasense: Gah!  :man_facepalming: 
Thanks, apparently my find --> replace skipped that one

api/v1alpha1/release_types_test.go
parasense: Moved to `BeforeEach`. Thanks. :+1:

api/v1alpha1/release_types_test.go
parasense: oopsie, I missed this one before. 
The redundant test has been removed. Thanks! :+1:

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,357 @@
...
+var _ = Describe("Release type", func() {
+
+	var r *Release
+	BeforeEach(func() {
+		r = &Release{}
+		r.Status.StartTime = &metav1.Time{Time: time.Now().Add(-time.Hour * 1)}
+		r.Status.CompletionTime = &metav1.Time{Time: time.Now()}
+		r.Status.Conditions = []metav1.Condition{{}}
+	})
+	AfterEach(func() { r = nil })
```
davidmogar, line 44: This line is not needed.

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,357 @@
...
+type commonArgs struct {
+	status  metav1.ConditionStatus
+	reason  ReleaseReason
+	message string
+}
+
+var _ = Describe("Release type", func() {
+
+	var r *Release
+	BeforeEach(func() {
```
davidmogar, line 38: Could you add an empty line just before?

api/v1alpha1/release_types_test.go
```diff
@@ -0,0 +1,357 @@
...
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type commonArgs struct {
```
davidmogar, line 29: Rename to `conditionValues` or something similar to reflect these are actually the conditions values we expect to see.

davidmogar (changes requested):
Huge and positive change. Just some minor comments.

theflockers (approved):
It Looks good to me as well!

PR #95, opened 2022-09-01
feat: add KCP sample yamls

johnbieren:
This commit adds some resource yaml files to the config/samples/kcp
directory. These are for showing how to have the controller
reconcile a simple release in a KCP environment that includes both
a dev and managed workspace. They do not use permission claims,
but may be updated to at a later date.

### DIFF
--- a/config/samples/kcp/dev_workspace/apiexport.yaml
+++ b/config/samples/kcp/dev_workspace/apiexport.yaml
@@ -1,0 +1,12 @@
+apiVersion: apis.kcp.dev/v1alpha1
+kind: APIExport
+metadata:
+  name: release
+spec:
+  latestResourceSchemas:
+    - rev-1492075.releaseplanadmissions.appstudio.redhat.com
+    - rev-1492080.applicationsnapshots.appstudio.redhat.com
+    - rev-1492087.pipelineruns.tekton.dev
+    - rev-1492107.releases.appstudio.redhat.com
+    - rev-1492112.releaseplans.appstudio.redhat.com
+    - rev-1492119.releasestrategies.appstudio.redhat.com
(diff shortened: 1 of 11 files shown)


johnbieren:
@hongweiliu17 FYI

davidmogar (commented):
If we go for the changes in the script that generates the yamls, we should use similar names from there.

johnbieren:
> If we go for the changes in the script that generates the yamls, we should use similar names from there.

what part of this pr are you referring to with this comment? These resources are using apiresourceschema from the syncer, not the ones in our repo now, that is why their names are different than both the other PR and what we already have committed. Unless you are referring to something else I can't think of

davidmogar:
> > If we go for the changes in the script that generates the yamls, we should use similar names from there.
> 
> what part of this pr are you referring to with this comment? These resources are using apiresourceschema from the syncer, not the ones in our repo now, that is why their names are different than both the other PR and what we already have committed. Unless you are referring to something else I can't think of

`rev-1492075.releaseplanadmissions.appstudio.redhat.com`. Those names are not generated by the script. That's what I meant.

johnbieren:
> > > If we go for the changes in the script that generates the yamls, we should use similar names from there.
> > 
> > 
> > what part of this pr are you referring to with this comment? These resources are using apiresourceschema from the syncer, not the ones in our repo now, that is why their names are different than both the other PR and what we already have committed. Unless you are referring to something else I can't think of
> 
> `rev-1492075.releaseplanadmissions.appstudio.redhat.com`. Those names are not generated by the script. That's what I meant.

Correct. Those are generated by the syncer. So the other PR does not affect them (and never will). I mentioned it in my recording, but until I get permission claims figured out, the only way I know to get things to work is to have the syncer provide all of the CRDs for me instead of applying my own apiresourceschema. So the names starting with `rev-` are based on what the syncer gives us

PR #96, opened 2022-09-02
fix: auto guess the fs path for pseudo versioned modules

theflockers:
fix: auto guess the fs path for pseudo versioned modules
    
- adds `GetRelativeDependencyPath` function to return the fs path of
  pseudo-versioned modules required by the test suites.
- adds unit tests
    
### DIFF
--- a/controllers/release/release_suite_test.go
+++ b/controllers/release/release_suite_test.go
@@ -18,10 +18,11 @@ package release
 
 import (
 	"context"
-	hasv1alpha1 "github.com/redhat-appstudio/application-service/api/v1alpha1"
 	"go/build"
 	"path/filepath"
 	"testing"
+
+	hasv1alpha1 "github.com/redhat-appstudio/application-service/api/v1alpha1"
 
 	"k8s.io/client-go/rest"
 
@@ -34,6 +35,7 @@ import (
 
 	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
 	appstudiov1alpha1 "github.com/redhat-appstudio/release-service/api/v1alpha1"
+	appstudiotest "github.com/redhat-appstudio/release-service/test"
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	clientsetscheme "k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -73,13 +75,11 @@ var _ = BeforeSuite(func() {
 			),
 			filepath.Join(
 				build.Default.GOPATH,
-				"pkg", "mod", "github.com", "redhat-appstudio",
-				"application-service@v0.0.0-20220310134421-5864c77a3d17", "config", "crd", "bases",
+				"pkg", "mod", appstudiotest.GetRelativeDependencyPath("application-service"), "config", "crd", "bases",
 			),
 			filepath.Join(
 				build.Default.GOPATH,
-				"pkg", "mod", "github.com", "redhat-appstudio", "managed-gitops",
-				"appstudio-shared@v0.0.0-20220603115212-1fb4d804a8c2", "config", "crd", "bases",
+				"pkg", "mod", appstudiotest.GetRelativeDependencyPath("appstudio-shared"), "config", "crd", "bases",
 			),
 		},
 		ErrorIfCRDPathMissing: true,
--- a/test/paths.go
+++ b/test/paths.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package test
+
+import (
+	"bufio"
+	"errors"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// GetRelativeDependencyPath returns the relative path for a given module or an empty string on error.
+func GetRelativeDependencyPath(moduleName string) string {
+	var crdPath string
+	currentWorkDir, _ := os.Getwd()
+
+	goModFilepath, err := FindGoModFilepath(currentWorkDir)
+	if err != nil {
+		return ""
+	}
+
+	goModReadFile, err := os.Open(filepath.Clean(goModFilepath))
+	if err != nil {
+		return ""
+	}
+	defer func() {
+		err := goModReadFile.Close()
+		if err != nil {
+			panic(err)
+		}
+	}()
+
+	scanner := bufio.NewScanner(goModReadFile)
+	for scanner.Scan() {
+		if strings.Contains(scanner.Text(), moduleName) {
+			crdPath = strings.Trim(string(scanner.Text()), "\t")
+			break
+		}
+	}
+
+	return strings.ReplaceAll(crdPath, " ", "@")
+}
+
+// FindGoModFilepath returns the go.mod file system path for the current project or empty string and an error when file is not found.
+func FindGoModFilepath(currentWorkDir string) (string, error) {
+	for {
+		currentWorkDir = filepath.Dir(currentWorkDir)
+		if currentWorkDir != "/" {
+			goModFilepath := currentWorkDir + "/go.mod"
+			_, err := os.Stat(goModFilepath)
+			if err == nil {
+				return goModFilepath, nil
+			}
+		} else {
+			return "", errors.New("go.mod file not found")
+		}
+	}
+}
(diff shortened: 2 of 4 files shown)


theflockers:
I also found the same issue with another two test_suite files, but don't want to duplicate the code. Where should be the proper place to keep this function? Or any other better idea to fix this better?

controllers/release/release_suite_test.go
```diff
@@ -115,3 +118,26 @@ var _ = AfterSuite(func() {
...
+// for pseudo versioned modules.
+func GetModuleCRDPath(goModPath string, moduleName string) string {
+	var CRDPath string
+	readFile, err := os.Open(goModPath)
+	if err != nil {
+		panic(err)
+	}
+	scanner := bufio.NewScanner(readFile)
+	for scanner.Scan() {
+		re := regexp.MustCompile(".*" + moduleName + ".*")
```
davidmogar, line 132: Wouldn't be this a simple `strings.Contains`?

controllers/release/release_suite_test.go
```diff
@@ -115,3 +118,26 @@ var _ = AfterSuite(func() {
...
+		match := re.Find([]byte(scanner.Text()))
+		if match != nil {
+			CRDPath = strings.Trim(string(match), "\t")
+			break
+		}
+	}
+	readFile.Close()
```
davidmogar, line 139: You can use `defer readFile.Close()` so it's closed when you leave the function.

controllers/release/release_suite_test.go
```diff
@@ -115,3 +118,26 @@ var _ = AfterSuite(func() {
...
+	CRDPath = strings.ReplaceAll(CRDPath, " ", "@")
```
davidmogar, line 140: Don't set the value just to return it afterwards. Also, this behavior should be documented. If not match is found or the scan doesn't work, you are just replacing text. Probably it's best to return `strings.ReplaceAll(strings.Trim(string(match), "\t")` directly inside the for loop. It would avoid you to having to use break and would move this line within the loop. In case of reaching this point, we would return `""` directly which is a healthier return IMHO. In addition, as we are moving to using defer, you don't need anything else after the loop.

controllers/release/release_suite_test.go
```diff
@@ -115,3 +118,26 @@ var _ = AfterSuite(func() {
 	err := testEnv.Stop()
 	Expect(err).NotTo(HaveOccurred())
 })
+
+// GetModuleCRDPath return the filesystem path after GOPATH/pkg/mod
+// for pseudo versioned modules.
+func GetModuleCRDPath(goModPath string, moduleName string) string {
```
davidmogar, line 124: Can we make `goModPath` a constant based on `build.Default.GoPath`?

Also, move this function to a new package called `test` and a file called `paths.go`. The name of the function should probably be `GetRelativeDependencyPath` as this function has no relationship with CRDs. You would pass a `dependency string` which could be a full name or a substring (document it).

davidmogar (changes requested):
Great and smart proposal overall. It just needs some cleaning.

controllers/release/release_suite_test.go
theflockers: the `goModPath` it the path to `go.mod`. The `build.Default.GoPath` points to `GOPATH`. Found that there is the `GOMOD` env. I will use this one instead.

theflockers: @davidmogar. The variable `GOMOD` that I was willing to use is not available unless you load the `go env`. So we have two options, or use the second function I added or add something like `source <$(go env)` so the `GOMOD` variable is availble, making it easyer to get the `go.mod` file path.

controllers/release/release_suite_test.go
theflockers: done

controllers/release/release_suite_test.go
theflockers: fixe.

controllers/release/release_suite_test.go
theflockers: fixed

controllers/release/release_suite_test.go
theflockers: Moved already to Makefile.

test/paths.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// GetRelativeDependencyPath returns the relative path for the specified module name
+//
+// moduleName string - module name or substring
+//
+// return string
+func GetRelativeDependencyPath(moduleName string) string {
+	var CRDPath string
+	readFile, err := os.Open(os.Getenv("GOMOD"))
```
davidmogar, line 32: I think we could use a relative path here instead of using GOMOD just in case is not defined in the CI environment for some reason. Your thoughts?

test/paths.go
```diff
@@ -0,0 +1,48 @@
...
+	"strings"
+)
+
+// GetRelativeDependencyPath returns the relative path for the specified module name
+//
+// moduleName string - module name or substring
+//
+// return string
+func GetRelativeDependencyPath(moduleName string) string {
+	var CRDPath string
```
davidmogar, line 31: Rename it to `crdPath` to be consistent with Go variables naming.

test/paths.go
theflockers: The problem with the relative path is that is varies due to that for each tests it `chdir` to its directory. For  `metrics` and `tekton` for instance `../go.mod` is enough but for `controllers/release` and `controllers/results` it is `../../go.mod`.

On commit before this  (which I removed), I added a function to check both places (`../`  `../../`) and returns once `go.mod` is found, but thought that _maybe_ by  setting the var it would require less code and would be more clear from where we are getting it.

Each of them has their advantages. I guess checking `../` and `../../` has the advantage that we don't need to set another variable that we might at some point forget why we need it. The other one is just more straight forward to understand.

theflockers: I will add it to the next commit so you can check it out then we decide what is best.

test/paths.go
theflockers: fixed

test/paths.go
davidmogar: The relative path would be taken always relative to this function (`GetRelativeDependencyPath`), not the caller. I don't think you need the new function at all.

test/paths.go
```diff
@@ -0,0 +1,67 @@
...
+
+package test
+
+import (
+	"bufio"
+	"os"
+	"strings"
+)
+
+// GetRelativeDependencyPath returns the relative path for the specified module name
```
davidmogar, line 25: Comments for functions in go are composed in the following way:
```
// <Name of the function> <what it does>.
```
Returns, parameters, packages... all of this is generated automatically, so you don't have to add it.

test/paths.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+// GetGoModFilepath returns the go.mod file location for the current project or empty string when file is not found
+//
+// return string
+func GetGoModFilepath() string {
+	var parentDir string
+	currentWorkDir, _ := os.Getwd()
+
+	for _, dir := range []string{"..", ".."} {
```
davidmogar, line 57: We can keep the function (check my other comment about relativity of the path), but I don't like that it's limited to two levels. I would use an infinite loop and walk backwards using filepath.Dir until you find the file or until the function returns `.` (I think this is the return when you are in the root, but double check).

davidmogar (approved):

controllers/release/release_suite_test.go
```diff
@@ -34,6 +35,7 @@ import (
 
 	appstudioshared "github.com/redhat-appstudio/managed-gitops/appstudio-shared/apis/appstudio.redhat.com/v1alpha1"
 	appstudiov1alpha1 "github.com/redhat-appstudio/release-service/api/v1alpha1"
+	appstudiotest "github.com/redhat-appstudio/release-service/test"
```
johnbieren, line 38: I think `releasetest` could've been a better name for it, but because of the line before I think we keep it as is for consistency's sake. I think because there is an appstudio team, my brain trips up on it thinking appstudio is not us

test/paths.go
```diff
@@ -0,0 +1,73 @@
...
+		if strings.Contains(scanner.Text(), moduleName) {
+			crdPath = strings.Trim(string(scanner.Text()), "\t")
+			break
+		}
+	}
+
+	return strings.ReplaceAll(crdPath, " ", "@")
+}
+
+// GetGoModFilepath returns the go.mod fs filepath for the current project or empty string and an error when file is not found.
```
johnbieren, line 59: name of the function wrong

test/paths_test.go
```diff
@@ -0,0 +1,66 @@
...
+		})
+
+		It("return an empty string and an error when the given directory does not exist", func() {
+			goModFilepath, err := FindGoModFilepath("/nonexistent/dir")
+			Expect(goModFilepath).Should(BeEmpty())
+			Expect(err).Should(And(HaveOccurred()), Equal("go.mod file not found"))
+		})
+	})
+
+	Context("When GetRelativeDependencyPath is called", func() {
```
johnbieren, line 55: Since I am already requesting a change, can we have the functions be in the same order in both paths.go and paths_test.go?

johnbieren (changes requested):

johnbieren (approved):

test/paths.go
theflockers: fixed.

test/paths_test.go
theflockers: done.

PR #97, opened 2022-09-13
fix(HACBS-1037): fix typos introduced with bindings feature

davidmogar:
Because of the upcoming demo, we had to push some code to main adding support for GitOpsDeployments. This change was not properly reviewed, so this commit address all the problems found.


johnbieren (approved):

controllers/release/cache.go
```diff
@@ -45,7 +45,7 @@ func SetupReleasePlanAdmissionCache(mgr ctrl.Manager) error {
 		"spec.origin.namespace", releasePlanAdmissionIndexFunc)
 }
 
-// SetupSnapshotEnvironmentBindingCache adds a new index field to be able to search SnapshotEnvironmentBinding by environment.
+// SetupSnapshotEnvironmentBindingCache adds a new index field to be able to search SnapshotEnvironmentBindings by environment.
```
scoheb, line 48: ApplicationSnapshotEnvironmentBinding?

davidmogar: The resource is getting renamed (Application prefix is getting dropped) so I renamed it already where I could to reduce the size of the patch once GitOps get it renamed.

scoheb (commented):
lgtm

controllers/release/cache.go
scoheb: got it. tx

PR #98, opened 2022-09-14
fix(HACBS-1126): test suite for controllers/release is flaky

theflockers:
Fixes flaky release_adapter tests

- moves releasePlan/releaseAdmissionPlan creation to BeforeAll Block
- moves the test of unexistent releasePlan to the bottom

### DIFF
--- a/controllers/release/release_adapter_test.go
+++ b/controllers/release/release_adapter_test.go
@@ -80,9 +80,7 @@ var _ = Describe("Release Adapter", Ordered, func() {
 			},
 		}
 		Expect(k8sClient.Create(ctx, applicationSnapshot)).Should(Succeed())
-	})
 
-	BeforeEach(func() {
 		releasePlanAdmission = &appstudiov1alpha1.ReleasePlanAdmission{
 			ObjectMeta: metav1.ObjectMeta{
 				GenerateName: "test-releaseplanadmission-",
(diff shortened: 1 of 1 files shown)


controllers/release/release_adapter_test.go
```diff
@@ -491,13 +472,27 @@ var _ = Describe("Release Adapter", Ordered, func() {
 		link, err := adapter.getReleasePlan()
 		Expect(err).Should(Succeed())
 		Expect(reflect.TypeOf(link)).To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleasePlan{})))
+	})
 
-		// It should err when the ReleasePlan does not exist
+	It("ensures the target context returns an error when no releaseplan exists", func() {
```
davidmogar, line 477: Can we use the change to  rewrite the description? Something like this:
```suggestion
	It("ensure an error is returned when setting the target context with a missing releaseplan", func() {
```

theflockers: sure.

davidmogar (approved):

johnbieren:
Shouldn't we have just done the `Eventually` trick in the creation/deletion in the `BeforeEach` and `AfterEach` blocks instead? With things this way, we rely on each test not modifying the `ReleasePlan` and `ReleasePlanAdmission` or if they do, properly resetting it

johnbieren:
@theflockers @davidmogar ping ^

davidmogar:
I think we have some problems all over the tests that are actually difficult to see and that they appear randomly. I myself found one yesterday that I've been unable to replicate again to fix it. I also send you an IM about a problem I saw in one of the context tests. I'm kinda ignoring those problems because adding the `loader.go` file will cause multiple changes in the `release_adapter_test.go` file and that's a good opportunity to revisit this.

johnbieren:
Sure, but if we create things in BeforeEach instead of BeforeAll, there is less chance that one test breaks another test. AIUI, once Leandro fixes the bug of the release can't be deleted if the release plan gets deleted first, I don't see any reason to not have things in BeforeEach

PR #99, opened 2022-09-14
feat(HACBS-1083): add KCP Permission Claims

johnbieren:
This commit adds permission claims to the KCP release APIExport for the CRDs we depend on, but do not maintain. These are pipelineruns, applicationsnapshots, environments, applications, components, and
applicationsnapshotenvironmentbindings.

### DIFF
--- a/config/kcp/apiexport_release.yaml
+++ b/config/kcp/apiexport_release.yaml
@@ -3,6 +3,27 @@ kind: APIExport
 metadata:
   name: release
 spec:
+  # The identityHash values should be populated with proper values at
+  # deployment time.
+  permissionClaims:
+  - resource: "applications"
+    group: "appstudio.redhat.com"
+    identityHash: "APPLICATION_API_IDENTITY_HASH_PLACEHOLDER"
+  - resource: "components"
+    group: "appstudio.redhat.com"
+    identityHash: "APPLICATION_API_IDENTITY_HASH_PLACEHOLDER"
+  - resource: "applicationsnapshots"
+    group: "appstudio.redhat.com"
+    identityHash: "APPLICATION_API_IDENTITY_HASH_PLACEHOLDER"
+  - resource: "environments"
+    group: "appstudio.redhat.com"
+    identityHash: "APPLICATION_API_IDENTITY_HASH_PLACEHOLDER"
+  - resource: "applicationsnapshotenvironmentbindings"
+    group: "appstudio.redhat.com"
+    identityHash: "APPLICATION_API_IDENTITY_HASH_PLACEHOLDER"
+  - resource: "pipelineruns"
+    group: "tekton.dev"
+    identityHash: null
   latestResourceSchemas:
     - md5-06f1498b5778196e58c2bc1fc88108b8.releaseplanadmissions.appstudio.redhat.com
     - md5-9959e22acaaf6d162a03980c75b93dfe.releaseplans.appstudio.redhat.com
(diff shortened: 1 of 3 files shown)


config/kcp/apiexport_release.yaml
```diff
@@ -6,6 +6,25 @@ kind: APIExport
 metadata:
   name: release
 spec:
+  permissionClaims:
```
johnbieren, line 9: @davidmogar am I missing any CRDs we rely on? I don't have dummy CRs to do an end to end flow since the binding changes so I am not sure. Maybe we should get together and run through the flow e2e in kcp together

johnbieren:
@dirgim fyi

config/kcp/apiexport_release.yaml
davidmogar: Not that I'm aware of. We will need GitOpsDeployments, but not yet. Regarding the getting together, it sounds good.

hack/generate-kcp-api.sh
```diff
@@ -52,6 +52,25 @@ kind: APIExport
 metadata:
   name: release
 spec:
+  permissionClaims:
+  - resource: "applications"
+    group: "appstudio.redhat.com"
+    #identityHash: "Get value from apiexport. It is not static" # has apiexport
```
davidmogar, line 58: Are this comments needed? I don't really get them btw. Not sure what they mean, but that's probably caused by my lack of knowledge on KCP.

In any case, I wouldn't add them like this. I would have an external file defined in... probably `config/kcp` with a list of the `permissionClaims` we need and then in the script process that file. It's cleaner and that way people could copy the script if they want.

johnbieren: Well, if you don't define the `identityHash`, it won't work. However, the values are essentially uuids. Nothing static I can put here will work. I opened a slack thread about this yesterday, I'll send it to you

johnbieren:
Another proposal for the identity hashes from the google doc is setting it to values like `GITOPS_IDENTITY_HASH_PLACEHOLDER` (so we wouldn't need the comment). I am fine with either. That example is just a proposal for now. If it is accepted and everyone adopts it, maybe we can too, but fine with how we have it for now

config/samples/kcp/managed_workspace/apibinding_permissionclaims.yaml
```diff
@@ -0,0 +1,22 @@
+# The intent of this file is to show what the APIBinding looks like if you run
+# a setup using permission claims instead of all the APIs provided by one
+# APIExport.
```
davidmogar, line 3: You can move it up.

davidmogar:
> Another proposal for the identity hashes from the google doc is setting it to values like `GITOPS_IDENTITY_HASH_PLACEHOLDER` (so we wouldn't need the comment). I am fine with either. That example is just a proposal for now. If it is accepted and everyone adopts it, maybe we can too, but fine with how we have it for now

Both are good.

config/samples/kcp/managed_workspace/apibinding_permissionclaims.yaml
johnbieren: Sorry, move what up?

davidmogar: `APIExport` to the line above. Not that long and feels weird.

davidmogar (approved):

PR #100, opened 2022-09-15
fix(HACBS-1145): releases not being properly terminated

theflockers:
The function getReleasePipelineRun should return nil,nil when the Release PipelineRun is not found instead of nil, err which causes the operation to be Requeued.

### DIFF
--- a/controllers/release/release_adapter.go
+++ b/controllers/release/release_adapter.go
@@ -24,6 +24,7 @@ import (
 
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/release-service/gitops"
+	"github.com/redhat-appstudio/release-service/kcp"
 	"github.com/redhat-appstudio/release-service/syncer"
 	ctrl "sigs.k8s.io/controller-runtime"
 
@@ -229,24 +230,40 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
 // EnsureTargetContextIsSet is an operation that will ensure that the targetContext of the adapter is not nil. It will
 // set it to a context from the workspace in the ReleasePlan if one is defined or to its own context if not.
 func (a *Adapter) EnsureTargetContextIsSet() (results.OperationResult, error) {
-	if a.targetContext == nil {
+	if a.release.Status.Target == (kcp.NamespaceReference{}) {
+		patch := client.MergeFrom(a.release.DeepCopy())
 		releasePlan, err := a.getReleasePlan()
 		if err != nil {
-			patch := client.MergeFrom(a.release.DeepCopy())
 			a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, err.Error())
 			return results.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.release, patch))
 		}
 
-		if releasePlan.Spec.Target.Workspace == "" {
-			a.logger.Info("Running in a traditional kubernetes environment")
-			a.targetContext = a.context
-		} else {
+		if _, ok := logicalcluster.ClusterFromContext(a.context); ok {
 			a.logger.Info("Running in a KCP environment")
-			a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(releasePlan.Spec.Target.Workspace))
+
+			if releasePlan.Spec.Target.Workspace == "" {
+				a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, "ReleasePlan Workspace is not set while running on KCP")
+				return results.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.release, patch))
+			}
+		} else if releasePlan.Spec.Target.Workspace != "" {
+			a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, "ReleasePlan Workspace is set while not running on KCP")
+			return results.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.release, patch))
 		}
 
+		a.release.Status.Target = releasePlan.Spec.Target
+		err = a.client.Status().Patch(a.context, a.release, patch)
+		if err != nil {
+			return results.RequeueWithError(err)
+		}
+	}
+
+	if a.release.Status.Target.Workspace == "" {
+		a.targetContext = a.context
+	} else {
+		a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(a.release.Status.Target.Workspace))
 		a.syncer.SetContext(a.targetContext)
 	}
+
 	return results.ContinueProcessing()
 }
 
@@ -434,7 +451,6 @@ func (a *Adapter) getReleasePlan() (*v1alpha1.ReleasePlan, error) {
 		Namespace: a.release.Namespace,
 		Name:      a.release.Spec.ReleasePlan,
 	}, releasePlan)
-
 	if err != nil {
 		return nil, err
 	}
--- a/controllers/release/release_adapter_test.go
+++ b/controllers/release/release_adapter_test.go
@@ -17,19 +17,21 @@ limitations under the License.
 package release
 
 import (
-	"github.com/redhat-appstudio/release-service/gitops"
 	"reflect"
 	"strings"
 	"time"
 
 	"github.com/kcp-dev/logicalcluster/v2"
+	"github.com/redhat-appstudio/release-service/gitops"
 	"github.com/redhat-appstudio/release-service/kcp"
 
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+
 	"github.com/operator-framework/operator-lib/handler"
 
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -633,43 +635,6 @@ var _ = Describe("Release Adapter", Ordered, func() {
 		})
 	})
 
-	It("ensures the target context is set when a workspace is targeted by releaseplan", func() {
-		// Set the targetContext to nil to check if the EnsureTargetContext sets its value
-		adapter.targetContext = nil
-
-		// Add a target workspace to the ReleasePlan
-		patch := client.MergeFrom(releasePlan.DeepCopy())
-		originalWorkspace := releasePlan.Spec.Target.Workspace
-		releasePlan.Spec.Target.Workspace = "test-workspace"
-		Expect(k8sClient.Patch(ctx, releasePlan, patch)).Should(Succeed())
-
-		Eventually(func() bool {
-			err := k8sClient.Get(ctx, types.NamespacedName{
-				Name:      releasePlan.Name,
-				Namespace: releasePlan.Namespace,
-			}, releasePlan)
-			return err == nil && releasePlan.Spec.Target.Workspace == "test-workspace"
-		}, time.Second*10).Should(BeTrue())
-
-		result, err := adapter.EnsureTargetContextIsSet()
-		clusterName, _ := logicalcluster.ClusterFromContext(adapter.targetContext)
-		Expect(!result.RequeueRequest).To(BeTrue())
-		Expect(err).NotTo(HaveOccurred())
-		Expect(clusterName.String()).To(Equal("test-workspace"))
-
-		// Restore original values
-		adapter.targetContext = adapter.context
-		patch = client.MergeFrom(releasePlan.DeepCopy())
-		releasePlan.Spec.Target.Workspace = originalWorkspace
-		Expect(k8sClient.Patch(ctx, releasePlan, patch)).Should(Succeed())
-	})
-
-	It("ensures the target context is set to default context when releaseplan does not contain a target workspace", func() {
-		result, err := adapter.EnsureTargetContextIsSet()
-		Expect(!result.RequeueRequest && err == nil).To(BeTrue())
-		Expect(adapter.targetContext).To(Equal(adapter.context))
-	})
-
 	It("ensures finalizers are called when a release is deleted", func() {
 		// Call to EnsureFinalizersAreCalled should succeed if the Release was not set to be deleted
 		result, err := adapter.EnsureFinalizersAreCalled()
@@ -878,25 +843,102 @@ var _ = Describe("Release Adapter", Ordered, func() {
 		Expect(reflect.TypeOf(link)).To(Equal(reflect.TypeOf(&appstudiov1alpha1.ReleasePlan{})))
 	})
 
-	It("ensure an error is returned when setting the target context with a missing releaseplan", func() {
-		// Delete the ReleasePlan
-		Expect(k8sClient.Delete(ctx, releasePlan)).Should(Succeed())
+	Context("When calling EnsureTargetContextIsSet", func() {
+		It("ensures the target context is set when a workspace is targeted by a ReleasePlan", func() {
+			// Set the targetContext to nil to check if the operation sets its value
+			adapter.targetContext = nil
+			patch := client.MergeFrom(release.DeepCopy())
 
-		Eventually(func() bool {
-			err := k8sClient.Get(ctx, types.NamespacedName{
-				Name:      releasePlan.Name,
-				Namespace: releasePlan.Namespace,
-			}, releasePlan)
-			return err != nil && errors.IsNotFound(err)
-		}, time.Second*10).Should(BeTrue())
+			// Add a target workspace to the Release Status to bypass the path loading and validating the ReleasePlan
+			release.Status.Target = kcp.NamespaceReference{
+				Namespace: "default",
+				Workspace: "test-workspace",
+			}
+			Expect(k8sClient.Status().Patch(ctx, release, patch)).To(Succeed())
 
-		// Set the targetContext to nil to check if the EnsureTargetContext sets its value
-		adapter.targetContext = nil
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{
+					Name:      release.Name,
+					Namespace: release.Namespace,
+				}, release)
+				return err == nil && release.Status.Target.Workspace == "test-workspace"
+			}, time.Second*10).Should(BeTrue())
 
-		result, err := adapter.EnsureTargetContextIsSet()
-		Expect(!result.RequeueRequest && result.CancelRequest && err == nil).To(BeTrue())
+			result, err := adapter.EnsureTargetContextIsSet()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(!result.RequeueRequest).To(BeTrue())
 
-		// Restore original targetContext value
-		adapter.targetContext = adapter.context
+			clusterName, _ := logicalcluster.ClusterFromContext(adapter.targetContext)
+			Expect(clusterName.String()).To(Equal("test-workspace"))
+
+			// Restore original values
+			adapter.targetContext = adapter.context
+		})
+
+		It("should mark release as invalid when the target workspace is set in a non-KCP environment", func() {
+			// Add a target workspace to the ReleasePlan
+			patch := client.MergeFrom(releasePlan.DeepCopy())
+			originalWorkspace := releasePlan.Spec.Target.Workspace
+			releasePlan.Spec.Target.Workspace = "test-workspace"
+			Expect(k8sClient.Patch(ctx, releasePlan, patch)).Should(Succeed())
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{
+					Name:      releasePlan.Name,
+					Namespace: releasePlan.Namespace,
+				}, releasePlan)
+				return err == nil && releasePlan.Spec.Target.Workspace == "test-workspace"
+			}, time.Second*10).Should(BeTrue())
+
+			result, err := adapter.EnsureTargetContextIsSet()
+			Expect(result.CancelRequest).To(BeTrue())
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{
+					Name:      release.Name,
+					Namespace: release.Namespace,
+				}, release)
+				condition := meta.FindStatusCondition(release.Status.Conditions, "Succeeded")
+				return err == nil && condition != nil && condition.Reason == string(appstudiov1alpha1.ReleaseReasonReleasePlanValidationError)
+			}, time.Second*10).Should(BeTrue())
+
+			// Restore original values
+			adapter.targetContext = adapter.context
+			patch = client.MergeFrom(releasePlan.DeepCopy())
+			releasePlan.Spec.Target.Workspace = originalWorkspace
+			Expect(k8sClient.Patch(ctx, releasePlan, patch)).Should(Succeed())
+		})
+
+		It("ensures the target context is set to the default context when the ReleasePlan does not contain a target workspace", func() {
+			result, err := adapter.EnsureTargetContextIsSet()
+			Expect(!result.RequeueRequest && err == nil).To(BeTrue())
+			Expect(adapter.targetContext).To(Equal(adapter.context))
+		})
+
+		It("ensures an error is returned when the ReleasePlan is missing", func() {
+			// Delete the ReleasePlan
+			Expect(k8sClient.Delete(ctx, releasePlan)).Should(Succeed())
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{
+					Name:      releasePlan.Name,
+					Namespace: releasePlan.Namespace,
+				}, releasePlan)
+				return err != nil && errors.IsNotFound(err)
+			}, time.Second*10).Should(BeTrue())
+
+			// Set the targetContext to nil to check if the EnsureTargetContextIsSet sets its value
+			adapter.targetContext = nil
+
+			result, err := adapter.EnsureTargetContextIsSet()
+			Expect(!result.RequeueRequest && result.CancelRequest && err == nil).To(BeTrue())
+
+			// Restore original targetContext value
+			adapter.targetContext = adapter.context
+
+			releasePlan.ResourceVersion = ""
+			Expect(k8sClient.Create(ctx, releasePlan)).Should(Succeed())
+		})
 	})
 })
--- a/api/v1alpha1/release_types.go
+++ b/api/v1alpha1/release_types.go
@@ -105,7 +105,7 @@ type ReleaseStatus struct {
 
 	// Target references where this relesae is intended to be released to
 	// +optional
-	Target kcp.NamespaceReference `json:"target"`
+	Target kcp.NamespaceReference `json:"target,omitempty"`
 }
 
 // +kubebuilder:object:root=true


theflockers:
The release can not be deleted because the releasePlan is often deleted before the release and this should not happen because the release needs the releasePlan (in getReleasePlan()) to set the target context. If the ReleasePlan is missing it keeps Requeuing it with an error.
 
The previous assumption that the release was getting stuck due to an bug in `getReleasePipelineRun` was not confirmed as it already return `nil,nil` in the expected situation. So I converted this back to draft to be further reworked or closed.

controllers/release/release_adapter.go
```diff
@@ -222,24 +222,20 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-
-		if releasePlan.Spec.Target.Workspace == "" {
-			a.logger.Info("Running in a traditional kubernetes environment")
-			a.targetContext = a.context
-		} else {
-			a.logger.Info("Running in a KCP environment")
-			a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(releasePlan.Spec.Target.Workspace))
-		}
+	if a.targetContext != nil {
+		a.logger.Info("target context is set")
```
davidmogar, line 226: We have to revisit logging in the future but I'd say that we want to skip any logging for the context. It's so annoying when going through the logs as it's always there for every single reconcile.

controllers/release/release_adapter.go
```diff
@@ -228,24 +228,41 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
 // EnsureTargetContextIsSet is an operation that will ensure that the targetContext of the adapter is not nil. It will
 // set it to a context from the workspace in the ReleasePlan if one is defined or to its own context if not.
 func (a *Adapter) EnsureTargetContextIsSet() (results.OperationResult, error) {
-	if a.targetContext == nil {
+	if a.targetContext != nil {
```
davidmogar, line 231: @johnbieren Are you ok with removing this check? This is not a real condition as the adapter is created per reconcile.

davidmogar (changes requested):

controllers/release/release_adapter.go
```diff
@@ -228,24 +228,41 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
+			return results.RequeueOnErrorOrStop(err)
+		}
+	}
 
-		a.syncer.SetContext(a.targetContext)
+	if a.release.Status.Target.Workspace == "" {
+		a.logger.Info("Running in a traditional kubernetes environment")
+		a.targetContext = a.context
+	} else {
+		a.logger.Info("Running in a KCP environment")
```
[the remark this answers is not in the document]
theflockers: By the way @johnbieren @davidmogar  I think this is not so correct as it gave me the impression that what is selecting if it is a KCP or not is the existance of the "Workspace" set in the Spec of the releasePlan. I guess it should check the existence of the cluster ID.

For instance, yesterday I forced adding "workspace" to the spec without an actual KCP environment, it breaks with _"Index with name cluster does not exist"_. 

So I guess by looking If the "logicalcluster" exists should be the proper way to define if it is a KCP env or not.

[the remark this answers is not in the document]
davidmogar: Ok, good point. Checking the code we can have this at the top of the operation I think:
```
if _, ok := logicalcluster.ClusterFromContext(a.context); !ok {
    a.targetContext = a.context
    return results.ContinueProcessing()
}
// Actual code to set the targetContext for kcp using the Release status or the ReleasePlan
```
The cluster is set in the context in the `release_controller` for KCP, so it should work.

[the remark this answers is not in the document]
johnbieren: I find the log line telling me if the controller started thinking we are in kcp or not *extremely* useful in my testing. Don't we also have some low priority task to add more logging to our controller? If it comes from main or somewhere else, I don't really care, but I would like to remain to be able to see what the controller thinks when it starts.

To Leandro's point, I think you *should* break if you define a target workspace that is not available to you. If I set my CRs up to release to a KCP workspace `foo` and `foo` doesn't exist to the controller, I don't want the controller to do some magic behind my back to still get releases going to who knows where. I want it to break and tell me that `foo` does not exist.

About removing the if check, I just had it for efficiency. There was no reason to go fetch CRs and parse them if the target context was already set

controllers/release/release_adapter.go
```diff
@@ -228,24 +228,41 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-			a.logger.Info("Running in a traditional kubernetes environment")
-			a.targetContext = a.context
-		} else {
-			a.logger.Info("Running in a KCP environment")
-			a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(releasePlan.Spec.Target.Workspace))
+		patch := client.MergeFrom(a.release.DeepCopy())
+		if releasePlan.Spec.Target.Workspace != "" {
+			a.release.Status.Target.Workspace = releasePlan.Spec.Target.Workspace
+		}
+		if releasePlan.Spec.Target.Namespace != "" {
```
johnbieren, line 248: This is a required field. We already check for errors retrieving the releasePlan so not sure what more this check gets us

controllers/release/release_adapter.go
```diff
@@ -228,24 +229,31 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-	if a.targetContext == nil {
+	if _, isKcp := logicalcluster.ClusterFromContext(a.context); !isKcp {
+		a.targetContext = a.context
+		return results.ContinueProcessing()
+	}
+
+	// If the target namespace/workspace is not set we should get the target from the ReleasePlan
+	if a.release.Status.Target == (kcp.NamespaceReference{}) {
 		releasePlan, err := a.getReleasePlan()
 		if err != nil {
```
davidmogar, line 240: Wondering if we could check here for the workspace to exist as well and mark the release as invalid if not. This is KCP code already so... why not ;)

controllers/release/release_adapter.go
```diff
@@ -228,24 +229,31 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
 
-		if releasePlan.Spec.Target.Workspace == "" {
-			a.logger.Info("Running in a traditional kubernetes environment")
-			a.targetContext = a.context
-		} else {
-			a.logger.Info("Running in a KCP environment")
-			a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(releasePlan.Spec.Target.Workspace))
+		patch := client.MergeFrom(a.release.DeepCopy())
+		a.release.Status.Target.Workspace = releasePlan.Spec.Target.Workspace
+		a.release.Status.Target.Namespace = releasePlan.Spec.Target.Namespace
```
davidmogar, line 248: Can this be `a.release.Status.Target = releasePlan.Spec.Target`?

controllers/release/release_adapter.go
```diff
@@ -228,24 +229,31 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
+		err = a.client.Status().Patch(a.context, a.release, patch)
+		if err != nil {
+			return results.RequeueOnErrorOrStop(err)
 		}
-
-		a.syncer.SetContext(a.targetContext)
+		a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(a.release.Status.Target.Workspace))
```
davidmogar, line 253: This like should be moved outside the block. If not, the `targetContext` will get set only for non-KCP and whenever the Release status doesn't have the target set.

davidmogar (changes requested):

controllers/release/release_adapter.go
```diff
@@ -228,24 +228,41 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
 // EnsureTargetContextIsSet is an operation that will ensure that the targetContext of the adapter is not nil. It will
 // set it to a context from the workspace in the ReleasePlan if one is defined or to its own context if not.
 func (a *Adapter) EnsureTargetContextIsSet() (results.OperationResult, error) {
-	if a.targetContext == nil {
+	if a.targetContext != nil {
+		return results.ContinueProcessing()
+	}
+
+	// If the target namespace/workspace is not set we should get the target from the ReleasePlan
+	if a.release.Status.Target.Namespace == "" && a.release.Status.Target.Workspace == "" {
```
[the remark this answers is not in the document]
theflockers: Good one.

controllers/release/release_adapter.go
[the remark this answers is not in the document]
theflockers: @johnbieren @davidmogar I change the code to check if the Logical cluster exists but It comes with a problem. I can not test anything regarding KCP. It is possible to make it reach the workspace part for the code but then the Release Plan can not be found inside it due to the new context. Maybe there is a way to make it work, but could not do it so far.

I might change this back to the old way until we have a way to test the release service in KCP. The code is pushed anyway. I will move this to draft again to not bother with my commits and pushes.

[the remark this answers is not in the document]
theflockers: By the "old way" I meant the current way (checking if workspace parameter exists in the given spec) instead of my proposal of checking the existence of a logicalcluster and, thus, consider that it is a KCP env. My only point is if this is correct or not.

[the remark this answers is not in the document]
davidmogar: As agreed on the meeting today:
1. The entry point to the function should be check if the Release status target is set to avoid loading the ReleasePlan.
2.  We are going to use logicalcluster so we can check if we are actually running on a KCP environment o not (nice suggestion @theflockers).
3. Based on the previous information we will validate if the workspace should be `""` (non-KCP) or have an actual value (KCP) and fail if necessary (abort other operations).
4. We'll be logging only whenever the Release status target gets updated.

controllers/release/release_adapter.go
```diff
@@ -229,25 +229,29 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
 // EnsureTargetContextIsSet is an operation that will ensure that the targetContext of the adapter is not nil. It will
 // set it to a context from the workspace in the ReleasePlan if one is defined or to its own context if not.
 func (a *Adapter) EnsureTargetContextIsSet() (results.OperationResult, error) {
```
johnbieren, line 231: Where is the release status being set in a non KCP env?

controllers/release/release_adapter.go
```diff
@@ -229,25 +229,29 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
+		if _, ok := logicalcluster.ClusterFromContext(a.context); !ok {
+			a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, fmt.Sprintf("Workspace %s does not exist", releasePlan.Spec.Target.Workspace))
+			return results.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.release, patch))
+		}
+		a.logger.Info("Running in a KCP environment")
 		a.release.Status.Target = releasePlan.Spec.Target
+		return results.RequeueOnErrorOrContinue(a.client.Status().Patch(a.context, a.release, patch))
 	}
 
 	if a.release.Status.Target.Workspace != "" {
```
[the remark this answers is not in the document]
theflockers: This is because it only enters that section if the `release.Status.Target` was not modified. So In case it was already set (from the releasePlan) inside that block it won't need to fetch the releasePlan once again, but from the release status.

controllers/release/release_adapter_test.go
```diff
@@ -635,6 +636,7 @@ var _ = Describe("Release Adapter", Ordered, func() {
 	})
 
 	It("ensures the target context is set when a workspace is targeted by releaseplan", func() {
+		Skip("this test can only run in KCP environments")
```
theflockers, line 639: This test only would work on KCP environments once this change is there. Should I actually remove this test or only skip  for when we are KCP ready?

controllers/release/release_adapter.go
theflockers: Is it required? I thought the Status would be important only to set the context workspace. I understood that the context is the same in non-KCP environments.

controllers/release/release_adapter_test.go
davidmogar: So far we are focusing tests on pre-kcp. If it cannot be tested in that environment I'd say we should remove it.

controllers/release/release_adapter.go
```diff
@@ -228,24 +229,36 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-			patch := client.MergeFrom(a.release.DeepCopy())
 			a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, err.Error())
 			return results.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.release, patch))
 		}
-
-		if releasePlan.Spec.Target.Workspace == "" {
-			a.logger.Info("Running in a traditional kubernetes environment")
-			a.targetContext = a.context
-		} else {
+		if releasePlan.Spec.Target.Workspace != "" {
```
davidmogar, line 240: You already know this is empty as `Target` is nil.

controllers/release/release_adapter.go
davidmogar: Status should be updated for both as we want to be consistent.

controllers/release/release_adapter.go
```diff
@@ -228,24 +229,36 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
+			if _, ok := logicalcluster.ClusterFromContext(a.context); !ok {
```
davidmogar, line 241: This check shouldn't trigger an error as we could be running in pre-kcp, which is fine.

davidmogar (changes requested):

controllers/release/release_adapter.go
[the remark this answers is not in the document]
johnbieren: I don't follow. This is how I read the code
```
if the release.status.target is empty, which is only true the first time we reconcile a release
     get releaseplan
     if the releaseplan workspace is not empty (we are expecting kcp), see if the cluster exists
           if the cluster exists, output we are running in kcp to the logger. but don't set the status to that for some reason?
     set the release.status.target to the target of the releaseplan. so, the first if will never be true again for this release
now let's again see if a workspace was defined
     if it is, set the target context to the kcp one
     else, set it equal to the default context
```
I do not see why we cannot set the target context when we set the status

controllers/release/release_adapter.go
theflockers: My entry point checks for `a.release.Status.Target == (kcp.NamespaceReference{}) ` which could be false as well if `Workspace` is not set, but namespace is. I will change to check only for the Workspace to avoid redoing this check.

controllers/release/release_adapter.go
```diff
@@ -229,25 +229,29 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-	patch := client.MergeFrom(a.release.DeepCopy())
 	// If the target namespace/workspace is not set we should get the target from the ReleasePlan
 	if a.release.Status.Target == (kcp.NamespaceReference{}) {
+		patch := client.MergeFrom(a.release.DeepCopy())
 		releasePlan, err := a.getReleasePlan()
 		if err != nil {
 			a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, err.Error())
 			return results.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.release, patch))
 		}
+		if _, ok := logicalcluster.ClusterFromContext(a.context); !ok {
```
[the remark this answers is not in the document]
theflockers: No because the releasePlan also has the `Target.Namespace` changing the object anyway.

controllers/release/release_adapter_test.go
```diff
@@ -259,11 +259,7 @@ var _ = Describe("Release Adapter", Ordered, func() {
 		Expect(adapter.targetContext).To(Equal(adapter.context))
 	})
 
-	It("ensures the target context is set when a workspace is targeted by releaseplan", func() {
```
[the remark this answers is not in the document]
theflockers: @johnbieren 
I understood that it should fail if it is a non-KCP env but the Workspace is set as in @davidmogar 's comment:

>3. Based on the previous information we will validate if the workspace should be "" (non-KCP) or have an actual value (KCP) and fail if necessary (abort other operations).

By doing this the test can't work because we are not able to set the context for an inexistent Workspace.

controllers/release/release_adapter.go
theflockers: I have treat this error somewhere. The problem is if the`Workspace` is set and it is not a KCP cluster, it will trigger an untreated error in the controller. Although after the last change, I noticed it makes more sense to check this after I know that the Workspace at the `release.Status.Target` is already set. I just pushed it. Hope the next version is ok.

controllers/release/release_adapter_test.go
johnbieren: Why can't it be tested? We already created one context, why can't we create a second one? If a second context exists, wouldn't the logical cluster check pass?

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
johnbieren: Yes, I agree it should fail in this scenario. Shouldn't this change introduce a test for that scenario?

controllers/release/release_adapter.go
theflockers: Fixed.

controllers/release/release_adapter.go
johnbieren: yes, I agree this should be an error since it is after the check that workspace is defined

controllers/release/release_adapter.go
theflockers: Indeed. I refactored this a bit as I was checking this 3 times. I will resolve this comment so you can re-check this.

controllers/release/release_adapter.go
```diff
@@ -228,24 +228,41 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-			a.logger.Info("Running in a KCP environment")
-			a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(releasePlan.Spec.Target.Workspace))
+		patch := client.MergeFrom(a.release.DeepCopy())
+		if releasePlan.Spec.Target.Workspace != "" {
+			a.release.Status.Target.Workspace = releasePlan.Spec.Target.Workspace
+		}
+		if releasePlan.Spec.Target.Namespace != "" {
+			a.release.Status.Target.Namespace = releasePlan.Spec.Target.Namespace
 		}
+		err = a.client.Status().Patch(a.context, a.release, patch)
```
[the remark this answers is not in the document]
theflockers: I refactored here that I will close this comment. Please review again.

controllers/release/release_adapter.go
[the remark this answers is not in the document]
theflockers: I refactored this once again. Please check if now it makes more sense.

controllers/release/release_adapter_test.go
theflockers: The problem is this actually here:
```
		a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(a.release.Status.Target.Workspace))
		a.syncer.SetContext(a.targetContext)
```
Before this proposed change, the test could work because the adapter does not care if the actual context exists. Although while I was testing it in the cluster, if the `Spec.Workspace` is set in the releasePlan for a non-KCP cluster, it throws an error in the `release_controller` as the context does not exist and because this error is not being caught anywhere it is not possible to know that the cause is the `Workspace` wrongly set.

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
theflockers: Yes. I have done it but noticed now that after the rebase it went to hell. Probably I removed it wrongly to resolve the conflicts.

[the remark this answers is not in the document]
theflockers: test added.

controllers/release/release_adapter.go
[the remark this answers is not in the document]
johnbieren: Nope, let's meet about it

controllers/release/release_adapter_test.go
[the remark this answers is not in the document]
johnbieren: I see the new test being added now, but I still see a test being removed. We already had a unit test that ensures the target context is properly modified when a workspace is defined. It isn't real KCP, but with the functions available, we could get it close enough for a working test. I do not know why we are removing this test. We still need to ensure that when a workspace is defined, the target context is modified appropriately. This is a very important part of the code to keep working

controllers/release/release_adapter.go
davidmogar: Got you. I got confused about that.

Can we though reverse the if logic. Always better to start with the positive case (just a good practice)

controllers/release/release_adapter.go
```diff
@@ -228,24 +229,39 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
+			return results.RequeueWithError(err)
 		}
 
+	}
+
+	if a.release.Status.Target.Workspace != "" {
+		// If this is a non-KCP cluster and the target workspace is set we should stop the release
+		// otherwise it fails at the release_controller because it can not switch to a non existent
+		// context
+		if _, ok := logicalcluster.ClusterFromContext(a.context); !ok {
```
davidmogar, line 253: This block should be moved to the first condition where we set the target in the Release status. We don't want to check this (nor log that we are running in a KCP environment) in every reconciliation.

controllers/release/release_adapter.go
```diff
@@ -228,24 +229,39 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
+			patch := client.MergeFrom(a.release.DeepCopy())
+			a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, fmt.Sprintf("KCP Workspace %s does not exist", a.release.Status.Target.Workspace))
```
davidmogar, line 255: The error should probably be something like `ReleasePlan Workspace set on a non-KCP environment`.

theflockers: But it also fails if the workspace does not exist - I.E. Is a KCP cluster but the name has a typo. What about:
_ReleasePlan Workspace inexistent or set on a non-KCP environment_

controllers/release/release_adapter.go
theflockers: done.

controllers/release/release_adapter.go
```diff
@@ -228,23 +229,41 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
+		}
+
+		if a.release.Status.Target.Workspace == "" {
 			a.targetContext = a.context
 		} else {
+			a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(a.release.Status.Target.Workspace))
+			// If this is a non-KCP cluster and the target workspace is set we should stop the release
+			// otherwise it fails at the release_controller because it can not switch to a non existent
+			// context
+			if _, ok := logicalcluster.ClusterFromContext(a.context); !ok {
```
johnbieren, line 257: hmmm. I think we need to separate this out. If this check fails on `a.context`, set in a non-KCP environment is true.

However, `ReleasePlan Workspace does not exist` is only true if this check fails for `a.targetContext`, not `a.context`. Maybe we have two separate checks? @davidmogar wdyt

theflockers:
new adjusts to code and tests. Please review @davidmogar - @johnbieren  there are some changes that you might need talk to David to clarify why this was changed.

controllers/release/release_adapter.go
```diff
@@ -227,24 +228,40 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-		if releasePlan.Spec.Target.Workspace == "" {
-			a.logger.Info("Running in a traditional kubernetes environment")
-			a.targetContext = a.context
-		} else {
+		if _, ok := logicalcluster.ClusterFromContext(a.context); ok {
 			a.logger.Info("Running in a KCP environment")
-			a.targetContext = logicalcluster.WithCluster(a.context, logicalcluster.New(releasePlan.Spec.Target.Workspace))
+
+			if releasePlan.Spec.Target.Workspace == "" {
+				a.release.MarkInvalid(v1alpha1.ReleaseReasonReleasePlanValidationError, "ReleasePlan Workspace is not set while running on KCP")
```
davidmogar, line 243: Maybe it should say `ReleasePlan Workspace not set...`. Leaving this (and the one below) to @johnbieren.

davidmogar (commented):
I'll take a look to the code once the operation's code is final.

controllers/release/release_adapter_test.go
```diff
@@ -878,25 +842,121 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			Expect(!result.RequeueRequest).To(BeTrue())
+
+			clusterName, _ := logicalcluster.ClusterFromContext(adapter.targetContext)
+			Expect(clusterName.String()).To(Equal("test-workspace"))
+
+			// Restore original values
+			adapter.targetContext = adapter.context
+		})
+
+		It("should mark release as invalid when the target workspace is set on a non-KCP environment", func() {
```
johnbieren, line 898: `s/on/in/`

controllers/release/release_adapter_test.go
```diff
@@ -878,25 +842,121 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			Expect(release.Status.Conditions[0].Reason).To(Equal("ReleasePlanValidationError"))
+
+			// Restore original values
+			adapter.targetContext = adapter.context
+			patch = client.MergeFrom(releasePlan.DeepCopy())
+			releasePlan.Spec.Target.Workspace = originalWorkspace
+			Expect(k8sClient.Patch(ctx, releasePlan, patch)).Should(Succeed())
+		})
+
+		It("ensures the target context is set to default context when releaseplan does not contain a target workspace", func() {
```
johnbieren, line 934: Isn't this the same as line 846?

controllers/release/release_adapter_test.go
```diff
@@ -878,25 +842,121 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			result, err := adapter.EnsureTargetContextIsSet()
+			Expect(!result.RequeueRequest && err == nil).To(BeTrue())
+			Expect(adapter.targetContext).To(Equal(adapter.context))
+		})
+
+		It("ensure an error is returned when setting the target context with a missing releaseplan", func() {
```
johnbieren, line 940: `s/ensure/ensures/`

controllers/release/release_adapter_test.go
```diff
@@ -878,25 +842,121 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			// Delete the ReleasePlan
+			Expect(k8sClient.Delete(ctx, releasePlan)).Should(Succeed())
```
johnbieren, line 942: This is why I wanted to have the releasePlan (and admission) created in BeforeEach instead of BeforeAll. This test *must* remain as our last test in the file now

johnbieren: Can we at least have `Expect(k8sClient.Create(ctx, releasePlan)).Should(Succeed())` after line 958 for now then?

controllers/release/release_adapter_test.go
theflockers: Partially. The difference from the line 846 is that in this case `release.Status.Target` has the Namespace not empty so it falls in the `if  a.release.Status.Target.Workspace == ""` block.

johnbieren: I don't think I follow. The checks in both tests are
```
			result, err := adapter.EnsureTargetContextIsSet()
			Expect(!result.RequeueRequest && err == nil).To(BeTrue())
			Expect(adapter.targetContext).To(Equal(adapter.context))
```

It seems the test on 846 just also checks the status fields first. I am not seeing any difference in the target context checks (which is all this second test checks). If the first test checks the first if statement and the second if statement and the second test checks just the second if statement, but the boolean outcome of the if statements are the same in both, I don't see a need for the second test

controllers/release/release_adapter_test.go
theflockers: Sure. Added.

controllers/release/release_adapter_test.go
theflockers: I guess you are right. I will look at both tests better tomorrow and if necessary join them in one, or only get rid of one of them.

theflockers: I removed the test at line 846. You are right and it is a duplication.

controllers/release/release_adapter_test.go
```diff
@@ -878,25 +842,103 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{
+					Name:      releasePlan.Name,
+					Namespace: releasePlan.Namespace,
+				}, releasePlan)
+				return err != nil && errors.IsNotFound(err)
+			}, time.Second*10).Should(BeTrue())
+
+			// Set the targetContext to nil to check if the EnsureTargetContext sets its value
```
davidmogar, line 931:
```suggestion
			// Set the targetContext to nil to check if the EnsureTargetContextIsSet sets its value
```

johnbieren (approved):

davidmogar (approved):
Good work :+1:

PR #101, opened 2022-09-20
feat(HACBS-1085): set owner reference to the binding

davidmogar:
The binding needs to have an owner reference so it's deleted when the Application is deleted. As we are the ones creating this bindings, there's no need to set the onwer reference when the binding is updated. If the owner reference is not present, it was removed manually and there must be a reason for that.

Tests will be added as part of HACBS-1085.


theflockers (approved):
LGTM

PR #105, opened 2022-09-28
feat(HACBS-1097): add check to skip Binding creation

davidmogar:
Whenever the Environment is not set in the ReleasePlanAdmission the controller should not create a Binding. This allows to have strategies without the deployment part. This commit adds a check to skip processing the Binding creation for those scenarios.

### DIFF
--- a/controllers/release/release_adapter.go
+++ b/controllers/release/release_adapter.go
@@ -195,12 +195,17 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
 		return results.ContinueProcessing()
 	}
 
-	err := a.syncResources()
+	releasePlanAdmission, err := a.getActiveReleasePlanAdmission()
 	if err != nil {
 		return results.RequeueWithError(err)
 	}
 
-	releasePlanAdmission, err := a.getActiveReleasePlanAdmission()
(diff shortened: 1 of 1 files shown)


johnbieren (approved):
Are tests for this going to be part of your "add tests" pr?
And I assume your local testing of this ensures that `== ""` is true when the field is missing (I am pretty sure that is the case and we've talked about it before, but I always forget)

davidmogar:
> Are tests for this going to be part of your "add tests" pr? And I assume your local testing of this ensures that `== ""` is true when the field is missing (I am pretty sure that is the case and we've talked about it before, but I always forget)

Actually I don't have a preference. Should we merge the other PR, rebase this one and add the tests?

Regarding the `== ""` yes, we discussed it and you did the same in other parts of the code ;)

johnbieren:
> > Are tests for this going to be part of your "add tests" pr? And I assume your local testing of this ensures that `== ""` is true when the field is missing (I am pretty sure that is the case and we've talked about it before, but I always forget)
> 
> Actually I don't have a preference. Should we merge the other PR, rebase this one and add the tests?
> 
> Regarding the `== ""` yes, we discussed it and you did the same in other parts of the code ;)

If you are already adding tests in another PR for this general functionality, I am fine with just making this part of those tests in that PR and merge this as is. I am also a little biased because I want this check merged for my stuff

PR #106, opened 2022-09-29
fix: restore releasePlan workspace in test modifying it

johnbieren:

### DIFF
--- a/controllers/release/release_adapter_test.go
+++ b/controllers/release/release_adapter_test.go
@@ -640,6 +640,7 @@ var _ = Describe("Release Adapter", Ordered, func() {
 
 		// Add a target workspace to the ReleasePlan
 		patch := client.MergeFrom(releasePlan.DeepCopy())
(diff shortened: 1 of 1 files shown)


controllers/release/release_adapter_test.go
```diff
@@ -657,8 +657,11 @@ var _ = Describe("Release Adapter", Ordered, func() {
 		Expect(err).NotTo(HaveOccurred())
 		Expect(clusterName.String()).To(Equal("test-workspace"))
 
-		// Restore original targetContext value
+		// Restore original values
 		adapter.targetContext = adapter.context
+		patch = client.MergeFrom(releasePlan.DeepCopy())
+		releasePlan.Spec.Target.Workspace = ""
```
davidmogar, line 663: I would store the value just before setting it in the previous patch. That way you can use a variable here and be certain you are restoring the original one.

davidmogar (approved):

PR #107, opened 2022-10-03
chore(HACBS-1201): migrate to shared api

davidmogar:
HAS and GitOps resources are now shared through a new repository. This commit changes old references to the new one.


johnbieren (approved):
needs `go.mod` update but otherwise lgtm

PR #114, opened 2022-10-06
fix: nit pick incorrect function names

johnbieren:


davidmogar (approved):
Ouch, my bad. Good eye!

PR #118, opened 2022-10-14
feat(HACBS-1157): add json EC Policy to Release PipelineRun

theflockers:
- adds the JSON representation of an EC Policy to the Release
  PipelineRun by adding `.WithEnterpriseContractPolicy()` to
  pipeline_run.go and `.getEnterpriseContractPolicy()` to the
  release_adapter.go as well changing the `createReleasePipelineRun()`
  to accept the EnterpriseContractPolicy object.
- adds and adjusts tests
- adds permission claims for ec resources
- adds enterprisecontractpolicy RBAC for controller-manager


controllers/release/release_adapter_test.go
```diff
@@ -49,17 +51,38 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+		environment              *applicationapiv1alpha1.Environment
+		release                  *appstudiov1alpha1.Release
+		releaseStrategy          *appstudiov1alpha1.ReleaseStrategy
+		releasePlan              *appstudiov1alpha1.ReleasePlan
+		releasePlanAdmission     *appstudiov1alpha1.ReleasePlanAdmission
+		enterpriseContractPolicy *ecapiv1alpha1.EnterpriseContractPolicy
 	)
 
 	BeforeAll(func() {
+
```
davidmogar, line 66:
```suggestion
```

tekton/pipeline_run.go
```diff
@@ -120,6 +122,18 @@ func (r *ReleasePipelineRun) WithApplicationSnapshot(snapshot *applicationapiv1a
 	return r
 }
 
+//  WithEnterpriseContractPolicy adds a param containing the EnterpriseContractPolicy Spec as a json string to the release PipelineRun.
+func (r *ReleasePipelineRun) WithEnterpriseContractPolicy(ecpolicy *ecapiv1alpha1.EnterpriseContractPolicy) *ReleasePipelineRun {
+	ecPolicy, _ := json.Marshal(ecpolicy.Spec)
+
+	r.WithExtraParam("policy", tektonv1beta1.ArrayOrString{
```
davidmogar, line 129: Do not hardcode the name. Do the same we do with the ApplicationSnapshot and change the Bundle.

davidmogar (changes requested):

controllers/release/release_adapter_test.go
theflockers: removed.

tekton/pipeline_run.go
theflockers: changed to use the ec policy kind.

theflockers:
Nice to know that. I will add the CRD claims to this PR and take a look in the RBAC as well. Thanks @davidmogar  and @johnbieren

config/kcp/apiexport_release.yaml
```diff
@@ -24,6 +24,9 @@ spec:
   - resource: "pipelineruns"
     group: "tekton.dev"
     identityHash: pipeline-service
+  - resource: "enterprisecontract"
+    group: "appstudio.redhat.com"
+    identityHash: enterprisecontract-service
```
johnbieren: no -service here

config/kcp/apiexport_release.yaml
```diff
@@ -24,6 +24,9 @@ spec:
   - resource: "pipelineruns"
     group: "tekton.dev"
     identityHash: pipeline-service
+  - resource: "enterprisecontract"
```
johnbieren: wrong name - https://github.com/hacbs-contract/enterprise-contract-controller/blob/main/config/crd/bases/appstudio.redhat.com_enterprisecontractpolicies.yaml#L14

hack/generate-kcp-api.sh
```diff
@@ -31,6 +31,9 @@ spec:
   - resource: "pipelineruns"
     group: "tekton.dev"
     identityHash: pipeline-service
+  - resource: "enterprisecontract"
```
johnbieren: same comments as in the apiexport

johnbieren (changes requested):

config/kcp/apiexport_release.yaml
johnbieren: you're quick!!

config/kcp/apiexport_release.yaml
theflockers: You are too fast :D. I was going through it now and found the name and the id hash i added were wrong. Please take a look again.

johnbieren (approved):
This lgtm. @davidmogar should there be a kubebuilder annotation for this? That is the only other thing I can think of that may be missing

controllers/release/release_adapter_test.go
```diff
@@ -717,13 +742,22 @@ var _ = Describe("Release Adapter", Ordered, func() {
 		// WithReleaseStrategy()
 		Expect(pipelineRun.Spec.PipelineRef.Name).To(Equal(releaseStrategy.Spec.Pipeline))
 		Expect(pipelineRun.Spec.PipelineRef.Bundle).To(Equal(releaseStrategy.Spec.Bundle))
+
 		Expect(pipelineRun.Spec.Params[0].Name).To(Equal("policy"))
```
davidmogar, line 746: Why is this here? We should not be passing the `policy` field any more, just the JSON representation on a `enterprisecontractpolicy` parameter.

In addition, we should not use `Params[n]`. If we change the order in the function itself, the tests would fail. Would be better to check if there's a Param containing the given values. Maybe it would make sense to change it here.

davidmogar:
> This lgtm. @davidmogar should there be a kubebuilder annotation for this? That is the only other thing I can think of that may be missing

He added the roles manually, but we should move away from this I think and specify everything using `kubebuilder` annotations as I asked you to do.

@theflockers, take a look to the release controller. There you will see rbac annotations. Define the ones you need for this change and then run `make manifest generate` to generate the changes in the role.

We should move all the other existing cluster roles to annotations eventually, but not as part of this PR.

controllers/release/release_adapter_test.go
theflockers: @davidmogar  `pipelineRun.Spec.Params[0]` is the `releaseStrategy` not the just added EC Policy which is the `pipelinerun.Spec.Params[1]`. 
I agree. I used the way it was before to not refactory the tests but I can do it NP.

davidmogar: We should not pass any `policy` as a parameter. Just the spec. That's the change you are working on here and it's actually missing. That should be removed in the test and in the code.

theflockers: I did not. You are looking to the wrong place. What you are looking for is a bit down the code. The code you pointed has nothing to do with the new feature, it regards the releaseStrategy. It was already there, but with different index. 
The test is at line `750`. This one:

```
		// WithEnterpriseContractPolicy()
		enterpriseContractPolicy, err := adapter.getEnterpriseContractPolicy(releaseStrategy)
		Expect(err).ShouldNot(HaveOccurred())

		enterpriseContractPolicyJsonSpec, _ := json.Marshal(enterpriseContractPolicy.Spec)
		Expect(pipelineRun.Spec.Params[1].Name).To(Equal("enterprisecontractpolicy"))
		Expect(pipelineRun.Spec.Params[1].Value.StringVal).To(Equal(string(enterpriseContractPolicyJsonSpec)))
``` 

That said, I am working on refactoring this full test.

theflockers: @davidmogar I refactored this full test. Please take a look.

tekton/pipeline_run.go
```diff
@@ -150,13 +165,6 @@ func (r *ReleasePipelineRun) WithReleaseStrategy(strategy *v1alpha1.ReleaseStrat
 
 	valueType := tektonv1beta1.ParamTypeString
 
-	if strategy.Spec.Policy != "" {
```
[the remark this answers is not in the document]
theflockers: update. Please take a look when you can!

davidmogar (approved):

PR #119, opened 2022-10-17
fix(HACBS-1257): add namePrefix to rbac resources

johnbieren:
The release-service has overlapping cluster roles with the integration service. This commit adds a namePrefix to the cluster roles in this repo so that they do not conflict with the integration service.

### DIFF
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -1,3 +1,5 @@
+namePrefix: release-service-
+
 resources:
 # All RBAC will be applied under this service account in
 # the deployment namespace. You may comment out this resource


johnbieren:
@dirgim @hongweiliu17 fyi

davidmogar (approved):
Leaving a comment here just to clarify, for future readers, that this impacts KCP exclusively. RBAC included on pre-KCP already had this prefix defined in the `default.yaml` file.

johnbieren:
> Leaving a comment here just to clarify, for future readers, that this impacts KCP exclusively. RBAC included on pre-KCP already had this prefix defined in the `default.yaml` file.

I added "for KCP" to the PR title so if a git blame trail is followed, it may be more obvious

PR #121, opened 2022-10-27
feat(hacbs-1193): copy pac metadata to pipelinerun

parasense:

### DIFF
--- a/controllers/release/release_adapter.go
+++ b/controllers/release/release_adapter.go
@@ -315,7 +315,7 @@ func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStra
 	snapshot *applicationapiv1alpha1.Snapshot) (*v1beta1.PipelineRun, error) {
 	pipelineRun := tekton.NewReleasePipelineRun("release-pipelinerun", releaseStrategy.Namespace).
 		WithOwner(a.release).
-		WithReleaseAndApplicationLabels(a.release.Name, a.release.Namespace, a.release.GetAnnotations()[logicalcluster.AnnotationKey], snapshot.Spec.Application).
+		WithReleaseAndApplicationMetadata(a.release, snapshot.Spec.Application).
 		WithReleaseStrategy(releaseStrategy).
 		WithEnterpriseContractPolicy(enterpriseContractPolicy).
 		WithSnapshot(snapshot).
--- a/metadata/metadata.go
+++ b/metadata/metadata.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strings"
+)
+
+// AddAnnotations copies the map into the resource's Annotations map.
+// When the destination map is nil, then the map will be created.
+// The unexported function addEntries is called with args passed.
+func AddAnnotations(obj v1.Object, entries map[string]string) {
+	if obj.GetAnnotations() == nil {
+		obj.SetAnnotations(map[string]string{})
+	}
+	addEntries(entries, obj.GetAnnotations())
+}
+
+// AddLabels copies the map into the resource's Labels map.
+// When the destination map is nil, then the map will be created.
+// The unexported function addEntries is called with args passed.
+func AddLabels(obj v1.Object, entries map[string]string) {
+	if obj.GetLabels() == nil {
+		obj.SetLabels(map[string]string{})
+	}
+	addEntries(entries, obj.GetLabels())
+}
+
+// GetAnnotationsWithPrefix is a method that returns a map of key/value pairs matching a prefix string.
+// The unexported function filterByPrefix is called with args passed.
+func GetAnnotationsWithPrefix(obj v1.Object, prefix string) map[string]string {
+	return filterByPrefix(obj.GetAnnotations(), prefix)
+}
+
+// GetLabelsWithPrefix is a method that returns a map of key/value pairs matching a prefix string.
+// The unexported function filterByPrefix is called with args passed.
+func GetLabelsWithPrefix(obj v1.Object, prefix string) map[string]string {
+	return filterByPrefix(obj.GetLabels(), prefix)
+}
+
+// addEntries copies key/value pairs in the source map adding them into the destination map.
+// The unexported function safeCopy is used to copy, and avoids clobbering existing keys in the destination map.
+func addEntries(source, destination map[string]string) {
+	for key, val := range source {
+		safeCopy(destination, key, val)
+	}
+}
+
+// filterByPrefix returns a map of key/value pairs contained in src that matches the prefix.
+// When the prefix is empty/nil, the source map is returned.
+// When source key does not contain the prefix string, no copy happens.
+func filterByPrefix(entries map[string]string, prefix string) map[string]string {
+	if len(prefix) == 0 {
+		return entries
+	}
+	dst := map[string]string{}
+	for key, val := range entries {
+		if strings.HasPrefix(key, prefix) {
+			dst[key] = val
+		}
+	}
+	return dst
+}
+
+// safeCopy conditionally copies a given key/value pair into a map.
+// When a key is already present in the map, no copy happens.
+func safeCopy(dst map[string]string, key, val string) {
+	if _, err := dst[key]; !err {
+		dst[key] = val
+	}
+}
--- a/metadata/metadata_test.go
+++ b/metadata/metadata_test.go
@@ -1,0 +1,211 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/redhat-appstudio/release-service/api/v1alpha1"
+	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
+)
+
+func TestMetadata(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Metadata Test Suite")
+}
+
+var _ = Describe("Metadata", func() {
+
+	Context("addEntries function", func() {
+		When("called with an empty destination map", func() {
+			src := map[string]string{
+				"pet/dog":   "bark",
+				"pet/cat":   "meow",
+				"pond/frog": "ribit",
+			}
+			dst := map[string]string{}
+			addEntries(src, dst)
+
+			It("should copy all src key/value pairs unconditionally", func() {
+				Expect(dst["pet/dog"]).To(Equal("bark"))
+				Expect(dst["pet/cat"]).To(Equal("meow"))
+				Expect(dst["pond/frog"]).To(Equal("ribit"))
+			})
+			It("source and destination maps should have the same length", func() {
+				Expect(len(dst)).To(Equal(len(src)))
+			})
+		})
+		When("called with a non-empty destination map containing duplicate keys", func() {
+			src := map[string]string{
+				"pet/dog":   "bark",
+				"pet/cat":   "meow",
+				"pond/frog": "ribit",
+			}
+			dst := map[string]string{
+				"pet/dog": "howl",
+			}
+			addEntries(src, dst)
+			It("should only copy source key/value pairs that do not already exist in the destination map", func() {
+				Expect(dst["pet/dog"]).To(Equal("howl"))
+				Expect(dst["pet/cat"]).To(Equal("meow"))
+				Expect(dst["pond/frog"]).To(Equal("ribit"))
+			})
+			It("should have increased the length of the destination map from 1 to 3", func() {
+				Expect(len(dst)).To(Equal(3))
+			})
+		})
+	})
+
+	Context("filterByPrefix function", func() {
+		When("called with an empty prefix string", func() {
+			src := map[string]string{
+				"pet/dog":   "bark",
+				"pet/cat":   "meow",
+				"pond/frog": "ribit",
+			}
+			dst := filterByPrefix(src, "")
+			It("should return a map identical to the map given", func() {
+				Expect(dst["pet/dog"]).To(Equal("bark"))
+				Expect(dst["pet/cat"]).To(Equal("meow"))
+				Expect(dst["pond/frog"]).To(Equal("ribit"))
+			})
+			It("input and returned maps should have the same length", func() {
+				Expect(len(dst)).To(Equal(len(src)))
+			})
+		})
+		When("called with an non-empty prefix string", func() {
+			src := map[string]string{
+				"pet/dog":   "bark",
+				"pet/cat":   "meow",
+				"pond/frog": "ribit",
+			}
+			dst := filterByPrefix(src, "pond")
+			It("should return a new map of key/value pairs where the key(s) matched the given prefix", func() {
+				Expect(dst["pond/frog"]).To(Equal("ribit"))
+			})
+			It("returned map length should equal exactly 1.", func() {
+				Expect(len(dst)).To(Equal(1))
+			})
+		})
+	})
+
+	Context("safeCopy function", func() {
+		When("called with a key not already present in the map", func() {
+			dst := map[string]string{"foo/dog": "bark"}
+			safeCopy(dst, "foo/cat", "meow")
+			It("should set the given key/value in the map", func() {
+				Expect(dst["foo/dog"]).To(Equal("bark"))
+				Expect(dst["foo/cat"]).To(Equal("meow"))
+			})
+			It("should have increased the map length from 1 to 2", func() {
+				Expect(len(dst)).To(Equal(2))
+			})
+		})
+		When("called with a key already present in dst map", func() {
+			dst := map[string]string{"foo/dog": "bark"}
+			safeCopy(dst, "foo/dog", "meow")
+			It("should not set the given key/value pair into the map", func() {
+				Expect(dst["foo/dog"]).To(Equal("bark"))
+			})
+			It("should not increase the map length", func() {
+				Expect(len(dst)).To(Equal(1))
+			})
+		})
+	})
+
+	Context("AddAnnotations function", func() {
+		When("called with a pipelineRun containing nil Annotations", func() {
+			pipelineRun := &v1beta1.PipelineRun{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: nil,
+					Labels:      nil,
+				},
+			}
+			AddAnnotations(pipelineRun, map[string]string{})
+			It("should create the Annotations map", func() {
+				Expect(pipelineRun.GetAnnotations()).ToNot(BeNil())
+			})
+			It("should do nothing with the Labels map", func() {
+				Expect(pipelineRun.GetLabels()).To(BeNil())
+			})
+		})
+	})
+
+	Context("AddLabels function", func() {
+		When("called with a pipelineRun containing nil Labels", func() {
+			pipelineRun := &v1beta1.PipelineRun{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: nil,
+					Labels:      nil,
+				},
+			}
+			AddLabels(pipelineRun, map[string]string{})
+			It("should create the Labels map", func() {
+				Expect(pipelineRun.GetLabels()).ToNot(BeNil())
+			})
+			It("should do nothing with the Annotations map", func() {
+				Expect(pipelineRun.GetAnnotations()).To(BeNil())
+			})
+		})
+	})
+
+	Context("GetAnnotationsWithPrefix function", func() {
+		When("calling filterByPrefix with GetAnnotations()", func() {
+			release := &v1alpha1.Release{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						"pet/dog": "bark",
+					},
+					Labels: map[string]string{
+						"foo": "bar",
+					},
+				},
+			}
+			dst := GetAnnotationsWithPrefix(release, "pet/")
+			It("should only fetch Annotations", func() {
+				Expect(dst["pet/dog"]).To(Equal("bark"))
+			})
+			It("should not fetch Labels", func() {
+				Expect(dst["foo"]).To(BeEmpty())
+			})
+		})
+	})
+
+	Context("GetLabelsWithPrefix function", func() {
+		When("calling filterByPrefix with GetLabels()", func() {
+			release := &v1alpha1.Release{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						"foo": "bar",
+					},
+					Labels: map[string]string{
+						"pet/dog": "bark",
+					},
+				},
+			}
+			dst := GetLabelsWithPrefix(release, "pet/")
+			It("should only fetch Labels", func() {
+				Expect(dst["pet/dog"]).To(Equal("bark"))
+			})
+			It("should not fetch Annotations", func() {
+				Expect(dst["foo"]).To(BeEmpty())
+			})
+		})
+	})
+})
--- a/tekton/pipeline_run.go
+++ b/tekton/pipeline_run.go
@@ -24,7 +24,9 @@ import (
 	"unicode"
 
 	ecapiv1alpha1 "github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
+	"github.com/kcp-dev/logicalcluster/v2"
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	"github.com/redhat-appstudio/release-service/metadata"
 
 	libhandler "github.com/operator-framework/operator-lib/handler"
 	"github.com/redhat-appstudio/release-service/api/v1alpha1"
@@ -48,6 +50,9 @@ const (
 
 	//PipelineTypeRelease is the type for PipelineRuns created to run a release Pipeline
 	PipelineTypeRelease = "release"
+
+	// pipelinesAsCodeMetadataPrefix is the prefix for pipelines-as-code metadata
+	pipelinesAsCodeMetadataPrefix = "pipelinesascode.tekton.dev"
 )
 
 var (
@@ -121,7 +126,7 @@ func (r *ReleasePipelineRun) WithSnapshot(snapshot *applicationapiv1alpha1.Snaps
 	return r
 }
 
-//  WithEnterpriseContractPolicy adds a param containing the EnterpriseContractPolicy Spec as a json string to the release PipelineRun.
+// WithEnterpriseContractPolicy adds a param containing the EnterpriseContractPolicy Spec as a json string to the release PipelineRun.
 func (r *ReleasePipelineRun) WithEnterpriseContractPolicy(enterpriseContractPolicy *ecapiv1alpha1.EnterpriseContractPolicy) *ReleasePipelineRun {
 	policyJson, _ := json.Marshal(enterpriseContractPolicy.Spec)
 
@@ -143,16 +148,18 @@ func (r *ReleasePipelineRun) WithOwner(release *v1alpha1.Release) *ReleasePipeli
 	return r
 }
 
-// WithReleaseAndApplicationLabels adds Release and Application labels to the release PipelineRun.
-func (r *ReleasePipelineRun) WithReleaseAndApplicationLabels(releaseName, releaseNamespace, releaseWorkspace string, applicationName string) *ReleasePipelineRun {
+// WithReleaseAndApplicationMetadata adds Release and Application metadata to the release PipelineRun.
+func (r *ReleasePipelineRun) WithReleaseAndApplicationMetadata(release *v1alpha1.Release, applicationName string) *ReleasePipelineRun {
 	r.ObjectMeta.Labels = map[string]string{
 		PipelinesTypeLabel:    PipelineTypeRelease,
-		ReleaseNameLabel:      releaseName,
-		ReleaseNamespaceLabel: releaseNamespace,
+		ReleaseNameLabel:      release.Name,
+		ReleaseNamespaceLabel: release.Namespace,
 		// PipelineRun does not allow labels with : in the value, which KCP workspaces have
-		ReleaseWorkspaceLabel: strings.ReplaceAll(releaseWorkspace, ":", "__"),
+		ReleaseWorkspaceLabel: strings.ReplaceAll(release.GetAnnotations()[logicalcluster.AnnotationKey], ":", "__"),
 		ApplicationNameLabel:  applicationName,
 	}
+	metadata.AddAnnotations(r.AsPipelineRun(), metadata.GetAnnotationsWithPrefix(release, pipelinesAsCodeMetadataPrefix))
+	metadata.AddLabels(r.AsPipelineRun(), metadata.GetLabelsWithPrefix(release, pipelinesAsCodeMetadataPrefix))
 
 	return r
 }
--- a/tekton/pipeline_run_test.go
+++ b/tekton/pipeline_run_test.go
@@ -19,12 +19,13 @@ package tekton
 import (
 	"context"
 	"encoding/json"
+	"os"
 	"reflect"
 
 	ecapiv1alpha1 "github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 
-	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 
 	"github.com/kcp-dev/logicalcluster/v2"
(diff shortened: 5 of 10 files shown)


tekton/pipeline_run_test.go
```diff
@@ -214,4 +214,38 @@ var _ = Describe("PipelineRun", func() {
 				To(Equal(persistentVolumeClaim))
 		})
 	})
+
+	Context("When PipelineAsCode metadata is present in the Release", func() {
+		//
+		It("copies Annotations & Labels matching the PAC prefix", func() {
+			release.ObjectMeta.Annotations["foo"] = "bar"
+			release.ObjectMeta.Annotations[pipelinesAsCodeMetaPrefix+"/meow"] = "cat"
```
johnbieren, line 222: maybe one more annotation or label test like `foo/ + pipelinesAsCodeMetaPrefix`?

johnbieren (commented):
lgtm

tekton/pipeline_run_test.go
parasense: Oh good idea! 
The code checks for a beginning PAC prefix, so `"foo/" + pipelinesAsCodeMetaPrefix` should not ever copy. I've updated the test with that test case, and the code still works. Thanks! :+1:

johnbieren (dismissed):

tekton/pipeline_run.go
```diff
@@ -135,6 +138,34 @@ func (r *ReleasePipelineRun) WithEnterpriseContractPolicy(enterpriseContractPoli
...
+// When Annotations or Labels maps do not exist, then they are created.
+// Only things who match pipelinesAsCodeMetaPrefix are allowed to be copied.
+func (pipelineRun *ReleasePipelineRun) WithPipelinesAsCodeMeta(release *v1alpha1.Release) *ReleasePipelineRun {
+	if pipelineRun.ObjectMeta.Annotations == nil {
+		pipelineRun.ObjectMeta.Annotations = make(map[string]string)
+	}
+	copyPipelinesAsCodeMeta(release.ObjectMeta.Annotations, pipelineRun.ObjectMeta.Annotations)
+
+	if pipelineRun.ObjectMeta.Labels == nil {
+		pipelineRun.ObjectMeta.Labels = make(map[string]string)
```
davidmogar, line 162: This makes the With functions order dependant. If WithReleaseAndApplicationLabels get called  after this function, labels will get removed.

tekton/pipeline_run.go
```diff
@@ -135,6 +138,34 @@ func (r *ReleasePipelineRun) WithEnterpriseContractPolicy(enterpriseContractPoli
...
+		if _, err := dst[key]; !err && strings.HasPrefix(key, pipelinesAsCodeMetaPrefix) {
+			dst[key] = val
+		}
+	}
+}
+
+// WithPipelinesAsCodeMeta conditionally copies Labels and Annotations from Release to PipelineRun
+// When Annotations or Labels maps do not exist, then they are created.
+// Only things who match pipelinesAsCodeMetaPrefix are allowed to be copied.
+func (pipelineRun *ReleasePipelineRun) WithPipelinesAsCodeMeta(release *v1alpha1.Release) *ReleasePipelineRun {
```
davidmogar, line 155: Please, be consistent with receiver names.

```suggestion
func (r *ReleasePipelineRun) WithPipelinesAsCodeMeta(release *v1alpha1.Release) *ReleasePipelineRun {
```

tekton/pipeline_run.go
```diff
@@ -135,6 +138,34 @@ func (r *ReleasePipelineRun) WithEnterpriseContractPolicy(enterpriseContractPoli
 	return r
 }
 
+// copyPipelinesAsCodeMeta copies key/value pairs in src adding them to dst.
+// When a key in src is already present in dst, no copy happens.
+// When the src key does not contain the prefix string, no copy happens.
+func copyPipelinesAsCodeMeta(src, dst map[string]string) {
```
davidmogar, line 144: We can keep this within the function as it's not used by any other function here. If that happens in the future, we will extract it to a more generic package.

tekton/pipeline_run.go
```diff
@@ -48,6 +48,9 @@ const (
 
 	//PipelineTypeRelease is the type for PipelineRuns created to run a release Pipeline
 	PipelineTypeRelease = "release"
+
+	// pipelinesAsCodeMetaPrefix is the prefix for pipelines-as-code metadata
+	pipelinesAsCodeMetaPrefix = "pipelinesascode.tekton.dev"
```
davidmogar, line 53:
```suggestion
	pipelinesAsCodeLabelsPrefix = "pipelinesascode.tekton.dev"
```

davidmogar (changes requested):

tekton/pipeline_run.go
davidmogar: You are setting Annotations here so probably the `WithReleaseAndApplicationLabels` function should be renamed to `WithReleaseAndApplicationMetadata`.

tekton/pipeline_run.go
davidmogar: Giving this a second thought, you use it twice in the same function... we should probably extract it already. Thinking on the best way to do this package naming wise.

davidmogar: We could have a `metadata` package with a `filter.go` file. Inside that file we would have two exported functions:
- `GetAnnotationsWithPrefix(object client.Object, prefix string)`
- `GetLabelsWithPrefix(object client.Object, prefix string)`

You could still have an utility (non-exported) function in this new file for shared code.

What do you think @johnbieren?

tekton/pipeline_run_test.go
```diff
@@ -247,4 +247,59 @@ var _ = Describe("PipelineRun", func() {
...
+			release.ObjectMeta.Annotations[pipelinesAsCodeLabelPrefix+"/meow"] = "cat"
+			release.ObjectMeta.Labels = map[string]string{}
+			release.ObjectMeta.Labels[pipelinesAsCodeLabelPrefix+"/bark"] = "dog"
+
+			releasePipelineRun.WithPipelinesAsCodeMeta(release)
+
+			Expect(len(releasePipelineRun.ObjectMeta.Annotations)).To(Equal(1))
+			Expect(len(releasePipelineRun.ObjectMeta.Labels)).To(Equal(1))
+			Expect(releasePipelineRun.ObjectMeta.Annotations[pipelinesAsCodeLabelPrefix+"/meow"]).To(Equal("cat"))
+			Expect(releasePipelineRun.ObjectMeta.Labels[pipelinesAsCodeLabelPrefix+"/bark"]).To(Equal("dog"))
```
theflockers, line 263: You think you can use `releasePipelineRun.GetLabels()["key"]`. I don't know if GetAnnotations() would work as well.

theflockers (changes requested):
Just one required change. The other one is optional.

tekton/pipeline_run.go
parasense: That makes sense. I did try to make it safe, not clobbering things already setup. Regardless, I've moved the new code into `WithReleaseAndApplicationLabels` and per another remark renamed that to `WithReleaseAndApplicationMetadata`. PTAL

tekton/pipeline_run.go
parasense: This function was deleted, and the code moved into `WithReleaseAndApplicationMetadata`. PTAL

tekton/pipeline_run.go
parasense: Good idea. I created a separate `metadata` package, and exposed an interface containing `GetAnnotationsWithPrefix` and `GetLabelsWithPrefix`, and two other exported methods to Copy Annotations & Labels metadata. The interface is mocked in `pipeline_run_test` so not to introduce any external effects on existing code, and fully unit tested & covereed in it's own package. PTAL.

tekton/pipeline_run_test.go
parasense: :smiling_face_with_tear: Regrettably that test was removed during a big refactor. 
:smile: The good news is the tests moved to a new location.
PTAL into `metadata/filter_test.go`

tekton/pipeline_run_test.go
```diff
@@ -247,4 +247,59 @@ var _ = Describe("PipelineRun", func() {
 			Expect(releasePipelineRun.Spec.Params).Should(ContainElement(HaveField("Value.StringVal", Equal(string(jsonSpec)))))
 		})
 	})
+
+	Context("When PipelineAsCode metadata is present in the Release", func() {
+		//
```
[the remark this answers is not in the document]
parasense: Oh yeah, that probably should not have been there. :man_facepalming: 
That probably got deleted during the big refactor, but I'm sure there are similar things in the new tests.
If you get a chance, I'd appreciate if you can look at `metadata/filter*.go` :+1:

metadata/filter.go
```diff
@@ -0,0 +1,110 @@
...
+func NewMetaGet() MetaGet {
+	return metadata{}
+}
+
+// NewMetaCopy is a factory function that return a MetaCopy interface.
+func NewMetaCopy() MetaCopy {
+	return metadata{}
+}
+
+func (metadata) GetAnnotationsWithPrefix(src MetaMap, prefix string) MetaMap {
```
davidmogar, line 58: These functions should have no receivers. These are utility functions and implementing them this way make its use more difficult. Interface, MetaMap, structs and `New*` functions should be removed from this file.

metadata/filter.go
```diff
@@ -0,0 +1,110 @@
...
+	} else {
+		for key, val := range src {
+			safeCopy(dst, key, val)
+		}
+	}
+}
+
+// getMetaMapWithPrefix returns a map of key/value pairs contained in src that matches the prefix.
+// When the prefix is not empty, and src key does not contain the prefix string, no copy happens.
+func getMetaMapWithPrefix(src MetaMap, prefix string) MetaMap {
```
davidmogar, line 98: Just use `map[string]string` and rename to `filterByPrefix`. Create a copy though so the original map remains intact.

metadata/filter.go
```diff
@@ -0,0 +1,110 @@
...
+	dst := MetaMap{}
+	if prefix != "" {
```
davidmogar, line 100: Not really needed as any string starts with `""`, which is fine. Remove this check please.

controllers/release/release_adapter.go
```diff
@@ -315,7 +316,7 @@ func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStra
 	applicationSnapshot *applicationapiv1alpha1.ApplicationSnapshot) (*v1beta1.PipelineRun, error) {
 	pipelineRun := tekton.NewReleasePipelineRun("release-pipelinerun", releaseStrategy.Namespace).
 		WithOwner(a.release).
-		WithReleaseAndApplicationLabels(a.release.Name, a.release.Namespace, a.release.GetAnnotations()[logicalcluster.AnnotationKey], applicationSnapshot.Spec.Application).
+		WithReleaseAndApplicationMetadata(a.release.Name, a.release.Namespace, a.release.GetAnnotations()[logicalcluster.AnnotationKey], applicationSnapshot.Spec.Application, a.release, metadata.NewMetaCopy()).
```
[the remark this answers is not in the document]
davidmogar:
```suggestion
		WithReleaseAndApplicationMetadata(a.release, applicationSnapshot.Spec.Application).
```

metadata/filter.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (metadata) GetAnnotationsWithPrefix(src MetaMap, prefix string) MetaMap {
+	return getMetaMapWithPrefix(src, prefix)
+}
+func (metadata) GetLabelsWithPrefix(src MetaMap, prefix string) MetaMap {
+	return getMetaMapWithPrefix(src, prefix)
+}
+
+func (metadata) CopyAnnotationsWithPrefix(src, dst MetaMap, prefix string) {
```
davidmogar, line 65: Let's keep the copy functions out of this file. This file is for object metadata filtering purposes not to copy/filter generic maps.

Those 4 functions could be replaced with [samber/lo#Assing](https://pkg.go.dev/github.com/samber/lo#Assign) in the `pipeline_run.go` file directly. Something like:
```
lo.Assing(r.Annotations, GetAnnotationsWithPrefix(release)
lo.Assing(r.Labels, GetLabelsWithPrefix(release)
```

metadata/filter.go
davidmogar: This function (and the similar one for labels) should receive a `client.Object`, not the map itself. Right now, both functions are exactly the same no matter what you pass.

davidmogar (changes requested):

controllers/release/release_adapter.go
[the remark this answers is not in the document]
parasense: I was hoping you might mention this, because I was thinking the same thing. But I didn't want to intrude into this function signature too much, so I was simply appending arguments. I agree that collapsing the common args to a single `Release` instance makes more sense, and honestly I'm surprised it wasn't already that way before. The only weird part is dealing with the `logicalcluster` bits inside `WithReleaseAndApplicationMetadata`, as I'm unsure if that violates any sense of separation, compartmentalization, or whatever... I'll get this added in the next refactor today. :+1:

metadata/filter.go
parasense: > These functions should have no receivers. These are utility functions and implementing them this way make its use more difficult.

Actually, this implementation probably makes things much easier.
That is mainly because interface methods are called the same way as a normal function, without any difficulty.
However, if you mean the added overhead of defining an empty struct, using a receiver prefix, etc... then I agree there is some overhead cost associated with the implementation. But the overhead is minimal, straightforward idiomatic GO, and not too difficult to wrap one's head-around.

The main benefit of coding this way involves testing, because using an interface is probably a **_bazillion_** :open_mouth:   times easier to test, isolate, and mock. That way these functions/methods don't influence other tests in the codebase where they can be easily **mocked** away. Also, we probably do not want to retest these functions multiple times, for example during any unit test of functions that utilize these methods, because all the unit tests are done in `metadata/filter_test.go`.

metadata/filter.go
parasense: > Just use map[string]string 
That is fine, but makes the code less concise or aesthetic, but arguably more concrete for lack of symbolic types. 

> and rename to `filterByPrefix`
Sure, we can rename to `filterByPrefix` or anything else. 

> Create a copy though so the original map remains intact.
I believe it does return a copy, because `dst` is a new map, that contains the prefix matches.
Or perhaps I misunderstood? Do you mean create a copy of the `src` map before iterating over it, or maybe passing a copy of `src` into this function ahead of calling the function? Regardless, making a copy of `src` seems needlessly non-optimal, but if there is a concern about corruption then it's probably worth the performance loss. Not a biggy, but I'd like to better understand this one, so if you could please elaborate, that would be great?

tekton/pipeline_run.go
```diff
@@ -142,8 +146,8 @@ func (r *ReleasePipelineRun) WithOwner(release *v1alpha1.Release) *ReleasePipeli
 	return r
 }
 
-// WithReleaseAndApplicationLabels adds Release and Application labels to the release PipelineRun.
-func (r *ReleasePipelineRun) WithReleaseAndApplicationLabels(releaseName, releaseNamespace, releaseWorkspace string, applicationName string) *ReleasePipelineRun {
+// WithReleaseAndApplicationMetadata adds Release and Application labels to the release PipelineRun.
+func (r *ReleasePipelineRun) WithReleaseAndApplicationMetadata(releaseName, releaseNamespace, releaseWorkspace, applicationName string, release *v1alpha1.Release, md metadata.MetaCopy) *ReleasePipelineRun {
```
johnbieren, line 150: If we're going to pass the entire release object in, can we stop passing in the name, namespace, and workspace of the release and just retrieve those values from the release object?

metadata/filter.go
parasense: Yeah I know, right?  :sunglasses: 
I hate that condition too... and I'd love to get rid of it, but...
_That condition is 100% required for the code to work as intended_.
It's actually a per performance optimization as much as a sanity check.

Instead of explaining in a comment, I'll provide a quick and dirty implementation in the playground to illustrate why this happens. https://go.dev/play/p/F-k_eRQbPwY

metadata/filter.go
johnbieren: I am also confused as to why we have the interfaces and structs and all this extra stuff in here now. Before (I had actually approved it but missed a bug I guess) there was much less new code, I found it very easy to follow, and it only had I think two new functions. Why do we need all this extra stuff now if we didn't before?

johnbieren (changes requested):

metadata/filter.go
parasense: > Those 4 functions could be replaced with [samber/lo#Assing](https://pkg.go.dev/github.com/samber/lo#Assign)

"assing" :rofl: :laughing: 
I totally get what you mean though, we could "assign" the data another way. 
Or, we could do it right here in this so-called "utility" package.
My feeling is if we are defining tools to get or filter _Labels & Annotations_ by prefix, we may as well assign them too, or at least make it convenient. I don't think we need to add another import. So I'll take a look at doing the merge in the calling code.

metadata/filter.go
davidmogar:
```
That condition is 100% required for the code to work as intended.
```
Could you comment on this? Removing the check would have the same behaviour. It would just be slower, but we don't care. We are just filtering around 10 annotations/labels in the worst case. Not a big deal.

If you really want to keep it, rewrite it to something like this:
```
	if prefix == "" {
		return src
	}
	
	for key, val := range src {
		if strings.HasPrefix(key, prefix) {
			dst[key] = val
		}
	}

	return dst
```
Also, AFAIK provided you delete the key you are currently looping over, it's safe to do so in Go. So the function would become:
```
func filterByPrefix(entries map[string]string, prefix string) map[string]string {
	if prefix != "" {
		for key, val := range entries {
			if !strings.HasPrefix(key, prefix) {
				delete(entries, key)
			}
		}
	}
	
	return entries
}
```
I renamed `src` to `entries` as in this version it makes more sense IMHO.

davidmogar (changes requested):

tekton/metadata_test.go
```diff
@@ -0,0 +1,30 @@
+/*
```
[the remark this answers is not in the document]
parasense: It _mocks_ the metacopy interface, to avoid influencing tests in the scope of the `tekton` package.

The metadata package functions are NOT tested in `tekton/pipeline_run_test.go`, but rather in the package tests in `metadata/filter_test.go`. 

This kind of thing is a common GO pattern to keep things separated, I guess sorta like OOP in other languages. There are some really cool capabilities this kind of thing allows, for example we could instrument these stub functions to have a `call_counter` , giving the ability for unit tests to do neat things like validating the expected number of function calls. There are packages to do exactly that:
https://pkg.go.dev/github.com/stretchr/testify/mock
https://pkg.go.dev/github.com/golang/mock/gomock
https://github.com/vektra/mockery

tekton/pipeline_run.go
parasense: Yep. That's the plan :wink:

metadata/filter.go
parasense: Okay, yeah... I refactored it to do the zero-copy NOOP near the top, like the first example you gave.

metadata/filter.go
parasense: Take another look.
I've simplified the copy functions in the most recent refactor.
They now receive the `Release` * `PipelineRun` objects directly. 

I don't think we need to export the `Get*` functions at all, but they are still exported, and tested for now.

metadata/filter.go
parasense: I've refactored the unexported functions to make an internal copy of the source map, so there is less chance for corruption on the pointer, or anything like that. PTAL

metadata/filter.go
parasense: > I don't think it makes things easier.

**¯\\_(ツ)_/¯**

But it really does make thing easier, testing is like a **_bazillion_** times easier this way. 
I think what you probably mean is "it makes things uglier", and I would agree.

My approach is to make code easily testable, but I am trying my best to make it look less ugly. Take a look at the most recent refactor, and let me know what you think...

I've incorporated more of your suggestions.
The receiver methods are now setup to receive the `Release` & `PipelineRun` objects more directly.

davidmogar: No, I meant what I said. Code should be easier to use specially with utility functions. Being easier to test comes after. I still would like to see the changes I suggested applied.

tekton/metadata_test.go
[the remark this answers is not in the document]
davidmogar: I don't understand it either tbf. In any case I still think those interfaces and structs should go away, which will remove the need for this file altogether.

metadata/filter.go
davidmogar: Names are completely different and in my opinion difficult to understand. They are also missing comments, which is a must specially for exported functions.

metadata/filter.go
davidmogar:
```
func filterByPrefix[M map[K]V, K, V, S ~string](entries M, prefix S) M {
```
I don't understand why to overengineer something that is specifically made for annotations and labels. Those are **always** going to be a `map[string]string`. This makes the code more complex to understand which is what I've been saying on my comments.

metadata/filter.go
davidmogar: Also, please, client.Object. Doesn't make any sense to use interfaces and templating on a utility file (which I don't agree with) and then pass specific types.

metadata/filter.go
parasense: The more recent refactors:

-  adds receivers for `Release` * `PipelineRun` which makes things much better :+1: 
-  removes the `Get*` functions entirely, because in the current form of this patch set, they were not being used at all. 

In my humble opinion, we probably do not need separate `getters` or `setters`, but I'm warm to the idea because that would [conform to how upstream k8s organizes](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/meta.go) everything in `ObjectMeta` with `gets` and `sets` on structure fields. The rational is based on the fact that we wanted to relocate helper functions to copy PAC metadata out of the `tekton` package. Also, an `Add*` method is just a `Copy*` by another name, so it comes down to choosing the name, and I'm indifferent on the names.

> The last two functions would be called with the result of the first ones

That could probably be achieved without much thrash & churn to the current design. So I'll give it a whirl on a local feature branch and see how it goes, and I'll ping you when it's pushed.

parasense: > That could probably be achieved without much thrash & churn to the current design. So I'll give it a whirl on a local feature branch and see how it goes, and I'll ping you when it's pushed.

Ping @davidmogar the get/add methods were incorporated in the most recent refactor (v). PTAL.

metadata/filter.go
parasense: See the most recent refactor (v).

- Method signatures uses the `Object` instead of maps.
- Should be easy to use: 
  - The `Copy*` method can do everything in one shot, or 
  - Using `Get*` + `Add*` methods together achieve the same.

PTAL

parasense:
Hey gang.
The recent refactor included many changes, and hopefully converges towards completion... PTAL. :smiley:

metadata/metadata.go
```diff
@@ -0,0 +1,86 @@
...
+*/
+
+package metadata
+
+import (
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strings"
+)
+
+// AddAnnotations is a method that copies a map into a ReleasePipelineRun.
```
johnbieren, line 24: can we say it copies a map into a resource's annotations? It doesn't seem to be specific to a release pipelinerun

metadata/metadata.go
```diff
@@ -0,0 +1,86 @@
...
+// When the destination map is nil, then the map will be created.
+// The unexported function addEntries is called with args passed.
+func AddAnnotations(obj v1.Object, entries map[string]string) {
+	if obj.GetAnnotations() == nil {
+		obj.SetAnnotations(map[string]string{})
+	}
+	addEntries(entries, obj.GetAnnotations())
+}
+
+// AddLabels is a method that copies a map into a ReleasePipelineRun.
```
johnbieren, line 34: Same comment as AddAnnotations

tekton/pipeline_run.go
```diff
@@ -48,6 +50,9 @@ const (
 
 	//PipelineTypeRelease is the type for PipelineRuns created to run a release Pipeline
 	PipelineTypeRelease = "release"
+
+	// pipelinesAsCodeLabelsPrefix is the prefix for pipelines-as-code metadata
```
johnbieren, line 54: Seems it should say `pipelinesAsCodeMetadataPrefix` not `pipelinesAsCodeLabelsPrefix`

tekton/pipeline_run.go
```diff
@@ -144,16 +149,18 @@ func (r *ReleasePipelineRun) WithOwner(release *v1alpha1.Release) *ReleasePipeli
 	return r
 }
 
-// WithReleaseAndApplicationLabels adds Release and Application labels to the release PipelineRun.
-func (r *ReleasePipelineRun) WithReleaseAndApplicationLabels(releaseName, releaseNamespace, releaseWorkspace string, applicationName string) *ReleasePipelineRun {
+// WithReleaseAndApplicationMetadata adds Release and Application labels to the release PipelineRun.
```
johnbieren, line 152: should we say it adds `metadata` instead of it adds `labels` to the rpr?

tekton/pipeline_run_test.go
```diff
@@ -187,11 +182,7 @@ var _ = Describe("PipelineRun", func() {
 		})
 
 		It("can append the release Name, Namespace, and Workspace to a ReleasePipelineRun object and that these label key names match the correct label format", func() {
-			releasePipelineRun.WithReleaseAndApplicationLabels(
-				release.Name,
-				release.Namespace,
-				release.GetAnnotations()[logicalcluster.AnnotationKey],
-				applicationName)
+			releasePipelineRun.WithReleaseAndApplicationMetadata(release, applicationName)
```
johnbieren, line 185: should we add an example PaC label/annotation to the release in line 88 and check if they are copied over?

tekton/pipeline_run_test.go
```diff
@@ -247,4 +238,35 @@ var _ = Describe("PipelineRun", func() {
 			Expect(releasePipelineRun.Spec.Params).Should(ContainElement(HaveField("Value.StringVal", Equal(string(jsonSpec)))))
 		})
 	})
+
+	Context("WithReleaseStrategy handles all limbs of PVC conditional branch", func() {
```
johnbieren, line 242: These look good, but I am not sure how they relate. Is this just you improving test coverage in the file for other functions you didn't touch?

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+)
+
+func TestMetadata(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Metadata Test Suite")
+}
+
+var _ = Describe("Metadata", func() {
+
+	Context("addEntries function", func() {
```
johnbieren, line 35: These tests seem to be the same as the `Context("filterByPrefix function", func() {` ones. Further, these tests don't call the `addEntries` function. Is this intended? They are good tests for `filterByPrefix`, but I am not sure they test `addEntries`

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+			It("should not set the src key/value in dst map", func() {
+				Expect(dst["foo/dog"]).To(Equal("bark"))
+			})
+			It("should not increased the length of the map", func() {
+				Expect(len(dst)).To(Equal(1))
+			})
+		})
+	})
+
+	//
```
johnbieren, line 125: I don't think we need this comment. The Context makes it clear

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{
+						"pet/dog":   "bark",
+						"pet/cat":   "meow",
+						"pond/frog": "ribit",
+					},
+					Labels: map[string]string{},
+				},
+			}
+			GetAnnotationsWithPrefix(release, "")
```
johnbieren, line 259: shouldn't you be setting the return value of this equal to something and checking the length of that?

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+					Annotations: map[string]string{
+						"pet/dog":   "bark",
+						"pet/cat":   "meow",
+						"pond/frog": "ribit",
+					},
+					Labels: map[string]string{},
+				},
+			}
+			dst := GetAnnotationsWithPrefix(release, "pet/")
+			It("should copy all src key/value pairs into dst that do not already exists", func() {
```
johnbieren, line 276: can we say release instead of src?

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+				Expect(dst["pet/dog"]).To(Equal("bark"))
+				Expect(dst["pet/cat"]).To(Equal("meow"))
+			})
+			It("should have increased the length of the dst map", func() {
```
johnbieren, line 280: dst was just created, it never had a length with anything other than 2. Can we say like `should create a dst map with length of 2`?

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{},
+					Labels: map[string]string{
+						"pet/dog":   "bark",
+						"pet/cat":   "meow",
+						"pond/frog": "ribit",
+					},
+				},
+			}
+			GetLabelsWithPrefix(release, "")
```
johnbieren, line 299: I think you forgot `dst := `

tekton/pipeline_run_test.go
parasense: Yeah, it's technically unrelated to hacbs-1193.
However, I was making edits to `pipeline_run.go` and noticed missing coverage that was super trivial.
So I wanted to leave the file better than I found it.
We could remove these lines if you want, I don't care.

tekton/pipeline_run_test.go
johnbieren: Admittedly I wrote that comment before I went through metadata_test.go. I think it could be nice to add here, similarly to how we check the other annotations we expect `WithReleaseAndApplicationMetadata` to add, but it is not required if you don't want to add it

tekton/pipeline_run_test.go
johnbieren: No, I think they are great to have, just wanted to double check. In an ideal world I think they are part of a separate commit, but could still be in the same PR (one commit with all the PaC stuff, one commit fixing this test coverage) but making it two separate commits may be a pain now and I don't think it matters too much

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+				Expect(len(dst)).To(Equal(1))
+			})
+		})
+	})
+
+	//
+	// Add* methods
+	//
+	Context("metadata.AddAnnotations method", func() {
+		When("calls the addEntries function with an empty map", func() {
```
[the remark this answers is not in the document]
parasense: `AddAnnotations` is really just a wrapper for the `addEntries` function, and so it does _**call**_ the `addEntries` function with with parameters... with a non-empty map. But I can see  the confusion. I'm going to remove the specifics and go with empty v non-empty maps.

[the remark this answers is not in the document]
johnbieren: perfect, thanks. I guess my point is that we don't really care (here at least) it calls `addEntries` underneath. If we changed that function name one day but didn't change `AddAnnotations` name, I wouldn't expect to have to change this. So big +1 to empty v non-empty maps

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: map[string]string{"pet/dog": "ruff"},
+				},
+			}
+			AddAnnotations(pipelineRun, map[string]string{})
+			It("src and dst maps should have the same length", func() {
+				Expect(len(pipelineRun.GetAnnotations())).To(Equal(1))
+			})
+		})
+		When("calls the addEntries function with a non-empty map", func() {
```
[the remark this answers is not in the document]
parasense: So FYI - after looking at this review topic, I realized this area of code was redundant, and most of this test was deleted. This code was duplicating the tests for `addEntries`. Once my next refactor uploads, this review might show as outdated.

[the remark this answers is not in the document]
johnbieren: ack, I'll give the entire metadata test file a lookover again once the update is pushed

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+				},
+			}
+			pipelineRun := &v1beta1.PipelineRun{
+				ObjectMeta: v1.ObjectMeta{
+					Annotations: nil,
+					Labels:      nil,
+				},
+			}
+			AddLabels(pipelineRun, release.GetAnnotations())
+			It("should create the Annotations map", func() {
```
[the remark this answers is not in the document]
parasense: Good catch.
Turns out the `release` object wasn't really needed in that test anymore, so I've updated this to the concrete type `map[string]string`

metadata/metadata_test.go
```diff
@@ -0,0 +1,326 @@
...
+					},
+				},
+			}
+			pipelineRun := &v1beta1.PipelineRun{
+				ObjectMeta: v1.ObjectMeta{
+					Labels: map[string]string{"pet/dog": "ruff"},
+				},
+			}
+			AddLabels(pipelineRun, release.GetLabels())
+			It("should copy all src key/value pairs into dst that do not already exist", func() {
```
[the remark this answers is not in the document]
parasense: As I write this, I've refactored most of these test strings to be less pithy.
But some of the `When` and `It` closures might have got wacked.

metadata/metadata_test.go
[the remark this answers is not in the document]
johnbieren: This line should still have Annotations replaced with Labels

tekton/pipeline_run_test.go
parasense: In that case I'll mark it resolved, and I'm happy to revisit the topic another time.

metadata/metadata_test.go
parasense: I'm pretty sure this was factored away in the latest refactor. It was on line 280, and the file is now only ~210 lines. All these wrapper functions were reduced.

metadata/metadata_test.go
parasense: It was removed in the refactor, but please take another look as you seem really good at catching my subtle bugs and grammar mistakes.

johnbieren (commented):
My comment on line 159 on metadata_test.go is my only outstanding thing to fix

metadata/metadata_test.go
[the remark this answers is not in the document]
parasense: Thanks!
I apparently missed this one in the most recent refactor, but actually mistakenly _**thought**_ I had fixed it. To use a cliche, it got swept under the rug... So my apologies if I resolved this one by mistake. :disappointed: 
I appreciate your relentless attention to details like this. :+1: 

FYI - instead of stacking another commit, I'm simply force pushing over top, and clobbering the previous. So this may show as "outdated".

johnbieren (approved):

PR #124, opened 2022-11-16
fix: remove double prefix on rbac resources

davidmogar:
For kcp we needed to move the name-prefix to the rbac kustomization. Now that kcp is gone, we are getting double prefixes on our deployments. This is causing problems in some resources which end up with a really long name.

### DIFF
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -1,5 +1,3 @@
-namePrefix: release-service-
-
 resources:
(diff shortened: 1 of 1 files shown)


scoheb (approved):

johnbieren (changes requested):
I think it makes way more sense to keep all the KCP removal stuff in one PR and not separate it across multiple for tracking and history purposes

PR #126, opened 2022-11-17
feat(hacbs-1414): change pac metadata prefix

parasense:
Resolves: hacbs-1414
Related: hacbs-1385


tekton/pipeline_run.go
```diff
@@ -50,7 +50,7 @@ const (
 	PipelineTypeRelease = "release"
 
 	// pipelinesAsCodeMetadataPrefix is the prefix for pipelines-as-code metadata
-	pipelinesAsCodeMetadataPrefix = "pipelinesascode.tekton.dev"
+	pipelinesAsCodeMetadataPrefix = "pac.test.appstudio.openshift.io"
```
scoheb, line 53: Why **test**?

johnbieren: replying on slack

johnbieren:
I've added the do-not-merge label as we shouldn't merge this until https://github.com/redhat-appstudio/integration-service/pull/69 merges

johnbieren:
@parasense the tests are flaky right now. If you click details on the check sources check, in the top right you can click the circular arrows button to rerun it (without having to amend this PR). I've been doing that until I get it to pass, in order to get by the flakiness

tekton/pipeline_run.go
parasense: Quoting from [hacbs-1385](https://issues.redhat.com/browse/HACBS-1385):

> Acceptance Criteria
>
>The integration controller replaces the prefix ("pipelinesascode.tekton.dev/") on the PaC labels with the AppStudio test prefix ("pac.test.appstudio.openshift.io/") when adding them to the Integration PipelineRun.

The **test** string is also reflected in the integration-service PR:
https://github.com/redhat-appstudio/integration-service/pull/69

johnbieren (changes requested):
Now that the integration-service PR changed, I think we should just import the constant from their service (which won't work in an IDE until they are merged)

tekton/pipeline_run.go
```diff
@@ -50,7 +51,7 @@ const (
 	PipelineTypeRelease = "release"
 
 	// pipelinesAsCodeMetadataPrefix is the prefix for pipelines-as-code metadata
-	pipelinesAsCodeMetadataPrefix = "pipelinesascode.tekton.dev"
+	pipelinesAsCodeMetadataPrefix = integrationServiceGitopsPkg.PipelinesAsCodePrefix
```
johnbieren, line 54: I think it would be better to not define this constant here and instead use `integrationServiceGitopsPkg.PipelineAsCodePrefix` everywhere `pipelinesAsCodeMetadataPrefix` is currently used

parasense: That works great for me. Thanks.

johnbieren (approved):
lgtm once the CI passes

PR #128, opened 2022-11-22
chore(hacbs_1032): move ReleasePlan and ReleasePlanAdmission label to a proper place

theflockers:
Looking around some `k8s` projects I noticed they use `labels.go` to keep their labels in. Not sure about the proper place and if should we move any other label here so far.

- moves release related labels to `labels.go` at `api/v1alpha1`

### DIFF
--- a/api/v1alpha1/release_types.go
+++ b/api/v1alpha1/release_types.go
@@ -17,9 +17,10 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"time"
+
 	"github.com/redhat-appstudio/release-service/conditions"
 	"github.com/redhat-appstudio/release-service/metrics"
-	"time"
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -37,11 +38,6 @@ type ReleaseSpec struct {
 	// +required
 	ReleasePlan string `json:"releasePlan"`
 }
-
-const (
-	// AutoReleaseLabel is the label name for the auto-release setting
-	AutoReleaseLabel = "release.appstudio.openshift.io/auto-release"
-)
 
 // ReleaseStatus defines the observed state of Release.
 type ReleaseStatus struct {
--- a/metadata/labels.go
+++ b/metadata/labels.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2022.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import "fmt"
+
+// Common constants
+const (
+	// rhtapDomain is the prefix of the application label
+	rhtapDomain = "appstudio.openshift.io"
+)
+
+// Labels used by the release api package
+var (
+	// AutoReleaseLabel is the label name for the auto-release setting
+	AutoReleaseLabel = fmt.Sprintf("release.%s/auto-release", rhtapDomain)
+
+	// AuthorLabel is the label name for the user who creates a CR
+	AuthorLabel = fmt.Sprintf("release.%s/author", rhtapDomain)
+)
+
+// Prefixes to be used by Release Pipeline Labels
+var (
+	// pipelinesLabelPrefix is the prefix of the pipelines label
+	pipelinesLabelPrefix = fmt.Sprintf("pipelines.%s", rhtapDomain)
+
+	// releaseLabelPrefix is the prefix of the release labels
+	releaseLabelPrefix = fmt.Sprintf("release.%s", rhtapDomain)
+)
+
+// Labels to be used within Release PipelineRuns
+var (
+	// ApplicationNameLabel is the label used to specify the application associated with the PipelineRun
+	ApplicationNameLabel = fmt.Sprintf("%s/%s", rhtapDomain, "application")
+
+	// PipelinesTypeLabel is the label used to describe the type of pipeline
+	PipelinesTypeLabel = fmt.Sprintf("%s/%s", pipelinesLabelPrefix, "type")
+
+	// ReleaseNameLabel is the label used to specify the name of the Release associated with the PipelineRun
+	ReleaseNameLabel = fmt.Sprintf("%s/%s", releaseLabelPrefix, "name")
+
+	// ReleaseNamespaceLabel is the label used to specify the namespace of the Release associated with the PipelineRun
+	ReleaseNamespaceLabel = fmt.Sprintf("%s/%s", releaseLabelPrefix, "namespace")
+)
--- a/metadata/metadata_test.go
+++ b/metadata/metadata_test.go
@@ -17,12 +17,13 @@ limitations under the License.
 package metadata
 
 import (
+	"testing"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"github.com/redhat-appstudio/release-service/api/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1"
-	"testing"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func TestMetadata(t *testing.T) {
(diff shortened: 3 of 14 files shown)


davidmogar:
Not against this change but if we go this direction I’d like to see some examples and give a think about the implications for other parts of the base code. Could you share some of those k8s projects?

theflockers:
Sure @davidmogar! So far it did not impact anything else because I kept it in the same package, although I saw some examples of `labels` (like for apimachinery) as a full package, meaning that all labels for all stuff could be contained there.  I did not considered it so far because the `labels.go` looked simpler and as for instance in`kubelet` 

https://github.com/kubernetes/apimachinery/blob/master/pkg/labels/labels.go
https://github.com/kubernetes/test-infra/blob/e574f0fd8a116eb2f7e19235295ca80d0374070b/prow/labels/labels.go
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/types/labels.go

johnbieren:
I'm fine with this, although I think it would make a little more sense if we had more labels to put there. Maybe we will over time

metadata/metadata_test.go
```diff
@@ -148,26 +149,26 @@ var _ = Describe("Metadata", func() {
 	})
 
 	Context("AddLabels function", func() {
-		When("called with a pipelineRun containing nil Labels", func() {
-			pipelineRun := &v1beta1.PipelineRun{
+		When("called with an object containing nil Labels", func() {
+			pod := &corev1.Pod{
```
scoheb, line 153: why are using `pod` here?
is it just to simplify the test case involving `GetLabelsWithPrefix`?

theflockers: We were using `Release` or `Pipeline` resources for it. The problem is that  by moving the labels to `metadata` package, when importing the `metadata`it  causes a cycle import error. Using the Pod instead, we avoid the error and keep the test the way it was initially written.

metadata/labels.go
```diff
@@ -0,0 +1,58 @@
...
+var (
+	// AutoReleaseLabel is the label name for the auto-release setting
+	AutoReleaseLabel = fmt.Sprintf("release.%s/auto-release", appstudioLabelPrefix)
+)
+
+// Labels used by releases's tekton Pipeline package
+const (
+
+	// pipelinesLabelPrefix is the prefix of the pipelines label
+	pipelinesLabelPrefix = "pipelines.appstudio.openshift.io"
```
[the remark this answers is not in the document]
theflockers: I discussed these ones with @davidmogar  and we agreed in keep these constants like this. Although I might have missed something.

metadata/labels.go
```diff
@@ -0,0 +1,58 @@
...
+
+	// releaseLabelPrefix is the prefix of the release labels
+	releaseLabelPrefix = "release.appstudio.openshift.io"
```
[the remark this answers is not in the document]
theflockers: same as above.

metadata/metadata_test.go
johnbieren: And, if I am not wrong, the tests are just about adding labels and annotations, so the object itself does not matter. Pod is the most basic object we can use for tests AIUI. +1 for using pod

tekton/pipeline_run.go
```diff
@@ -37,34 +36,6 @@ import (
...
-	appstudioLabelPrefix = "appstudio.openshift.io"
-
-	// pipelinesLabelPrefix is the prefix of the pipelines label
-	pipelinesLabelPrefix = "pipelines.appstudio.openshift.io"
-
-	// releaseLabelPrefix is the prefix of the release labels
-	releaseLabelPrefix = "release.appstudio.openshift.io"
-
-	//PipelineTypeRelease is the type for PipelineRuns created to run a release Pipeline
-	PipelineTypeRelease = "release"
```
johnbieren, line 51: David told me that we were only moving labels and annotations to metadata, and other constants, such as this one which describes a type, would remain where it is. Not sure if that changed after I talked to him, it wasn't conveyed to you, or including this was a mistake

metadata/labels.go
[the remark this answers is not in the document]
johnbieren: Well, `AutoReleaseLabel` wasn't using `appstudioLabelPrefix` before, so +1 to what Scott said. If changing how that one is formatted is in scope of this PR, I think all the values in here should be changed similarly

metadata/labels.go
```diff
@@ -0,0 +1,58 @@
...
+
+	//PipelineTypeRelease is the type for PipelineRuns created to run a release Pipeline
+	PipelineTypeRelease = "release"
+)
+
+var (
```
johnbieren, line 46: Can you put a comment here too for what this grouping is for?

metadata/labels.go
```diff
@@ -0,0 +1,58 @@
...
+// Labels used by releases's tekton Pipeline package
+const (
+
+	// pipelinesLabelPrefix is the prefix of the pipelines label
+	pipelinesLabelPrefix = "pipelines.appstudio.openshift.io"
+
+	// releaseLabelPrefix is the prefix of the release labels
+	releaseLabelPrefix = "release.appstudio.openshift.io"
+
+	//PipelineTypeRelease is the type for PipelineRuns created to run a release Pipeline
```
johnbieren, line 42: Missing a space before after `//`

tekton/pipeline_run.go
theflockers: You are right. It was basically a full cut and paste. Did not pay attention that this one is not a label.

johnbieren (approved):

johnbieren:
@theflockers I am quite sure integration-service relies on some of these (at least auto release label almost surely). Can you open a PR to change the import for them too so they are not broken?
@dirgim FYI

theflockers:
> @theflockers I am quite sure integration-service relies on some of these (at least auto release label almost surely). Can you open a PR to change the import for them too so they are not broken? @dirgim FYI

I created a PR already for integration-service but not yet for e2e. I will take care as well.

metadata/labels.go
```diff
@@ -0,0 +1,55 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+import "fmt"
+
+// Common constants
+const (
```
davidmogar, line 22: Can you move unexported variables to the bottom of the file? It would be nice to get into the labels.go file and see the actual labels.

davidmogar (changes requested):

metadata/labels.go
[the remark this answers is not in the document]
davidmogar: Yeah, actually what I said is totally the opposite. We should be form them using the already declared appstudio.openshift.io piece so when we change them to rhtap we don't need a bigger PR.

metadata/labels.go
```diff
@@ -0,0 +1,55 @@
...
+	// appstudioLabelPrefix is the prefix of the application label
+	appstudioLabelPrefix = "appstudio.openshift.io"
```
davidmogar, line 24: Can you rename the variable? This is not a prefix anymore. Change it to `rhtapDomain` or something similar. @johnbieren feel free to pitch in if you have a better proposal.

theflockers: I would change it to something like _applicationPrefix_ to not stick with any name.

theflockers:
Adding PR to apply the change to e2e-tests as well.

johnbieren (dismissed):

johnbieren (dismissed):

johnbieren (approved):

api/v1alpha1/release_types.go
```diff
@@ -38,11 +39,36 @@ type ReleaseSpec struct {
 	ReleasePlan string `json:"releasePlan"`
 }
 
+// ReleaseReason represents a reason for the release "Succeeded" condition.
+type ReleaseReason string
```
johnbieren, line 43: I think this ReleaseReason stuff was a rebase error. I don't see it in this file in main

johnbieren (approved):

johnbieren:
May be worth rebasing before going through all the attempts to get CI to pass, as it can't be merged as is anyways

metadata/labels.go
```diff
@@ -0,0 +1,55 @@
...
+*/
+
+package metadata
+
+import "fmt"
+
+// Common constants
+const (
+	// rhtapDomain is the prefix of the application label
+	rhtapDomain = "appstudio.openshift.io"
```
johnbieren, line 24: @scoheb https://github.com/redhat-appstudio/book/blob/main/ADR/0024-release-attribution.md#attribution-for-a-particular-release talks about `rhtap.openshift.io` as our "domain." Do we want the author and attribution labels to have this new rhtap domain or be the same as our other labels and use this appstudio domain?

scoheb: David is asking in slack...

davidmogar: An ADR will be prepared to address this. We can leave it as it is for the moment.

johnbieren: Okay, just pointing out in case it was missed, the author label uses this so this pr changes the author label domain from rhtap to appstudio. But we can align them all once the ADR is prepared

johnbieren (approved):

davidmogar (approved):

PR #130, opened 2022-11-27
refactor: move test utils to an external library

davidmogar:
These funcitons could be used by other operators, so I moved them to an external library and updated the code to fix a bug I saw early this week.


controllers/release/release_suite_test.go
```diff
@@ -31,13 +31,12 @@ import (
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
-	ctrl "sigs.k8s.io/controller-runtime"
-
 	ecapiv1alpha1 "github.com/hacbs-contract/enterprise-contract-controller/api/v1alpha1"
+	"github.com/redhat-appstudio/operator-goodies/test"
```
theflockers, line 35: I am wondering if you should name this module properly to resemble what it actually does. It made sense inside be named `test` inside  _release_service_ but outside it, it sounds/looks weird. What do you think?

davidmogar: Oh yeah, makes sense. When I moved it I only had the test stuff in there and I guess my mind flew to that. Great suggestion!

johnbieren (approved):

PR #131, opened 2022-11-29
feat(HACBS-1274): track GitOps deployment status in release

johnbieren:
This commit causes the release controller to trigger when a SnapshotEnvironmentBinding has its status changed to True or False. When this happened, it is reflected in the Release status.


gitops/predicates.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	goodies "github.com/redhat-appstudio/operator-goodies/predicates"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// GitopsDeploymentStatusChangedPredicate returns a predicate which filters out all events except
+// update events to a SnapshotEnvironmentBinding where the component deployment status goes from
+// unknown to true/false and the SnapshotEnvironmentBinding has not had its generation changed.
+func GitopsDeploymentStatusChangedPredicate() predicate.Predicate {
```
davidmogar, line 28: Functions shouldn't start with the name of the package.

api/v1alpha1/release_types.go
```diff
@@ -44,6 +45,10 @@ const (
 	// releaseConditionType is the type used when setting a release status condition
 	releaseConditionType string = "Succeeded"
 
+	// BindingDeploymentStatusConditionType is the condition type to retrieve from the ComponentDeploymentConditions
+	// in the SnapshotEnvironmentBinding's status to copy into the Release status
+	BindingDeploymentStatusConditionType string = "AllComponentsDeployed"
```
davidmogar, line 50: The type should not be exported. Same as the release type.

controllers/release/release_adapter.go
```diff
@@ -213,54 +214,59 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
 	}
 
-	err = a.syncResources()
+	environment, err := a.getEnvironment(releasePlanAdmission)
 	if err != nil {
 		return results.RequeueWithError(err)
 	}
 
-	binding, err := a.createOrUpdateSnapshotEnvironmentBinding(releasePlanAdmission)
+	err = a.syncResources()
```
davidmogar, line 222: Resources should be sync only whenever a binding is created, so it should go inside the `if binding == nil` check.

controllers/release/release_adapter.go
```diff
@@ -213,54 +214,59 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
+			return results.RequeueWithError(err)
+		}
+
+		a.logger.Info("Created/updated SnapshotEnvironmentBinding",
+			"SnapshotEnvironmentBinding.Name", binding.Name, "SnapshotEnvironmentBinding.Namespace", binding.Namespace)
+
+		a.release.Status.SnapshotEnvironmentBinding = fmt.Sprintf("%s%c%s", binding.Namespace, types.Separator, binding.Name)
+	}
 
 	return results.RequeueOnErrorOrContinue(a.client.Status().Patch(a.context, a.release, patch))
```
davidmogar, line 247: The release is only modified inside the `if` so the patch and the update of the status should go inside as well.

controllers/release/release_adapter.go
```diff
@@ -213,54 +214,59 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
 	patch := client.MergeFrom(a.release.DeepCopy())
-	a.release.Status.SnapshotEnvironmentBinding = fmt.Sprintf("%s%c%s", binding.Namespace, types.Separator, binding.Name)
+
+	if binding == nil {
+		binding, err := a.createSnapshotEnvironmentBinding(releasePlanAdmission, environment)
+		if err != nil {
+			return results.RequeueWithError(err)
+		}
+
+		a.logger.Info("Created/updated SnapshotEnvironmentBinding",
```
davidmogar, line 241:
```suggestion
		a.logger.Info("Created SnapshotEnvironmentBinding",
```

controllers/release/release_adapter.go
```diff
@@ -213,54 +214,59 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-	// The binding information needs to be updated no matter if it already exists or not
-	binding := gitops.NewSnapshotEnvironmentBinding(components, snapshot, environment)
-
 	// Search for an existing binding
-	existingBinding, err := a.getSnapshotEnvironmentBinding(environment, releasePlanAdmission)
-	if err != nil && !errors.IsNotFound(err) {
-		return nil, err
+	binding, err := a.getSnapshotEnvironmentBindingFromReleaseStatus()
+	if err != nil {
+		return results.RequeueWithError(err)
```
davidmogar, line 260: This forces the operation to work only with existing bindings which is not always the case. You should check if the error is a NotFound and work accordingly.

Alternatively and probably better, the first condition should look something like:
```
if release hasn't succeeded or status.binding is empty -> continue
```

To support this you could add a `IsBeingDeployed` function on the type. In addition, there's now a function, `HasBeenDeployed`, that has a terrible name taking in account what it does. This change sets `Status.SnapshotEnvironmentBinding` on binding creation time, so the logic should be changed to check for the condition status. This function should be used as a guard as well. If it was deployed already, no need to keep tracking the status.

api/v1alpha1/release_types.go
```diff
@@ -196,23 +201,29 @@ func (r *Release) MarkSucceeded() {
...
-	r.setStatusCondition(metav1.ConditionTrue, ReleaseReasonSucceeded)
+	r.setStatusCondition(releaseConditionType, metav1.ConditionTrue, ReleaseReasonSucceeded)
 
 	go metrics.RegisterCompletedRelease(ReleaseReasonSucceeded.String(), r.Status.ReleaseStrategy, r.Status.Target,
 		r.Status.StartTime, r.Status.CompletionTime, false)
 }
 
-// SetCondition creates a new condition with the given status and reason. Then, it sets this new condition,
+// RegisterDeploymentStatus registers the AllComponentsDeployed status from the SnapshotEnvironmentBinding to the release.
+func (r *Release) RegisterDeploymentStatus(binding *applicationapiv1alpha1.SnapshotEnvironmentBinding) {
```
davidmogar, line 211: The Release should know nothing about the binding. This function should be in the adapter and use Mark functions in here to set the status (eg. MarkInvalid, MarkRunning...). Good names for those would be MarkDeploying and MarkDeployed.

The function doing the same work for the Release PipelineRun is `registerReleasePipelineRunStatus` in the adapter.

controllers/release/release_adapter.go
```diff
@@ -213,54 +214,59 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsCreated() (results.Operation
...
-		patch := client.MergeFrom(existingBinding.DeepCopy())
-		existingBinding.Spec = binding.Spec
-		return existingBinding, a.client.Patch(a.context, existingBinding, patch)
+	patch := client.MergeFrom(a.release.DeepCopy())
+
+	if binding == nil {
+		a.release.RegisterDeploymentStatus(binding)
 	}
+
+	return results.RequeueOnErrorOrContinue(a.client.Status().Patch(a.context, a.release, patch))
```
davidmogar, line 269: Like in the previous one, the status is only modified within the `if`, so the patch creation and call should be inside.

controllers/release/release_adapter.go
```diff
@@ -286,6 +292,26 @@ func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStra
...
+	binding := gitops.NewSnapshotEnvironmentBinding(components, snapshot, environment)
+	err = ctrl.SetControllerReference(application, binding, a.client.Scheme())
+	if err != nil {
+		return nil, err
+	}
+	err = libhandler.SetOwnerAnnotations(a.release, binding)
+	if err != nil {
+		return nil, err
+	}
+	return binding, a.client.Create(a.context, binding)
```
davidmogar, line 312: Could you add an empty line before?

controllers/release/release_adapter.go
```diff
@@ -286,6 +292,26 @@ func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStra
...
+	if err != nil {
+		return nil, err
+	}
+
+	binding := gitops.NewSnapshotEnvironmentBinding(components, snapshot, environment)
+	err = ctrl.SetControllerReference(application, binding, a.client.Scheme())
+	if err != nil {
+		return nil, err
+	}
+	err = libhandler.SetOwnerAnnotations(a.release, binding)
```
davidmogar, line 308: This needs an empty line before and a comment stating this is added to watch for status updates of the binding so they can be tracked in the release.

controllers/release/release_adapter.go
```diff
@@ -286,6 +292,26 @@ func (a *Adapter) createReleasePipelineRun(releaseStrategy *v1alpha1.ReleaseStra
...
+// createSnapshotEnvironmentBinding creates a SnapshotEnvironmentBinding for the Release being processed.
+func (a *Adapter) createSnapshotEnvironmentBinding(releasePlanAdmission *v1alpha1.ReleasePlanAdmission,
+	environment *applicationapiv1alpha1.Environment) (*applicationapiv1alpha1.SnapshotEnvironmentBinding, error) {
+	application, components, snapshot, err := a.getSnapshotEnvironmentResources(releasePlanAdmission)
+	if err != nil {
+		return nil, err
+	}
+
+	binding := gitops.NewSnapshotEnvironmentBinding(components, snapshot, environment)
+	err = ctrl.SetControllerReference(application, binding, a.client.Scheme())
```
davidmogar, line 304: This needs an empty line before and a comment stating that this resource should be deleted if the application is deleted (I could have added it before in the original PR but it's never late).

controllers/release/release_controller.go
```diff
@@ -146,6 +151,12 @@ func setupControllerWithManager(manager ctrl.Manager, reconciler *Reconciler) er
 
 	return ctrl.NewControllerManagedBy(manager).
 		For(&v1alpha1.Release{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		Watches(&source.Kind{Type: &applicationapiv1alpha1.SnapshotEnvironmentBinding{}}, &libhandler.EnqueueRequestForAnnotation{
+			Type: schema.GroupKind{
+				Kind:  "Release",
+				Group: "appstudio.redhat.com",
+			},
+		}, builder.WithPredicates(gitops.DeploymentStatusChangedPredicate(), goodies.GenerationUnchangedOnUpdatePredicate{})).
```
davidmogar, line 159: Should the order be reversed? That way we filter more items I think.

davidmogar (changes requested):
Overall I think it's good but it needs changes. I haven't taken a look to any test yet. I'll do that after the changes on the logic.

api/v1alpha1/release_types.go
johnbieren: Needs to be. It has to be accessible in `gitops/utils.go`

davidmogar: Fair enough. Missed that.

controllers/release/release_controller.go
johnbieren: ? I got this from you. I didn't know order mattered here. What does order do?

davidmogar: My example was with a different trivial thing.

I haven't seen the code and I'm assuming predicates are executed in order so in this case the DeploymentStatusChangedPredicate would be executed first and would be applied for every single event. Then, the other predicate would be applied with the filtered events.

I have the impression that it would be more efficient to do it the other way around.

johnbieren: Sure thing, will swap them

gitops/utils.go
```diff
@@ -0,0 +1,46 @@
...
+			return false
+		}
+	}
+	if newBinding, ok := objectNew.(*applicationapiv1alpha1.SnapshotEnvironmentBinding); ok {
+		newCondition = meta.FindStatusCondition(newBinding.Status.ComponentDeploymentConditions, v1alpha1.BindingDeploymentStatusConditionType)
+		if newCondition == nil {
+			return false
+		}
+	}
+	return oldCondition.Status == metav1.ConditionUnknown && (newCondition.Status == metav1.ConditionTrue || newCondition.Status == metav1.ConditionFalse)
```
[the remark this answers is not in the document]
davidmogar: As pointed by @johnbieren offline, these are meta conditions so those functions are not there.

davidmogar (approved):

PR #135, opened 2022-12-01
feat(HACBS-1274): add deploy status helper functions to release CRD

johnbieren:
This commit adds functions to the Release CRD to help set its status to reflect the deployment status of the SnapshotEnvironmentBinding.

### DIFF
--- a/api/v1alpha1/release_types.go
+++ b/api/v1alpha1/release_types.go
@@ -44,6 +44,10 @@ const (
 	// releaseConditionType is the type used when setting a release status condition
 	releaseConditionType string = "Succeeded"
 
+	// BindingDeploymentStatusConditionType is the condition type to retrieve from the ComponentDeploymentConditions
+	// in the SnapshotEnvironmentBinding's status to copy into the Release status
+	BindingDeploymentStatusConditionType string = "AllComponentsDeployed"
+
 	// ReleaseReasonValidationError is the reason set when the Release validation failed
 	ReleaseReasonValidationError ReleaseReason = "ReleaseValidationError"
 
@@ -102,7 +106,7 @@ type ReleaseStatus struct {
 	// +optional
 	ReleaseStrategy string `json:"releaseStrategy,omitempty"`
 
-	// Target references where this relesae is intended to be released to
+	// Target references where this release is intended to be released to
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +optional
 	Target string `json:"target,omitempty"`
@@ -127,9 +131,10 @@ type Release struct {
 	Status ReleaseStatus `json:"status,omitempty"`
 }
 
-// HasBeenDeployed checks whether the Release has an associated SnapshotEnvironmentBinding.
+// HasBeenDeployed checks whether the Release has been deployed via GitOps.
 func (r *Release) HasBeenDeployed() bool {
-	return r.Status.SnapshotEnvironmentBinding != ""
+	condition := meta.FindStatusCondition(r.Status.Conditions, BindingDeploymentStatusConditionType)
+	return condition != nil && condition.Status != metav1.ConditionUnknown
 }
 
 // HasStarted checks whether the Release has a valid start time set in its status.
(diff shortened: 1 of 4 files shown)


api/v1alpha1/release_types.go
```diff
@@ -44,6 +44,13 @@ const (
 	// releaseConditionType is the type used when setting a release status condition
 	releaseConditionType string = "Succeeded"
 
+	// AutoReleaseLabel is the label name for the auto-release setting
+	AutoReleaseLabel = "release.appstudio.openshift.io/auto-release"
+
+	// BindingDeploymentStatusConditionType is the condition type to retrieve from the ComponentDeploymentConditions
+	// in the SnapshotEnvironmentBinding's status to copy into the Release status
+	BindingDeploymentStatusConditionType string = "AllComponentsDeployed"
```
davidmogar, line 52: This will probably change in the future but leave the AutoReleaseLabel were it was. The content of this block is used for conditions exclusively.

api/v1alpha1/release_types.go
```diff
@@ -145,11 +148,19 @@ func (r *Release) HasSucceeded() bool {
...
+// MarkDeployed sets the AllComponentsDeployed status in the Release to True or False with the provided reason and message.
+func (r *Release) MarkDeployed(status metav1.ConditionStatus, reason, message string) {
+	if status != metav1.ConditionUnknown {
+		r.setStatusConditionWithMessage(BindingDeploymentStatusConditionType, status, ReleaseReason(reason), message)
 	}
+}
 
-	return false
+// MarkDeploying sets the AllComponentsDeployed status in the Release to Unknown with the provided reason and message.
+func (r *Release) MarkDeploying(reason, message string) {
```
davidmogar, line 162: I don't think we should allow to pass generic reasons in here.

api/v1alpha1/release_types_test.go
```diff
@@ -122,6 +138,31 @@ var _ = Describe("Release type", func() {
 		})
 	})
 
+	Context("When MarkDeployed method is called", func() {
+		It("should properly register the status if passed status is true or false", func() {
+			r.MarkDeployed(metav1.ConditionTrue, "CommitsSynced", "3 of 3 components deployed")
+			statusCondition := meta.FindStatusCondition(r.Status.Conditions, BindingDeploymentStatusConditionType)
+			Expect(statusCondition.Status).To(Equal(metav1.ConditionTrue))
```
davidmogar, line 145: You could/should use MatchFields in here and in the other instances where you check the content of structs. You have an example in the `binding_test.go` file.

davidmogar (changes requested):

api/v1alpha1/release_types.go
johnbieren: Why? It isn't commented that in the constant blocks. There are already comments for each constant saying what it is for. Why would we want 
```
const (
 some constants
)
const (
 some more constants
)
```
in the same file?

api/v1alpha1/release_types.go
johnbieren: Why? I want to copy whatever reason gitops tells me

api/v1alpha1/release_types.go
davidmogar: Way more clear to have a block just for condition. Also, this particular change is unrelated and should not be part of this change.

api/v1alpha1/release_types.go
davidmogar: What you really want is the message which is the one containing the number of deployments. The reason... I think we should have some more control over it but I'm ok with copying over. Let's reevaluate whenever gitops give us some code.

johnbieren: No, that is why I pass message as well. I want both. But yes we can reevaluate later

api/v1alpha1/release_types.go
johnbieren: Disagree on way clearer, but agree it is an unrelated change so I will remove it

api/v1alpha1/release_types_test.go
```diff
@@ -122,6 +138,35 @@ var _ = Describe("Release type", func() {
...
+	Context("When MarkDeployed method is called", func() {
+		It("should properly register the status if passed status is true or false", func() {
+			r.MarkDeployed(metav1.ConditionTrue, "CommitsSynced", "3 of 3 components deployed")
+			statusCondition := meta.FindStatusCondition(r.Status.Conditions, BindingDeploymentStatusConditionType)
+			Expect(*statusCondition).To(MatchFields(IgnoreExtras, Fields{
+				"Status":  Equal(metav1.ConditionTrue),
+				"Reason":  Equal("CommitsSynced"),
+				"Message": Equal("3 of 3 components deployed"),
+			}))
+		})
```
davidmogar, line 150:
```suggestion
		})

```

davidmogar (approved):

PR #138, opened 2022-12-03
feat(HACBS-1438): move get functions to a separated file

davidmogar:
The Release adapter is becoming more difficult to maintain. This change moves all get functions to a separate file to leave in the adapter only relevant stuff. In addition, with this change multiple controllers can use the same functions.


api/v1alpha1/release_types.go
```diff
@@ -54,9 +54,6 @@ const (
 	// ReleaseReasonPipelineFailed is the reason set when the release PipelineRun failed
 	ReleaseReasonPipelineFailed ReleaseReason = "ReleasePipelineFailed"
 
-	// ReleaseReasonReleasePlanValidationError is the reason set when there is a validation error with the ReleasePlan
```
johnbieren, line 57: We already agreed long ago as part of HACBS-980 that there would be a separate error reported to the user for this. Why are you changing it as a PR to just move things around?

api/v1alpha1/release_types_test.go
```diff
@@ -222,7 +222,7 @@ var _ = Describe("Release type", func() {
 			r.Status.Conditions[0] = metav1.Condition{
 				Type:    "Fail",
 				Status:  metav1.ConditionUnknown,
-				Reason:  ReleaseReasonReleasePlanValidationError.String(),
+				Reason:  ReleaseReasonValidationError.String(),
```
johnbieren, line 225: I do not agree with this change

loader/loader.go
```diff
@@ -0,0 +1,281 @@
...
+
+// convertUnstructuredObject converts an unstructured object into an instance of the crd passed as an argument.
+func convertUnstructuredObject[T any](unstructuredObject *unstructured.Unstructured, crd *T) error {
+	return runtime.DefaultUnstructuredConverter.FromUnstructured(unstructuredObject.UnstructuredContent(), crd)
+}
+
+// getObject loads an object from the cluster. This is a generic function that requires the object to be passed as an
+// argument. The object is modified during the invocation, but it's returned as well to simplify the implementation of
+// functions using this function.
+func getObject[T any](name, namespace string, cli client.Client, ctx context.Context, object *T) (*T, error) {
```
johnbieren, line 26: I think this makes more sense in the goodies repo

loader/loader.go
```diff
@@ -0,0 +1,281 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return object, convertUnstructuredObject(unstructuredObject, object)
+}
+
+// getUnstructuredObject returns an unstructured object containing the same data as the object passed as an argument.
+func getUnstructuredObject[T any](object T) (*unstructured.Unstructured, error) {
```
johnbieren, line 45: I think this makes more sense in the goodies repo

loader/loader.go
```diff
@@ -0,0 +1,281 @@
...
+
+	if !activeReleasePlanAdmissionFound {
+		return nil, fmt.Errorf("no ReleasePlanAdmission found in the target (%+v) for application '%s'",
+			releasePlan.Spec.Target, releasePlan.Spec.Application)
+	}
+
+	return &releasePlanAdmissions.Items[0], nil
+}
+
+// GetActiveReleasePlanAdmissionFromRelease returns the ReleasePlanAdmission targeted by the ReleasePlan referenced by
```
johnbieren, line 87: docstring was not updated for this function. I don't know why we want it

loader/loader.go
```diff
@@ -0,0 +1,281 @@
...
+	for _, binding := range bindingList.Items {
+		if binding.Spec.Application == releasePlanAdmission.Spec.Application {
+			return &binding, nil
+		}
+	}
+
+	return nil, nil
+}
+
+// Composite functions
```
johnbieren, line 196: I don't think we should add these composite functions. Just in my last PR I had to split apart an existing composite function (getting the binding resources). I don't think they help us and I think they are just going overboard with trying to simplify the adapter code. The extra level of indirection doesn't help much in my opinion... Having 3-4 get methods in a row in the adapter is very easy to follow

controllers/release/release_controller_test.go
```diff
@@ -119,11 +116,14 @@ var _ = Describe("Release Controller", func() {
 		}).Should(BeNil())
 	})
 
-	It("can Reconcile function prepare the adapter and return the result of the reconcile handling operation", func() {
-		result, err := releaseReconciler.Reconcile(ctx, req)
-		Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(reconcile.Result{})))
-		Expect(err).To(BeNil())
-	})
+	// With current resources this would return a different thing. This "error" was hidden by a "bad" implementation of
+	// one of the operations. To be fixed during the tests refactor.
```
johnbieren, line 120: I am confused. What is the future tests refactor? I don't think we should merge things that break the operator

johnbieren (changes requested):

controllers/release/release_controller_test.go
```diff
@@ -119,11 +116,14 @@ var _ = Describe("Release Controller", func() {
...
+	//
+	//It("can Reconcile function prepare the adapter and return the result of the reconcile handling operation", func() {
+	//	result, err := releaseReconciler.Reconcile(ctx, req)
+	//	Expect(reflect.TypeOf(result)).To(Equal(reflect.TypeOf(reconcile.Result{})))
+	//	Expect(err).To(BeNil())
+	//})
```
kasemAlem, line 126: if you choose to comment this test, should it stay ? why not to delete it ?

loader/loader.go
johnbieren: I don't think we need all the details about the label and list operator here. They are not things in this function, but in the functions called by this function

loader/loader.go
```diff
@@ -0,0 +1,250 @@
...
+	releasePlan, err := l.GetReleasePlan(ctx, cli, release)
+	if err != nil {
+		return nil, err
+	}
+
+	return l.GetActiveReleasePlanAdmission(ctx, cli, releasePlan)
+}
+
+// GetApplication returns the Application referenced by the ReleasePlanAdmission. If the Application is not found or
+// the Get operation failed, an error will be returned.
```
johnbieren, line 96: `s/failed/fails/`

loader/loader_test.go
```diff
@@ -0,0 +1,423 @@
...
+	Context("When calling GetActiveReleasePlanAdmission", func() {
+		It("returns an active release plan admission", func() {
+			returnedObject, err := loader.GetActiveReleasePlanAdmission(ctx, k8sClient, releasePlan)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(returnedObject).NotTo(Equal(&v1alpha1.ReleasePlanAdmission{}))
+			Expect(returnedObject.Name).To(Equal(releasePlanAdmission.Name))
+		})
+
+		It("fails to return an active release plan admission if the target does not match", func() {
+			modifiedReleasePlan := releasePlan.DeepCopy()
```
johnbieren, line 74: this should be deleted, similar to line 95, right?

loader/loader_test.go
```diff
@@ -0,0 +1,423 @@
...
+	Context("When calling GetActiveReleasePlanAdmissionFromRelease", func() {
+		It("returns an active release plan admission", func() {
+			returnedObject, err := loader.GetActiveReleasePlanAdmissionFromRelease(ctx, k8sClient, release)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(returnedObject).NotTo(Equal(&v1alpha1.ReleasePlanAdmission{}))
+			Expect(returnedObject.Name).To(Equal(releasePlanAdmission.Name))
+		})
+
+		It("fails to return an active release plan admission if the release plan does not match", func() {
+			modifiedRelease := release.DeepCopy()
```
johnbieren, line 108: this should be deleted like line 95, right?

loader/loader_test.go
```diff
@@ -0,0 +1,423 @@
...
+	Context("When calling GetReleasePipelineRun", func() {
+		It("returns a PipelineRun if the labels match with the release data", func() {
+			returnedObject, err := loader.GetReleasePipelineRun(ctx, k8sClient, release)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(returnedObject).NotTo(Equal(&v1beta1.PipelineRun{}))
+			Expect(returnedObject.Name).To(Equal(pipelineRun.Name))
+		})
+
+		It("fails to return a PipelineRun if the labels don't match with the release data", func() {
+			modifiedRelease := release.DeepCopy()
```
johnbieren, line 172: this should be deleted like line 95 right?

loader/loader_test.go
```diff
@@ -0,0 +1,423 @@
...
+	Context("When calling GetSnapshotEnvironmentBinding", func() {
+		It("returns a snapshot environment binding if the environment field value matches the release plan admission one", func() {
+			returnedObject, err := loader.GetSnapshotEnvironmentBinding(ctx, k8sClient, releasePlanAdmission)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(returnedObject).NotTo(Equal(&applicationapiv1alpha1.SnapshotEnvironmentBinding{}))
+			Expect(returnedObject.Name).To(Equal(snapshotEnvironmentBinding.Name))
+		})
+
+		It("fails to return a snapshot environment binding if the environment field value doesn't match the release plan admission one", func() {
+			modifiedReleasePlanAdmission := releasePlanAdmission.DeepCopy()
```
johnbieren, line 217: this should be deleted like line 95 right?

loader/loader_test.go
```diff
@@ -0,0 +1,423 @@
...
+
+	Context("When calling GetSnapshotEnvironmentBindingFromReleaseStatus", func() {
+		It("fails to return a snapshot environment binding if the reference is not in the release", func() {
+			returnedObject, err := loader.GetSnapshotEnvironmentBindingFromReleaseStatus(ctx, k8sClient, release)
+			Expect(returnedObject).To(BeNil())
+			Expect(err.Error()).To(ContainSubstring("release doesn't contain a valid reference to an SnapshotEnvironmentBinding"))
+		})
+
+		It("fails to return a snapshot environment binding if the environment field value doesn't match the release plan admission one", func() {
+			modifiedRelease := release.DeepCopy()
```
johnbieren, line 234: missing deletion?

loader/loader_test.go
```diff
@@ -0,0 +1,423 @@
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(*resources).To(MatchFields(IgnoreExtras, Fields{
+				"Application":           Not(BeNil()),
+				"ApplicationComponents": Not(BeNil()),
+				"Snapshot":              Not(BeNil()),
+			}))
+		})
+
+		It("fails if any resource fails to be fetched", func() {
+			modifiedReleasePlanAdmission := releasePlanAdmission.DeepCopy()
```
johnbieren, line 257: missing deletion?

controllers/release/release_controller_test.go
johnbieren: I also agree with deleting it over commenting it out. Better would be to fix it as part of this PR, but

controllers/release/release_adapter.go
```diff
@@ -47,24 +48,26 @@ import (
...
-func NewAdapter(release *v1alpha1.Release, logger logr.Logger, client client.Client, context context.Context) *Adapter {
+func NewAdapter(ctx context.Context, client client.Client, release *v1alpha1.Release, loader loader.ObjectLoader, logger logr.Logger) *Adapter {
 	return &Adapter{
-		release: release,
-		logger:  logger,
 		client:  client,
-		context: context,
-		syncer:  syncer.NewSyncerWithContext(client, logger, context),
+		ctx:     ctx,
+		loader:  loader,
```
johnbieren, line 67: Why again do we need to do `a.loader.func` and have it be part of the adapter instead of `loader.func`?

johnbieren (commented):
I think the loader stuff could really use a README, especially considering all the mock stuff is in there. It'd save you from having to explain it 10 times. I also think the loader stuff should be in the goodies, so the README could be there, but I've mentioned that and sounds like it may be done later if at all

loader/loader.go
davidmogar: To be done once the loader is established in the release-service and in the internal-services repos.

loader/loader.go
davidmogar: I think we should keep it. From the caller point of view, they don't have any idea what we do inside this function. We would be hiding the behavior.

loader/loader_test.go
davidmogar: No. In this case I'm just passing an struct. In the test for line 95 I'm creating a resource in the cluster.

loader/loader_test.go
davidmogar: Nope. No object is created here.

loader/loader_test.go
davidmogar: Nope. No object is created here.

loader/loader_test.go
davidmogar: Yes! Just kidding. Like before, no object is created here, so no need to delete anything.

loader/loader_test.go
davidmogar: Nope.

loader/loader_test.go
davidmogar: Guess what!!! Nope.

controllers/release/release_adapter.go
davidmogar: Because that way we can create an adapter which operations use a mock.

loader/loader.go
johnbieren: I suppose..

loader/loader.go
```diff
@@ -0,0 +1,265 @@
...
+		}
+		activeReleasePlanAdmission = &releasePlanAdmissions.Items[i]
+	}
+
+	if activeReleasePlanAdmission == nil {
+		return nil, fmt.Errorf("no ReleasePlanAdmission found in the target (%+v) for application '%s'",
+			releasePlan.Spec.Target, releasePlan.Spec.Application)
+	}
+
+	return &releasePlanAdmissions.Items[0], nil
```
johnbieren, line 87: this is wrong, should be activeReleasePlanAdmission

loader/loader_test.go
johnbieren: This is what I get for being thorough :roll_eyes:

loader/loader_test.go
johnbieren: ha ha

controllers/release/release_adapter.go
```diff
@@ -119,70 +122,71 @@ func (a *Adapter) EnsureFinalizerIsAdded() (reconciler.OperationResult, error) {
 // EnsureReleasePlanAdmissionEnabled is an operation that will ensure that the ReleasePlanAdmission is enabled.
 // If it is not, no further operations will occur for this Release.
 func (a *Adapter) EnsureReleasePlanAdmissionEnabled() (reconciler.OperationResult, error) {
-	_, err := a.getActiveReleasePlanAdmission()
+	_, err := a.loader.GetActiveReleasePlanAdmissionFromRelease(a.ctx, a.client, a.release)
```
johnbieren, line 125: I don't think you should do this `FromRelease`.. it is the first operation, so it won't be there the first time we reconcile a release, right?

johnbieren (changes requested):

johnbieren (approved):

theflockers (commented):
I just don't like the fact that for every new Get function we need to pass the `context` and `adapter` again. Not exactly against, but the adapter looked simpler to get the resources. Any reason for that? Is it possible to simplify it again?

EDIT: Also I would tidy up all imports. Easier say it here then in all places.

controllers/release/release_adapter.go
```diff
@@ -19,6 +19,7 @@ package release
 import (
 	"context"
 	"fmt"
+	"github.com/redhat-appstudio/release-service/loader"
```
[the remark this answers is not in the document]
davidmogar: > I just don't like the fact that for every new Get function we need to pass the `context` and `adapter` again. Not exactly against, but the adapter looked simpler to get the resources. Any reason for that? Is it possible to simplify it again?
> 
> EDIT: Also I would tidy up all imports. Easier say it here than in all places.

We don't pass the adapter but the context and the client. Two use cases:

- KCP required two different contexts. If we move back to that, the loader will need to know which context to use.
- The internal services operator uses two different clients. One for the internal cluster and the other for the public cluster. The loader needs to know which one to use.

theflockers (approved):
LGTM.

PR #145, opened 2022-12-21
chore: Format setupControllerWithManager

gbenhaim:
Break long lines.

Signed-off-by: gbenhaim <gbenhaim@redhat.com>

### DIFF
--- a/controllers/release/release_controller.go
+++ b/controllers/release/release_controller.go
@@ -119,18 +119,34 @@ func setupControllerWithManager(manager ctrl.Manager, reconciler *Reconciler) er
 	}
 
 	return ctrl.NewControllerManagedBy(manager).
-		For(&v1alpha1.Release{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
-		Watches(&source.Kind{Type: &applicationapiv1alpha1.SnapshotEnvironmentBinding{}}, &libhandler.EnqueueRequestForAnnotation{
-			Type: schema.GroupKind{
-				Kind:  "Release",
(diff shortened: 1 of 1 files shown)


gbenhaim:
I think that `[It] ensures finalizers are called when a release is deleted` is a bit racy. It failed sporadically when I ran the unit test locally.

davidmogar:
> I think that `[It] ensures finalizers are called when a release is deleted` is a bit racy. It failed sporadically when I ran the unit test locally.

The actual problem is with the tests themselves and more specifically with the use of a cache. We are aware of it and I have a rough idea of the changes needed in the code and in the tests to fix this. It will take some time though.

Regarding this PR I’ll take a look next year 😉

johnbieren:
> I just properly reviewed . Actually, I prefer the current style with the exception of the `Watch`, so I would change it to:
> 
> ```
>                 For(&v1alpha1.Release{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
> 		Watches(&source.Kind{Type: &v1beta1.PipelineRun{}},
> 			&libhandler.EnqueueRequestForAnnotation{
> 				Type: schema.GroupKind{
> 					Kind:  "Release",
> 					Group: "appstudio.redhat.com",
> 				},
> 			}, builder.WithPredicates(tekton.ReleasePipelineRunSucceededPredicate()),
> 		).
> 		Complete(reconciler)
> ```
> 
> I don't think the For line needs any change as it's short enough. @johnbieren take a look and comment, please.

I don't have a strong opinion on this one. I am good with whatever you two agree on

PR #146, opened 2023-01-01
feat (HACBS-1306) : Confirm HTTPS is used to communicate on the network and block all egress

naftalysh:
The network policy allows ingress traffic on TCP ports 443, 444, 8080, 8443, 4443, and 9443 (HTTPS) if the TLS termination is set to "passthrough" And blocks all egress traffic for pods with the label "control-plane: controller-manager" In the system namespace.

Signed-Off-By: Naftaly Shprai <naftalysh@gmail.com>

### DIFF
--- a/config/manager/network_policy.yaml
+++ b/config/manager/network_policy.yaml
@@ -2,7 +2,7 @@
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: allow-ingress-https-only-and-block-egress
+  name: allow-https-only-for-specific-ports-ingress
   namespace: system
 spec:
   podSelector:
@@ -12,10 +12,19 @@ spec:
   - Ingress
   - Egress
   ingress:
-  - ports:
-    - protocol: TCP
-      port: 443,444,8080,8443,4443,9443
-      tls:
-        termination: passthrough
+  - matchExpressions:
+    - key: port
+      operator: In
+      values:
+      - "443"
+      - "444"
+      - "4443"
+      - "8080"
+      - "8081"
+      - "8443"
+      - "9443"
+    - key: scheme
+      operator: In
+      values:
+      - https
   egress: []
----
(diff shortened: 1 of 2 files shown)


naftalysh:
//CC @davidmogar

config/manager/network_policy.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - Ingress
+  - Egress
+  ingress:
+  - ports:
+    - protocol: TCP
+      port: 443,444,8080,8443,4443,9443
+      tls:
+        termination: passthrough
+  egress: []
+---
```
davidmogar, line 21:
```suggestion
```
Remove this line and make sure you have a newline at the end (notice the red symbol).

config/manager/network_policy.yaml
```diff
@@ -0,0 +1,21 @@
...
+  podSelector:
+    matchLabels:
+      control-plane: controller-manager
+  policyTypes:
+  - Ingress
+  - Egress
+  ingress:
+  - ports:
+    - protocol: TCP
+      port: 443,444,8080,8443,4443,9443
```
davidmogar, line 17: Sort them, please. Also, this is missing the health probe port. Is this on purpose? Regarding the rest of the ports? Why? Do we need them?

davidmogar:
Title for this PR is wrong. This change allow incoming traffic only on a given set of ports which happen to include HTTPS ports that we probably don't even need. Could you please reflect this on the PR title and the commit (after squashing)?

config/manager/network_policy.yaml
```diff
@@ -0,0 +1,21 @@
+---
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: allow-ingress-https-only-and-block-egress
```
johnbieren, line 5: The name concerns me. The other PR adds more things to the network policy... I don't think we can have every single thing the policy does in the name. IMO, we should rely on inspecting the policy to see what it does, and we can use the commit history to see why certain aspects were added. Maybe it should be something like `controller-manager` or something so we can easily see it is for the manager pod

PR #147, opened 2023-01-11
refactor(HACBS-1463): use ubi go 1.18 image as base image

mmalina:
We had to move to Docker's Go image as there was
no official Red Hat 1.18 image. Now that it exists, let's switch to it.

Also, switch to ubi-micro for runtime base image.

Signed-off-by: Martin Malina <mmalina@redhat.com>

This PR is inspired by https://github.com/redhat-appstudio/integration-service/pull/78

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,5 @@
 # Build the manager binary
-FROM golang:1.18 as builder
-
-WORKDIR /workspace
+FROM registry.access.redhat.com/ubi9/go-toolset:1.18 as builder
 
 # Copy the Go Modules manifests
 COPY go.mod go.mod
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -26,11 +24,10 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go
 ARG ENABLE_WEBHOOKS=true
 ENV ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS}
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
-WORKDIR /
-COPY --from=builder /workspace/manager .
+# Use ubi-minimal as minimal base image to package the manager binary
```
davidmogar, line 27: Maybe you meant micro here.
```suggestion
# Use ubi-micro as minimal base image to package the manager binary
```

davidmogar:
> 1. Do you know if there is a reason why they used ubi8 and not ubi9 in [feat(HACBS-1215): use the Red Hat go 1.18 image in the Integration service integration-service#78](https://github.com/redhat-appstudio/integration-service/pull/78) ? I would have thought to go with ubi9 by default.
No reason.

> 2. Do we care about standardizing across the services and so we should also use ubi8?
Not really.

> 3. ubi-micro instead of ubi-minimal should be fine, right? It's definitely not more stripped down than the distroless image that was used before. But same as with the builder image - do we care about standardizing across services?
If micro is good enough, we should go for micro. Smaller -> better.

> 4. And finally, do we prefer to stick to a version for the runtime base image? Given that it's provided by Red Hat, maybe we could use `:latest`?
I'd say that we should avoid using `latest` everywhere. Easy way to get into problems in the future with unrelated changes. Let's tie this version to bumps in the Go version and keep both up to date. Also, using latest is not that great for reproducibility.

Dockerfile
mmalina: @davidmogar yep, fixed, thanks.

davidmogar (approved):

PR #149, opened 2023-01-11
fix: don't error if binding status missing

johnbieren:
This fixes a bug where the controller would error out if it tried to run the EnsureSnapshotEnvironmentBindingIsTracked operation and the binding CR is missing the AllComponentsDeployed Status condition.

### DIFF
--- a/controllers/release/release_adapter.go
+++ b/controllers/release/release_adapter.go
@@ -578,21 +578,25 @@ func (a *Adapter) getSnapshotEnvironmentResources(releasePlanAdmission *v1alpha1
 // registerGitOpsDeploymentStatus updates the status of the Release being processed by monitoring the status of the
 // associated SnapshotEnvironmentBinding and setting the appropriate state in the Release.
 func (a *Adapter) registerGitOpsDeploymentStatus(binding *applicationapiv1alpha1.SnapshotEnvironmentBinding) error {
-	if binding != nil {
-		patch := client.MergeFrom(a.release.DeepCopy())
-
-		condition := meta.FindStatusCondition(binding.Status.ComponentDeploymentConditions,
-			applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed)
-		if condition.Status == metav1.ConditionUnknown {
-			a.release.MarkDeploying(condition.Reason, condition.Message)
-		} else {
-			a.release.MarkDeployed(condition.Status, condition.Reason, condition.Message)
-		}
-
-		return a.client.Status().Patch(a.context, a.release, patch)
+	if binding == nil {
+		return nil
 	}
 
-	return nil
+	condition := meta.FindStatusCondition(binding.Status.ComponentDeploymentConditions,
+		applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed)
+	if condition == nil {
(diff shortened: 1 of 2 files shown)


controllers/release/release_adapter.go
```diff
@@ -582,13 +582,15 @@ func (a *Adapter) registerGitOpsDeploymentStatus(binding *applicationapiv1alpha1
...
+		if condition != nil {
+			if condition.Status == metav1.ConditionUnknown {
+				a.release.MarkDeploying(condition.Reason, condition.Message)
+			} else {
+				a.release.MarkDeployed(condition.Status, condition.Reason, condition.Message)
+			}
 
-		return a.client.Status().Patch(a.context, a.release, patch)
+			return a.client.Status().Patch(a.context, a.release, patch)
+		}
```
davidmogar, line 593: Too many levels. Let's take the chance to change the function to something like:
```
if binding == nil
    return nil

condition := ...

if condition == nil
    return nil
    
if condition.Status ... {
}

return a.client.Status...
```

controllers/release/release_adapter.go
```diff
@@ -578,21 +578,22 @@ func (a *Adapter) getSnapshotEnvironmentResources(releasePlanAdmission *v1alpha1
...
 	}
 
-	return nil
+	patch := client.MergeFrom(a.release.DeepCopy())
+	if condition.Status == metav1.ConditionUnknown {
+		a.release.MarkDeploying(condition.Reason, condition.Message)
+	} else {
+		a.release.MarkDeployed(condition.Status, condition.Reason, condition.Message)
+	}
+	return a.client.Status().Patch(a.context, a.release, patch)
```
davidmogar, line 596: Do you mind to add an empty line before this one?

controllers/release/release_adapter.go
```diff
@@ -578,21 +578,22 @@ func (a *Adapter) getSnapshotEnvironmentResources(releasePlanAdmission *v1alpha1
...
+	}
+	condition := meta.FindStatusCondition(binding.Status.ComponentDeploymentConditions,
+		applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed)
+	if condition == nil {
+		return nil
 	}
 
-	return nil
+	patch := client.MergeFrom(a.release.DeepCopy())
+	if condition.Status == metav1.ConditionUnknown {
```
davidmogar, line 591: Do you mind to add an empty line before this one?

controllers/release/release_adapter.go
```diff
@@ -578,21 +578,22 @@ func (a *Adapter) getSnapshotEnvironmentResources(releasePlanAdmission *v1alpha1
...
-			a.release.MarkDeploying(condition.Reason, condition.Message)
-		} else {
-			a.release.MarkDeployed(condition.Status, condition.Reason, condition.Message)
-		}
-
-		return a.client.Status().Patch(a.context, a.release, patch)
+	if binding == nil {
+		return nil
+	}
+	condition := meta.FindStatusCondition(binding.Status.ComponentDeploymentConditions,
```
davidmogar, line 584: Do you mind to add an empty line before this one?

davidmogar (approved):

PR #153, opened 2023-01-20
feat(HACBS-1645): also watch binding status going from nil to true/false

johnbieren:
The release controller should reconcile for SnapshotEnvironmentBindings that it owns and has the AllComponentsDeployed status condition go from nonexisting to true/false, similarly to how it reconciles when the condition goes from unknown to true/false. This commit adds that functionality.

### DIFF
--- a/gitops/utils.go
+++ b/gitops/utils.go
@@ -27,22 +27,23 @@ import (
 // is a SnapshotEnvironmentBinding with the componentDeployment status Unknown and the second
 // passed object is a SnapshotEnvironmentBinding with the componentDeployment status True/False.
 func hasDeploymentFinished(objectOld, objectNew client.Object) bool {
+	var ok bool
+	var oldBinding, newBinding *applicationapiv1alpha1.SnapshotEnvironmentBinding
 	var oldCondition, newCondition *metav1.Condition
 
-	if oldBinding, ok := objectOld.(*applicationapiv1alpha1.SnapshotEnvironmentBinding); ok {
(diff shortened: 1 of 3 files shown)


gitops/utils.go
```diff
@@ -32,17 +32,14 @@ func hasDeploymentFinished(objectOld, objectNew client.Object) bool {
...
 		newCondition = meta.FindStatusCondition(newBinding.Status.ComponentDeploymentConditions,
 			applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed)
-		if newCondition == nil {
+		if newCondition == nil || newCondition.Status == metav1.ConditionUnknown {
 			return false
 		}
 	}
 
-	return oldCondition.Status == metav1.ConditionUnknown && newCondition.Status != metav1.ConditionUnknown
+	return oldCondition == nil || oldCondition.Status == metav1.ConditionUnknown
```
davidmogar, line 44: You can reach this part of the code with `objectNew` not being a Binding. Not that is going to happen, but we should fix that.

davidmogar: I'd change the code to something like this:
```
if oldObject not a binding -> return
if newObject not a binding -> return

get condition for oldBinding
get condition for newBinding

return (oldCondition == nil || oldCondition.Status == Unknown) && (newCondition != nil && newCondition.Status != Unkown)
```

davidmogar (changes requested):

gitops/utils.go
johnbieren: Fixed and added a test case for this

happybhati (approved):

davidmogar (approved):

PR #157, opened 2023-01-24
feat(HACBS-1668): Release service failing to the container after infra-deployment on aws cluster

naftalysh:
Signed-Off-By: Naftaly Shprai <naftalysh@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,9 @@ ENV ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS}
 
 # Use ubi-micro as minimal base image to package the manager binary
 # See https://catalog.redhat.com/software/containers/ubi9/ubi-micro/615bdf943f6014fa45ae1b58
-FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# for debugging purposes, I use my made image in the FROM clause with client utils
+FROM quay.io/nshprai/fedora-35:latest
 COPY --from=builder /opt/app-root/src/manager /
 USER 65532:65532
 
--- a/config/manager/network_policy.yaml
+++ b/config/manager/network_policy.yaml
@@ -2,15 +2,10 @@
 apiVersion: networking.k8s.io/v1
 kind: NetworkPolicy
 metadata:
-  name: prevent-all-egress
-  namespace: system
+  name: allow-all-egress
 spec:
-  podSelector:
-    matchLabels:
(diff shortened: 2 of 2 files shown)


Dockerfile
```diff
@@ -26,7 +26,9 @@ ENV ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS}
 
 # Use ubi-micro as minimal base image to package the manager binary
 # See https://catalog.redhat.com/software/containers/ubi9/ubi-micro/615bdf943f6014fa45ae1b58
-FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# for debugging purposes, I use my made image in the FROM clause with client utils
+FROM quay.io/nshprai/fedora-35:latest
```
davidmogar, line 31: Is this a PR for testing? If so, delete the PR please.

config/manager/network_policy.yaml
```diff
@@ -2,11 +2,28 @@
...
 kind: NetworkPolicy
 metadata:
-  name: prevent-all-egress
+  name: prevent-all-egress-allow-api-server
   namespace: system
 spec:
   podSelector:
     matchLabels:
-      control-plane: controller-manager
+      control-plane: controller-man
```
mmalina, line 10: Is this intentional? It looks like a mistake.

config/manager/network_policy.yaml
```diff
@@ -2,11 +2,28 @@
...
   policyTypes:
     - Egress
+  egress:
+    - to:
+        - ipBlock:
+            cidr: 172.30.0.0/24
```
mmalina, line 16: kubeapi ip can be different on different clusters. I'm not sure what the best approach is. But it doesn't seem great to hardcode it here. Or maybe we can have something here, but then patch it using kustomize in the infra repo. And then also customize it using preview.sh when preparing a testing cluster. @davidmogar wdyt?

davidmogar: This was just a simple workaround that I asked @naftalysh to test. What I don't get is why we have a PR in here for this. Feature branches and forks are for that. I'm closing it.

config/manager/network_policy.yaml
naftalysh: This is a mistake , I by mistake deleted the label value, it should be controller-manager

config/manager/network_policy.yaml
naftalysh: I can change it to a generic solution

PR #159, opened 2023-01-26
fix: return ReleasePlanAdmission with matching application

johnbieren:
If there are two ReleasePlanAdmissions in the managed namespace with the same origin, but different applications, the one with the matching application should be returned, not just index 0 of the list of ReleasePlanAdmissions found in the managed namespace that have a matching origin.


davidmogar:
You need to fix the gosec errors. Apart from this, the new logic is right.

davidmogar (approved):

PR #164, opened 2023-01-31
Hacbs 1308/release service block all egress network policy new

naftalysh:
JIRA story - https://issues.redhat.com/browse/HACBS-1308
Put a network policy around release-service controller, prevent egress to everywhere

This network policy was adapted also for OVN network provider type of OpenShift clusters
Block every egress besides important services - like apiserver, kubelet and the dns

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,9 @@ ENV ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS}
 
 # Use ubi-micro as minimal base image to package the manager binary
 # See https://catalog.redhat.com/software/containers/ubi9/ubi-micro/615bdf943f6014fa45ae1b58
-FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# for debugging purposes, I use my made image in the FROM clause with client utils
+FROM quay.io/nshprai/fedora-35:latest
 COPY --from=builder /opt/app-root/src/manager /
 USER 65532:65532
 
--- a/config/manager/network_policy.yaml
+++ b/config/manager/network_policy.yaml
@@ -12,6 +12,9 @@ spec:
     - Egress
   egress:
     - to:
+      - ipBlock:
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -26,7 +26,9 @@ ENV ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS}
 
 # Use ubi-micro as minimal base image to package the manager binary
 # See https://catalog.redhat.com/software/containers/ubi9/ubi-micro/615bdf943f6014fa45ae1b58
-FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# for debugging purposes, I use my made image in the FROM clause with client utils
+FROM quay.io/nshprai/fedora-35:latest
```
davidmogar, line 31: If this PR is a draft, mark it as such please. If not, this change can't go.

config/manager/network_policy.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - Egress
+  egress:
+    - to:
+      - namespaceSelector:
+          matchLabels:
+            name: kube-system
+        podSelector:
+          matchLabels:
+            k8s-app: kube-apiserver-proxy
+      ports:
```
davidmogar, line 21: Are the ports needed if you already use the pod/namespace selectors?

kasemAlem: Also here I agree with David's  comment 
If using namespaces selector ports are required ?

Dockerfile
kasemAlem: I's using your repo of registry

kasemAlem (changes requested):

Dockerfile
naftalysh: I did so, thanks

config/manager/network_policy.yaml
naftalysh: I wanted to be precise, but maybe its not needed, I shall remove the ports

PR #168, opened 2023-02-03
feat(HACBS-1096): add a metric that details number of deployments

johnbieren:
This commit adds two metrics - one that measures the total number of deployments released to managed environments, and one that measures release durations from when the SnapshotEnvironmentBinding was created til the release is marked as deployed.

### DIFF
--- a/api/v1alpha1/release_types.go
+++ b/api/v1alpha1/release_types.go
@@ -83,6 +83,14 @@ type ReleaseStatus struct {
 	// +optional
 	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
 
+	// DeploymentStartTime is the time when the SnapshotEnvironmentBinding was created
+	// +optional
+	DeploymentStartTime *metav1.Time `json:"deploymentStartTime,omitempty"`
+
+	// DeploymentCompletionTime is the time when the SnapshotEnvironmentBinding has all components deployed
+	// +optional
+	DeploymentCompletionTime *metav1.Time `json:"deploymentCompletionTime,omitempty"`
+
 	// Conditions represent the latest available observations for the release
 	// +optional
 	Conditions []metav1.Condition `json:"conditions"`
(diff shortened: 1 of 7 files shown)


controllers/release/release_adapter.go
```diff
@@ -354,7 +354,8 @@ func (a *Adapter) registerGitOpsDeploymentStatus(binding *applicationapiv1alpha1
 	patch := client.MergeFrom(a.release.DeepCopy())
 
 	if condition.Status == metav1.ConditionTrue {
-		a.release.MarkDeployed(condition.Reason, condition.Message)
+		a.release.MarkDeployed(condition.Reason, condition.Message,
+			binding.CreationTimestamp, condition.LastTransitionTime)
```
davidmogar, line 358: Why not to do like we do in the release? `MarkDeploying` function to set the start time and `MarkDeployed` to set the completion time.

davidmogar: Sounds reasonable to me. We should track that information in the Release. We can restructure the status after this change so it's more readable though. Something like:
```
Conditions: []
Deployment:
    Binding: foo
    StartTime: ...
    CompletionTime: ...
    Target: bar
Release:
    PipelineRun: baz
    Strategy: qux
    StartTime: ...
    CompletionTime: ...
```

api/v1alpha1/release_types.go
```diff
@@ -158,19 +168,28 @@ func (r *Release) IsDone() bool {
 	return condition != nil && condition.Status != metav1.ConditionUnknown
 }
 
-// MarkDeployed sets the AllComponentsDeployed status in the Release to True with the provided reason and message.
-func (r *Release) MarkDeployed(reason, message string) {
+// MarkDeployed registers the deployment completion time and sets the AllComponentsDeployed status in the
+// Release to True with the provided reason and message.
+func (r *Release) MarkDeployed(reason, message string, deployedTime metav1.Time) {
```
davidmogar, line 173: Don’t pass the time. Same as the existing functions for the release. The moment you call it, the time is generated internally. We want the completion time to our eyes, not the real one.

metrics/release.go
```diff
@@ -32,6 +32,22 @@ var (
 		},
 	)
 
+	ReleaseAttemptDeploymentSeconds = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "release_attempt_deployment_seconds",
+			Help:    "Release durations from the moment the SnapshotEnvironmentBinding was created til the release is marked as deployed",
+			Buckets: []float64{0.5, 1, 2, 3, 4, 5, 6, 7, 10, 15, 30},
```
davidmogar, line 39: You sure about this buckets? They seem quite small tbh.

davidmogar (changes requested):

metrics/release.go
johnbieren: No I have no idea what average deployment times are

davidmogar: They seem unrealistic to be honest. @scoheb do you have some numbers for us?

api/v1alpha1/release_types.go
johnbieren: But then if the operation is called again, we will be updating the deployed time to right now when there could be no change at all in the deployment.. something else could have caused the release to be reconciled again

davidmogar: Same in the release. We have guards for that.

johnbieren: fixed

davidmogar (approved):
I’m approving this PR but the buckets for the histogram need to be reviewed.

PR #169, opened 2023-02-03
fix(HACBS-1720): release not updated when binding goes false to true

johnbieren:
The release status should be updated when the AllComponentsDeployed status condition on the SnapshotEnvironmentBinding goes from False to True. This fixes that functionality.

### DIFF
--- a/api/v1alpha1/release_types.go
+++ b/api/v1alpha1/release_types.go
@@ -138,16 +138,18 @@ func (r *Release) HasSucceeded() bool {
 	return meta.IsStatusConditionTrue(r.Status.Conditions, releaseConditionType)
 }
 
-// IsDeployed checks whether the Release has been deployed via GitOps.
+// IsDeployed checks whether the Release has been successfully deployed via GitOps.
 func (r *Release) IsDeployed() bool {
 	condition := meta.FindStatusCondition(r.Status.Conditions, applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed)
-	return condition != nil && condition.Status != metav1.ConditionUnknown
+	return condition != nil && condition.Status == metav1.ConditionTrue
 }
 
 // IsDeploying returns a boolean indicating whether the Release's status indicates that it is being deployed.
 func (r *Release) IsDeploying() bool {
 	return meta.IsStatusConditionPresentAndEqual(r.Status.Conditions,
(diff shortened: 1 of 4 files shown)


api/v1alpha1/release_types.go
```diff
@@ -156,20 +158,20 @@ func (r *Release) IsDone() bool {
...
-	if status != metav1.ConditionUnknown {
+// MarkDeployed sets the AllComponentsDeployed status in the Release to True with the provided reason and message.
+func (r *Release) MarkDeployed(reason, message string) {
+	r.setStatusConditionWithMessage(applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed,
+		metav1.ConditionTrue, ReleaseReason(reason), message)
+}
+
+// MarkDeploying sets the AllComponentsDeployed status in the Release to Unknown or False with the provided reason and message.
+func (r *Release) MarkDeploying(status metav1.ConditionStatus, reason, message string) {
+	if status != metav1.ConditionTrue {
```
davidmogar, line 169: I don't think we should pass the status here. We should set it to False as that's the current behaviour from GitOps. If they fix it, this will be changed to Unknown.

davidmogar (changes requested):

api/v1alpha1/release_types.go
johnbieren: Why change the code to not support one of the condition types when as it is written, all three are supported? Who is going to watch gitops github repo to check for if/when they make the change?

davidmogar: Because `False` should be a final state. I think this is something wrong in the GitOps side of the code. We could contribute a fix there.

davidmogar: We can leave it as it is. If so, could you at least add a comment saying in some nice and kind words that the condition set by GitOps should provide an Unknown status and, as they don't, we have to consider False as a transient status? Or something properly written.

davidmogar: Just for us to remember the reason and that we could/should provide a proper fix in the GitOps service.

johnbieren: let me know (or just resolve) if what I wrote captures what you'd like written

davidmogar (approved):

PR #171, opened 2023-02-13
feat(HACBS-1780): add bundle resolver

happybhati:
Implementation of tekton bundle resolver for the release-service pipelineRef.bundle and taskRef.bundle are being deprecated.

Signed-off-by: Happy <hbhati@redhat.com>


tekton/pipeline_run_test.go
```diff
@@ -210,8 +210,15 @@ var _ = Describe("PipelineRun", func() {
 
 		It("can add the ReleaseStrategy information to a PipelineRun object and ", func() {
```
davidmogar, line 211: Let's fix the title as part of this change. Also, we need a test for both scenarios:

- Pipeline is present and bundle is not.
- Both pipeline and bundle are present.

davidmogar (changes requested):

tekton/pipeline_run_test.go
happybhati: Thanks, Done! :)

tekton/pipeline_run_test.go
```diff
@@ -262,4 +269,58 @@ var _ = Describe("PipelineRun", func() {
...
 	})
+
+	Context("When calling getPipelineRef", func() {
+		It("should return a PipelineRef without resolver if the releaseStrategy contains no bundle", func() {
+			releaseStrategy := &v1alpha1.ReleaseStrategy{
+				Spec: v1alpha1.ReleaseStrategySpec{
+					Pipeline:              "release-pipeline",
+					Bundle:                "",
+					Policy:                "testpolicy",
+					PersistentVolumeClaim: persistentVolumeClaim,
```
johnbieren, line 280: I don't think this is needed as it is an optional field and doesn't matter in this function

tekton/pipeline_run_test.go
```diff
@@ -262,4 +269,58 @@ var _ = Describe("PipelineRun", func() {
...
+					ServiceAccount:        serviceAccountName,
```
johnbieren, line 281: I don't think this is needed as it is an optional field and doesn't matter in this function

tekton/pipeline_run_test.go
```diff
@@ -262,4 +269,58 @@ var _ = Describe("PipelineRun", func() {
 			})
 		})
 	})
+
+	Context("When calling getPipelineRef", func() {
+		It("should return a PipelineRef without resolver if the releaseStrategy contains no bundle", func() {
+			releaseStrategy := &v1alpha1.ReleaseStrategy{
+				Spec: v1alpha1.ReleaseStrategySpec{
+					Pipeline:              "release-pipeline",
+					Bundle:                "",
```
johnbieren, line 278: `Bundle` is optional, I think omitting the field is the same as putting the empty string (and is closer to what a user would actually do so we are testing like a user)

tekton/pipeline_run_test.go
```diff
@@ -262,4 +269,58 @@ var _ = Describe("PipelineRun", func() {
...
+					Policy:                "testpolicy",
+					PersistentVolumeClaim: persistentVolumeClaim,
+					ServiceAccount:        serviceAccountName,
+					Params: []v1alpha1.Params{
```
johnbieren, line 282: I don't think this is needed as it is an optional field and doesn't matter in this function

tekton/pipeline_run_test.go
```diff
@@ -262,4 +269,58 @@ var _ = Describe("PipelineRun", func() {
 			})
 		})
 	})
+
+	Context("When calling getPipelineRef", func() {
+		It("should return a PipelineRef without resolver if the releaseStrategy contains no bundle", func() {
```
johnbieren, line 274: can we say "does not contain a bundle" instead?

tekton/pipeline_run_test.go
happybhati: Ultimately makes sense. Thanks for highlighting, @johnbieren  Fixed.

tekton/pipeline_run_test.go
happybhati: Thanks for highlighting, @johnbieren  Fixed.

tekton/pipeline_run_test.go
```diff
@@ -262,4 +269,48 @@ var _ = Describe("PipelineRun", func() {
 			})
 		})
 	})
+
+	Context("When calling getPipelineRef", func() {
+		It("should return a PipelineRef without resolver if the releaseStrategy does not contain a bundle", func() {
+			releaseStrategy := &v1alpha1.ReleaseStrategy{
+				Spec: v1alpha1.ReleaseStrategySpec{
+					Pipeline: "release-pipeline",
```
johnbieren, line 277: policy is required. One too many deletes :slightly_smiling_face:

johnbieren: Not sure how it didn't raise a failure tbh

happybhati: Already! :)

johnbieren (approved):

tekton/pipeline_run_test.go
happybhati: @davidmogar, as we verified the test offline, you were Happy. I am resolving this comment.

PR #172, opened 2023-02-17
Testing CI workflow

mmalina:


mmalina:
It seems it's working - e2e tests were triggered. What I don't like is the APPROVALNOTIFIER message. I'm trying to get rid of it here: https://github.com/openshift/release/pull/36507

PR #176, opened 2023-02-22
chore: STONEBLD-777 - tekton-ci: update bundle

Michkov:
Switching to new bundle since development of old one was stopped before attempt to migrate to kcp. New bundle is removing dependency on shared secret.

### DIFF
--- a/.tekton/pull-request.yaml
+++ b/.tekton/pull-request.yaml
@@ -17,7 +17,7 @@ spec:
       value: 'quay.io/redhat-appstudio/pull-request-builds:release-service-{{ revision }}'
   pipelineRef:
     name: docker-build
-    bundle: quay.io/redhat-appstudio/hacbs-core-service-templates-bundle:latest
+    bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-core-services-docker-build:latest
   workspaces:
     - name: workspace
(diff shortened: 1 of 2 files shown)


mmalina:
@Michkov your commit message does not conform to the gitlint rules. I would suggest: `chore(STONEBLD-777): tekton-ci: update bundle`

.tekton/pull-request.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       value: 'quay.io/redhat-appstudio/pull-request-builds:release-service-{{ revision }}'
   pipelineRef:
     name: docker-build
-    bundle: quay.io/redhat-appstudio/hacbs-core-service-templates-bundle:latest
+    bundle: quay.io/redhat-appstudio-tekton-catalog/pipeline-core-services-docker-build:latest
```
mmalina, line 20: What do you think of tying this to a specific version? Not saying it's needed, just something to consider. (Same thing in the push pipeline.)

Michkov: the idea here was to be customer one of the pipelines, so rather fail for us than for customer.

mmalina: Alright, fair point.

mmalina (approved):
LGTM

PR #178, opened 2023-03-02
feat(HACBS-1308): add an egress network policy

happybhati:
This commit adds an Egress Network Policy
to block all outgoing(Egress) traffic from release service
pod(controller-manager)

The `spec.podSelector` contains the name of a pod 
on which we want to apply network policy
`spec.egress` have Whitelist rules to allow the necessary
 traffic to certain pods and namespaces.

## Note
This change is tested on the hyper-shift child cluster(OVN) and
QuickLab SDN and OVN clusters.

Child Hypershift cluster requires adding an endpoint in  
`ipBlock:` `cidr`. 

To get `CIDR` perform `oc get ep -n default` and add
Kubernetes endpoints in `ipBlock`.  This will allow the pod to talk with
k8s API. 

***Reach out to the PR owner if you want to test on the Hypershift cluster.***

## How to test the change

* For testing this change you need to have an Openshift cluster
* Clone your fork of infra-deployment and login into your cluster
* setup `preview.env` follow[ this](https://redhat-appstudio.github.io/infra-deployments/docs/development/deployment.html#setting-preview-mode) for instructions
* In preview.env add this information to the Release service section ~line#44
for testing this PR.
File hack/preview.env should never be included in the commit.
```
## Release service
### Change of the image
export RELEASE_IMAGE_REPO=quay.io/hbhati/release
export RELEASE_IMAGE_TAG=0.37
export RELEASE_SERVICE_PR_OWNER=happybhati
export RELEASE_SERVICE_PR_SHA=8f7114a5dddcc1015309851e3a591c55a4967c30
```
* Run Bootstrap script `./hack/bootstrap-cluster.sh preview ` follow doc for [guide](https://redhat-appstudio.github.io/infra-deployments/docs/development/deployment.html#getting-started)
* once done you can go to Openshift UI and on the side panel click on Networking --> NetworkPolicy
 to verify if you can see `release-service-egress-controller-manager` in the network policy.
* check the controller pod and Argocd UI to verify everything is work normal.

### DIFF
--- a/config/manager/network_policy.yaml
+++ b/config/manager/network_policy.yaml
@@ -1,0 +1,26 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: egress-controller-manager
+spec:
+  podSelector:
+    matchLabels:
+      control-plane: controller-manager
+  policyTypes:
+  - Egress
+  egress:
+    - to:
+      - namespaceSelector: {}
+        podSelector:
+          matchLabels:
+            apiserver: "true"
+            app: openshift-kube-apiserver
+      ports:
+        - port: 6443
+          protocol: TCP
+    - to:
+      - ipBlock:
+          cidr: 172.20.0.1/32
+      ports:
+        - port: 6443
+          protocol: TCP
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,5 +1,6 @@
 resources:
 - manager.yaml
+- network_policy.yaml
 
 generatorOptions:
   disableNameSuffixHash: true


config/manager/network_policy.yaml
```diff
@@ -0,0 +1,26 @@
...
+  name: egress-controller-manager
+spec:
+  podSelector:
+    matchLabels:
+      control-plane: controller-manager
+  policyTypes:
+  - Egress
+  egress:
+    - to:
+        - namespaceSelector: {}
```
johnbieren, line 13: can you make the indenting similar to line 21-22?

davidmogar:
Question about:
```
To get CIDR perform oc get ep -n default and add
Kubernetes endpoints in ipBlock. This will allow the pod to talk with
k8s API.
```

Is this something that should be done by the Release deployment or during a bootstrap of the cluster?

happybhati:
> Question about:
> 
> ```
> To get CIDR perform oc get ep -n default and add
> Kubernetes endpoints in ipBlock. This will allow the pod to talk with
> k8s API.
> ```
> 
> Is this something that should be done by the Release deployment or during a bootstrap of the cluster?

For testing this on hypershift cluster following steps must be followed if the master node is not present.

- From the terminal login to your Openshift cluster and perform `oc get ep -n default` and add
Kubernetes endpoints in `ipBlock` [here](https://github.com/redhat-appstudio/release-service/blob/ba832b817f25477d6274b9f866d509a708d8a450/config/manager/network_policy.yaml#L23).

- Now commit the change and build-push the release image.
- Add change to  infra-deployment, the commit sha, and the newly built image name and tag to `/hack/preview.env`
-  Now you can run `./hack/preview.sh` 
- Refresh the ArgoCD dashboard to see the changes. 
- monitor the release-service pod to see if all went well or not.

Note
```
If you are going to test this on any Quicklab cluster it will work without changing anything.
For deploying on hypershift cluster endpoint need to be added to network policy ipBlock

davidmogar:
> * From the terminal login to your Openshift cluster and perform `oc get ep -n default` and add
>   Kubernetes endpoints in `ipBlock` [here](https://github.com/redhat-appstudio/release-service/blob/ba832b817f25477d6274b9f866d509a708d8a450/config/manager/network_policy.yaml#L23).
So my question is more related to stage. Do we need to do anything there? Also, in the case of Hypershift, is there a way to automate this? If not, we will be fighting ArgoCD as it will revert any change we do.

kasemAlem (commented):
for me looks fine, 
just I think it should have some tests that validate in this case the release-service functionality is not broken .

happybhati:
> > * From the terminal login to your Openshift cluster and perform `oc get ep -n default` and add
> >   Kubernetes endpoints in `ipBlock` [here](https://github.com/redhat-appstudio/release-service/blob/ba832b817f25477d6274b9f866d509a708d8a450/config/manager/network_policy.yaml#L23).
> >   So my question is more related to stage. Do we need to do anything there? Also, in the case of Hypershift, is there a way to automate this? If not, we will be fighting ArgoCD as it will revert any change we do.

@davidmogar 
- I believe this should work in stage, the only way to test is after merging this PR. 
- So side effect of using hypershift the pod selector cannot select the k8s api pods, because they are not in its "cluster"
the only way around it is to use `ipBlock` like we are doing. luckily I guess the default API address in hypershift is 172.20.0.1 according to hypershift [git repo](https://github.com/openshift/hypershift/blob/b5d40fab5e30d22e6a0f5f94f32b098493199dd5/api/v1beta1/hostedcluster_types.go#L559). and we are using the same as `CIDR` in our `ipBlock`.

davidmogar (approved):

PR #179, opened 2023-03-06
refactor(HACBS-1777): reorg Release status

davidmogar:
This commit is a refactor of the Release CRD to add proper conditions and to reorganize the status field itself showing information of each phase of a release.

New metrics, loader functions and operations have been added to improve the code and to address changes in the Release CRD.


tekton/pipeline_run_test.go
```diff
@@ -208,7 +208,7 @@ var _ = Describe("PipelineRun", func() {
 			Expect(unmarshaledSnapshotSpec.Application).To(Equal(applicationName))
 		})
 
-		It("can add the ReleaseStrategy information to a PipelineRun object and ", func() {
+		It("can add the Strategy information to a PipelineRun object and ", func() {
```
johnbieren, line 211: I liked this better as `ReleaseStrategy`. The line isn't too long and we don't have a concept of "strategy" only the release strategy

tekton/utils_test.go
```diff
@@ -91,9 +91,9 @@ var _ = Describe("Utils", func() {
 				AsPipelineRun())).To(Equal(true))
 		})
 
-		It("returns true when ReleasePipelineRun.Status is `Succeeded` or false otherwise", func() {
+		It("returns true when PipelineRun.Status is `Succeeded` or false otherwise", func() {
 			releasePipelineRun.AsPipelineRun().Status.InitializeConditions(clock.RealClock{})
-			// MarkRunning sets Status to Unknown
+			// MarkReleasing sets Status to Unknown
```
johnbieren, line 96: the comment change doesn't make sense to me. It is preceding `MarkRunning` not `MarkReleasing`

loader/loader.go
```diff
@@ -120,7 +120,7 @@ func (l *loader) GetApplicationComponents(ctx context.Context, cli client.Client
 	return applicationComponents.Items, nil
 }
 
-// GetEnterpriseContractPolicy returns the EnterpriseContractPolicy referenced by the given ReleaseStrategy. If the
+// GetEnterpriseContractPolicy returns the EnterpriseContractPolicy referenced by the given Strategy. If the
```
johnbieren, line 123: I don't like this change

loader/loader.go
```diff
@@ -165,7 +165,7 @@ func (l *loader) GetReleasePlan(ctx context.Context, cli client.Client, release
 	return releasePlan, getObject(release.Spec.ReleasePlan, release.Namespace, cli, ctx, releasePlan)
 }
 
-// GetReleaseStrategy returns the ReleaseStrategy referenced by the given ReleasePlanAdmission. If the ReleaseStrategy
+// GetReleaseStrategy returns the Strategy referenced by the given ReleasePlanAdmission. If the Strategy
```
johnbieren, line 168: I don't like this change

api/v1alpha1/release_conditions.go
```diff
@@ -0,0 +1,37 @@
...
+
+	// releasedConditionType is the type used to track the status of a Release
+	releasedConditionType conditions.ConditionType = "Released"
+)
+
+const (
+	// DeployedReason is the reason set when a Release has been deployed successfully
+	DeployedReason conditions.ConditionReason = "Deployed"
+
+	// DeployingReason is the reason set when a Release deployment is triggered
```
johnbieren, line 17: has been triggered? or has started? to be more consistent with line 14

conditions/conditions.go
```diff
@@ -0,0 +1,37 @@
+package conditions
+
+import (
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ConditionReason represents a condition reason.
```
johnbieren, line 8: this comment doesn't help at all imo. Remove it or change it would be my suggestion

conditions/conditions.go
```diff
@@ -0,0 +1,37 @@
...
+type ConditionReason string
+
+func (cr ConditionReason) String() string {
+	return string(cr)
+}
+
+// ConditionType represents a condition type.
```
johnbieren, line 15: same as line 8

api/v1alpha1/release_types.go
```diff
@@ -38,95 +38,71 @@ type ReleaseSpec struct {
...
-	// StartTime is the time when the Release PipelineRun was created and set to run
+	// Conditions represent the latest available observations for the release
 	// +optional
-	StartTime *metav1.Time `json:"startTime,omitempty"`
+	Conditions []metav1.Condition `json:"conditions"`
 
-	// CompletionTime is the time the Release PipelineRun completed
 	// +optional
-	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
+	Deployment DeploymentInfo `json:"deployment,omitempty"`
```
johnbieren, line 53: missing comment

api/v1alpha1/release_types.go
```diff
@@ -38,95 +38,71 @@ type ReleaseSpec struct {
...
 
-	// DeploymentStartTime is the time when the SnapshotEnvironmentBinding was created
 	// +optional
-	DeploymentStartTime *metav1.Time `json:"deploymentStartTime,omitempty"`
+	Release ReleaseInfo `json:"release,omitempty"`
```
johnbieren, line 56: missing comment

api/v1alpha1/release_types.go
```diff
@@ -38,95 +38,71 @@ type ReleaseSpec struct {
...
+	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
+
+	// PipelineRun contains the namespaced name of the release PipelineRun executed as part of this release
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?\/[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +optional
-	ReleasePipelineRun string `json:"releasePipelineRun,omitempty"`
+	PipelineRun string `json:"pipelineRun,omitempty"`
 
-	// ReleaseStrategy contains the namespaced name of the ReleaseStrategy used for this release
+	// Strategy contains the namespaced name of the Strategy used for this release
```
johnbieren, line 90: I don't like this change

api/v1alpha1/release_types.go
```diff
@@ -38,95 +38,71 @@ type ReleaseSpec struct {
...
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?\/[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +optional
-	ReleaseStrategy string `json:"releaseStrategy,omitempty"`
```
johnbieren, line 112: I prefer ReleaseStrategy

api/v1alpha1/release_types.go
```diff
@@ -138,128 +114,128 @@ type Release struct {
...
 
-	go metrics.RegisterInvalidRelease(reason.String())
+	go metrics.RegisterCompletedRelease(reason.String(), r.Status.Release.Strategy, r.Status.Target,
+		r.Status.Release.StartTime, r.Status.Release.CompletionTime, false)
 }
 
-// MarkRunning registers the start time and changes the Succeeded condition to Unknown.
-func (r *Release) MarkRunning() {
-	if r.HasStarted() && r.Status.StartTime != nil {
+// MarkReleaseInvalid changes the Succeeded condition to False with the provided reason and message.
```
johnbieren, line 205: it is in the release package. why was `MarkInvalid` not a good enough name?

api/v1alpha1/release_types.go
```diff
@@ -138,128 +114,128 @@ type Release struct {
...
 func (r *Release) IsDeploying() bool {
-	return r.Status.DeploymentStartTime != nil && !r.Status.DeploymentStartTime.IsZero()
+	return r.Status.Deployment.StartTime != nil && !r.Status.Deployment.StartTime.IsZero()
 }
 
-// IsDone returns a boolean indicating whether the Release's status indicates that it is done or not.
-func (r *Release) IsDone() bool {
-	condition := meta.FindStatusCondition(r.Status.Conditions, releaseConditionType)
-	return condition != nil && condition.Status != metav1.ConditionUnknown
+// IsReleased checks whether the Release has succeeded or not.
```
johnbieren, line 139: I don't like `isReleased` and `isReleasing` as function names. Telling someone "the release is releasing" is not helpful at all imo

controllers/release/adapter.go
```diff
@@ -268,7 +266,7 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingIsTracked() (reconciler.Operat
 
 // createReleasePipelineRun creates and returns a new release PipelineRun. The new PipelineRun will include owner
 // annotations, so it triggers Release reconciles whenever it changes. The Pipeline information and the parameters to it
-// will be extracted from the given ReleaseStrategy. The Release's Snapshot will also be passed to the release
+// will be extracted from the given Strategy. The Release's Snapshot will also be passed to the release
```
johnbieren, line 269: I don't like this change

controllers/release/adapter.go
```diff
@@ -19,14 +19,12 @@ package release
 import (
```
johnbieren, line 19: I think the changes to this file are a great example of the old function names being better. Checking if a release.isrunning makes sense logically to read. Checking if a release.isreleasing does not. The constant name changes look good, but I don't like the function name changes

api/v1alpha1/release_types_test.go
```diff
@@ -17,437 +17,315 @@ limitations under the License.
...
-type conditionValues struct {
-	status  metav1.ConditionStatus
-	reason  ReleaseReason
-	message string
-}
-
 var _ = Describe("Release type", func() {
-
-	var r *Release
+	var release *Release
```
johnbieren, line 31: I'm not saying revert this change because I am reviewing it as is and both are fine, but I don't think such a simple change like changing `r` to `release` in a file with this scope gives us any benefit besides making the review process longer because the diff is bigger

api/v1alpha1/release_types.go
```diff
@@ -138,128 +114,128 @@ type Release struct {
 	Status ReleaseStatus `json:"status,omitempty"`
 }
 
-// HasStarted checks whether the Release has a valid start time set in its status.
-func (r *Release) HasStarted() bool {
-	return r.Status.StartTime != nil && !r.Status.StartTime.IsZero()
+// HasDeploymentFinished checks whether the Release has been successfully deployed via GitOps.
+func (r *Release) HasDeploymentFinished() bool {
+	condition := meta.FindStatusCondition(r.Status.Conditions, deployedConditionType.String())
+	return condition != nil && condition.Status != metav1.ConditionUnknown
```
johnbieren, line 120: this reintroduces bugs I already fixed. The logic for deployment status needs to be left as is, reflecting how the gitops code currently works, not changed to what a ticket suggests it should be

api/v1alpha1/release_types.go
```diff
@@ -138,128 +114,128 @@ type Release struct {
 	Status ReleaseStatus `json:"status,omitempty"`
 }
 
-// HasStarted checks whether the Release has a valid start time set in its status.
```
johnbieren, line 141: alphabetizing this list of function makes the review process so much harder because github can't show the diff properly at all

johnbieren (changes requested):
When did we decide to rename `ReleaseStrategy` to `Strategy`?

tekton/pipeline_run_test.go
davidmogar: I think that I messed it up as I refactored the status field. It changed everything else...

The only place where we are going to have `Strategy` instead of `ReleaseStrategy` is in the status itself as it will read the following way:
```
Release:
    PipelineRun: ...
    Strategy: ...
```
I'll fix those occurrences.

loader/loader.go
davidmogar: You are right. Mistake.

api/v1alpha1/release_types.go
davidmogar: This is the only place where is going to read like this. The second `Strategy` should be `ReleaseStrategy` though. Changing it.

api/v1alpha1/release_types.go
davidmogar: In the yaml would appear as:
```
Release:
    ReleaseStrategy:
```
Same goes for code references:
```
release.Status.Release.ReleaseStrategy
```
instead of
```
release.Status.Release.Strategy
```
Not a fan. Exactly the same applies to Go packages and struct names. They "forbid" it.

api/v1alpha1/release_types.go
davidmogar: I agree. Next time I'll have a single commit for the reordering.

api/v1alpha1/release_types.go
davidmogar: You sure about this? Take into account this is referring to the new condition, not the GitOps one. For the new condition, True and False are final conditions which means that the deployment has finished (successfully or not).

api/v1alpha1/release_types.go
davidmogar: They refer to the releasing process. I get your comment though. Propose new names.

api/v1alpha1/release_types.go
davidmogar: Because it was very generic and it wasn't clear if it was referring to the Release or to the Deployment. That MarkInvalid could happen again in the PostActions maybe.

api/v1alpha1/release_types_test.go
davidmogar: I rewrote the test to simplify it and I took the chance to fix a variable with a wrong name. Those kind of names are reserved for receivers, not for variables. Variable names should be meaningful.

conditions/conditions.go
davidmogar: Changed to:
```
// ConditionReason is a string representing a Kubernetes condition reason.
```

conditions/conditions.go
davidmogar: Changed to:
```
// ConditionType is a string representing a Kubernetes condition type.
```

controllers/release/adapter.go
davidmogar: I can agree for the case of the release, but either we have the wrong name for the CRD itself or we need a better name for the "release" process.

If we understand a Release like a process that involved a preparation, a deployment and the execution of post actions, we can think about a better name for that "preparation".

controllers/release/adapter.go
davidmogar: Mistake. Fixed.

loader/loader.go
davidmogar: Mistake. Fixed.

tekton/utils_test.go
davidmogar: Yep, mistake from the name refactoring. Fixed.

loader/loader.go
```diff
@@ -165,7 +165,7 @@ func (l *loader) GetReleasePlan(ctx context.Context, cli client.Client, release
 	return releasePlan, getObject(release.Spec.ReleasePlan, release.Namespace, cli, ctx, releasePlan)
 }
 
-// GetReleaseStrategy returns the ReleaseStrategy referenced by the given ReleasePlanAdmission. If the ReleaseStrategy
+// GetReleaseStrategy returns the ReleaseStrategy referenced by the given ReleasePlanAdmission. If the Strategy
```
johnbieren, line 168: still has the change at the end of the line

api/v1alpha1/release_types.go
johnbieren: I kind of prefer keeping the things in `Status.Release` toplevel. I do not like going to `Strategy`. A tekton task has a `taskref` not just `ref`. I don't think we should be calling things this short name. If you really want it grouped, maybe pick a name for this key not called `Release`

api/v1alpha1/release_types.go
```diff
@@ -138,128 +114,128 @@ type Release struct {
...
-func (r *Release) HasStarted() bool {
-	return r.Status.StartTime != nil && !r.Status.StartTime.IsZero()
+// HasDeploymentFinished checks whether the Release has been successfully deployed via GitOps.
```
johnbieren, line 117: I think `IsDeployed` was a better function name. The code reads better in the adapter. if release.IsDeployed .... "if a release is deployed" makes perfect sense. if release.hasdeploymentfinished ... "if a release has deployment finished" isn't english

api/v1alpha1/release_types.go
```diff
@@ -138,128 +114,128 @@ type Release struct {
...
+func (r *Release) HasDeploymentFinished() bool {
+	condition := meta.FindStatusCondition(r.Status.Conditions, deployedConditionType.String())
+	return condition != nil && condition.Status != metav1.ConditionUnknown
 }
 
-// HasSucceeded checks whether the Release has succeeded or not.
-func (r *Release) HasSucceeded() bool {
-	return meta.IsStatusConditionTrue(r.Status.Conditions, releaseConditionType)
+// HasReleaseFinished returns a boolean indicating whether the Release's status indicates that it is done or not.
```
johnbieren, line 123: this would also become `isXed` when we decide on X

johnbieren (changes requested):

api/v1alpha1/release_types.go
```diff
@@ -38,95 +37,102 @@ type ReleaseSpec struct {
...
+	StartTime *metav1.Time `json:"startTime,omitempty"`
+}
+
+// ValidationInfo defines the observed state of the release validation.
+type ValidationInfo struct {
+	// FailedPostValidation indicates whether the Release was marked as invalid after being initially marked as valid
+	FailedPostValidation bool `json:"failedPostValidation,omitempty"`
+
+	// +optional
+	Time *metav1.Time `json:"startTime,omitempty"`
```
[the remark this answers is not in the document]
davidmogar: Validation doesn't start nor completes. It happens within the same operation. That's why I called it "Time".

api/v1alpha1/release_types.go
```diff
@@ -38,95 +37,102 @@ type ReleaseSpec struct {
...
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?\/[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +optional
-	ReleasePipelineRun string `json:"releasePipelineRun,omitempty"`
+	PipelineRun string `json:"pipelineRun,omitempty"`
 
 	// ReleaseStrategy contains the namespaced name of the ReleaseStrategy used for this release
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?\/[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +optional
-	ReleaseStrategy string `json:"releaseStrategy,omitempty"`
+	ReleaseStrategy string `json:"strategy,omitempty"`
```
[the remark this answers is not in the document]
davidmogar: Done.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+
+// IsDeployed checks whether the Release deployment has finished successfully.
 func (r *Release) IsDeployed() bool {
-	condition := meta.FindStatusCondition(r.Status.Conditions, applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed)
-	return condition != nil && condition.Status == metav1.ConditionTrue
+	return meta.IsStatusConditionTrue(r.Status.Conditions, deployedConditionType.String())
 }
 
-// IsDeploying checks whether the Release has a valid start time for the deployment set in its status.
+// IsDeploying checks whether the Release deployment is progressing.
```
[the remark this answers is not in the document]
davidmogar: Agree.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
-	r.Status.DeploymentCompletionTime = &metav1.Time{Time: time.Now()}
-	r.setStatusConditionWithMessage(applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed,
-		metav1.ConditionTrue, ReleaseReason(reason), message)
+	if !r.IsDeploying() {
+		r.Status.Deployment.StartTime = &metav1.Time{Time: time.Now()}
+	}
 
-	go metrics.RegisterDeployedRelease(reason, r.Status.Target, string(metav1.ConditionTrue),
-		r.Status.DeploymentStartTime, r.Status.DeploymentCompletionTime)
+	conditions.SetConditionWithMessage(&r.Status.Conditions, deployedConditionType, metav1.ConditionFalse, ProgressingReason, message)
```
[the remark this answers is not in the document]
davidmogar: No. We talked about this. Only final conditions (eg. deployed) are True. The rest will be False. This particular function sets the deployment as `deploying` which is not a final state, so the status should be False.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+	return r.isPhaseProgressing(deployedConditionType)
+}
+
+// IsEveryPostActionExecuted checks whether the Release post-actions execution has finished successfully.
+func (r *Release) IsEveryPostActionExecuted() bool {
+	return meta.IsStatusConditionTrue(r.Status.Conditions, postActionsExecutedConditionType.String())
+}
+
+// IsEachPostActionExecuting checks whether the Release post-actions execution is progressing.
+func (r *Release) IsEachPostActionExecuting() bool {
```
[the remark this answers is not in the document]
davidmogar: PostActions will be basically a single Pipeline with multiple actions (tasks) so it's difficult to name the function without using `are`. That's why I called it `isEach` before if we are executing a pipeline, each of the actions (tasks) will be executing. From your options I'd go with `IsAnyPostActionExecuting` but I think it's more confusing tbh.

controllers/release/adapter.go
```diff
@@ -239,31 +168,104 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingExists() (reconciler.Operation
...
 
 	return reconciler.RequeueOnErrorOrContinue(a.client.Status().Patch(a.ctx, a.release, patch))
 }
 
-// EnsureSnapshotEnvironmentBindingIsTracked is an operation that will ensure that the SnapshotEnvironmentBinding
-// Deployment status is tracked in the Release being processed.
-func (a *Adapter) EnsureSnapshotEnvironmentBindingIsTracked() (reconciler.OperationResult, error) {
-	if !a.release.HasSucceeded() || a.release.Status.SnapshotEnvironmentBinding == "" || a.release.IsDeployed() {
+// EnsureReleaseIsRunning is an operation that will ensure that a Release has not finished already and that
+// is marked as releasing. If the Release has finished, no other operation after this one will be executed.
```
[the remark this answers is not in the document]
davidmogar: Done.

controllers/release/adapter.go
```diff
@@ -239,31 +168,104 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingExists() (reconciler.Operation
...
+				"PipelineRun.Name", pipelineRun.Name, "PipelineRun.Namespace", pipelineRun.Namespace)
+		}
+
+		return reconciler.RequeueOnErrorOrContinue(a.registerProcessingData(pipelineRun, resources.ReleaseStrategy))
+	}
+
+	return reconciler.ContinueProcessing()
+}
+
+// EnsureReleaseIsValid is an operation that will ensure that a Release is valid but checking all the resources needed
```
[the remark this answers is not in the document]
davidmogar: Yeah. Lately I put `but` instead of `by` for some reason... Fixed.

controllers/release/adapter.go
```diff
@@ -239,31 +168,104 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingExists() (reconciler.Operation
...
+// to process it.
+func (a *Adapter) EnsureReleaseIsValid() (reconciler.OperationResult, error) {
```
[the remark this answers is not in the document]
davidmogar: As you said the change is big, so I could be wrong, but if I'm not mistaken, that happens here:
```
if resources.ReleasePlanAdmission == nil {
    a.release.MarkValidationFailed(err.Error())
    a.release.MarkReleaseFailed("Release validation failed")
    return reconciler.RequeueOnErrorOrStop(a.client.Status().Patch(a.ctx, a.release, patch))
}
```
Basically any error getting the ReleasePlanAdmission is a validation error. Within the function to get the ReleasePlanAdmission 3 errors are thrown:
- multiple ReleasePlanAdmissions found
- auto-release label set to false
- no ReleasePlanAdmission found

So, as all of them are a validation error, I just set the validation message to the one thrown.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+	return r.hasPhaseFinished(releasedConditionType)
 }
 
-// IsDeployed checks whether the Release has been successfully deployed via GitOps.
+// HasValidationFinished checks whether the Release validation has finished.
+func (r *Release) HasValidationFinished() bool {
+	return r.hasPhaseFinished(validatedConditionType)
+}
+
+// IsDeployed checks whether the Release deployment has finished successfully.
```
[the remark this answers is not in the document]
davidmogar: Changed the others as well.

api/v1alpha1/release_conditions.go
```diff
@@ -0,0 +1,31 @@
+package v1alpha1
+
+import "github.com/redhat-appstudio/release-service/conditions"
+
+const (
+	// deployedConditionType is the type used to track the status of a Release deployment
+	deployedConditionType conditions.ConditionType = "Deployed"
+
+	// postActionsExecutedConditionType is the type used to track the status of a Release post-actions
```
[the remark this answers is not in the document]
davidmogar: Done.

api/v1alpha1/release_types.go
johnbieren: Didn't we already pick process for this?

api/v1alpha1/release_types.go
```diff
@@ -138,128 +114,128 @@ type Release struct {
...
+func (r *Release) MarkDeployed() {
+	if !r.IsDeploying() || (r.HasDeploymentFinished() && r.Status.Deployment.CompletionTime != nil) {
 		return
 	}
 
-	r.Status.DeploymentCompletionTime = &metav1.Time{Time: time.Now()}
-	r.setStatusConditionWithMessage(applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed,
-		metav1.ConditionTrue, ReleaseReason(reason), message)
+	r.Status.Deployment.CompletionTime = &metav1.Time{Time: time.Now()}
+	conditions.SetCondition(&r.Status.Conditions, deployedConditionType, metav1.ConditionTrue, DeployedReason)
```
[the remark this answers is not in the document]
johnbieren: The *whole* point of copying the status is to *not* have to go check the SEB. Why are we saying we are smarter than gitops and are adding our own logic layer to maintain (and for anyone looking at these CRs to understand)? Why can't we (continue to) say we are bubbling up the status from the SEB in the release? That's it. If someone says to us "the status in the release for deployment doesn't make much sense" we say that we are just bubbling up the status from the SEB, take it up with them. If the pipeline-service team had their own logic interpreting task results and put that on their pipelineruns instead of what the task actually reported, I would be extremely disappointed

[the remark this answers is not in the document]
davidmogar: This is an old comment that was pending and sent with my last changes. What we discussed and agreed on offline (what we have implemented right now) is what prevails.

api/v1alpha1/release_types.go
johnbieren: Processed

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
johnbieren: it is called startTime in the json. Can we be consistent

api/v1alpha1/release_types.go
```diff
@@ -138,128 +114,128 @@ type Release struct {
...
 // currently treats False as a transient status, so we accept False (temporarily) as a status here until it is changed.
-func (r *Release) MarkDeploying(status metav1.ConditionStatus, reason, message string) {
-	if status != metav1.ConditionTrue {
-		r.setStatusConditionWithMessage(applicationapiv1alpha1.ComponentDeploymentConditionAllComponentsDeployed,
-			status, ReleaseReason(reason), message)
-
-		if r.Status.DeploymentStartTime == nil {
-			r.Status.DeploymentStartTime = &metav1.Time{Time: time.Now()}
-		}
+func (r *Release) MarkDeploying(message string) {
```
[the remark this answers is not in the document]
davidmogar: Old comment. Apologies.

api/v1alpha1/release_types.go
davidmogar: Old comment.

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
johnbieren: If the "post actions" are contained in a single pipelinerun, why can't we call it "post action" .. the operator is doing one action - starting a single pipelinerun

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
davidmogar: Missing a generate. Fixing it.

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
davidmogar: Cause they could be unrelated actions (eg. posting to slack and pushing something to some other place).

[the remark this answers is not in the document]
johnbieren: then by that logic processing should be processings because they could be doing unrelated actions too. I think it makes things nice to use a singular word because then all the functions can be nicely and similarly named, and if it is just one pipelinerun, we can get away with using the singular, even if multiple tasks are in it

controllers/release/adapter.go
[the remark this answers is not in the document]
davidmogar: It should be there now.

metrics/release.go
```diff
@@ -17,128 +17,197 @@ limitations under the License.
...
-		"strategy":  strategy,
-		"succeeded": strconv.FormatBool(succeeded),
-		"target":    target,
+func RegisterCompletedRelease(release *v1alpha1.Release) {
+	if !release.IsReleased() {
+		return
 	}
 
+	labels := prometheus.Labels{
+		"deployment_reason":   "",
```
johnbieren, line 130: do you think instead of the empty string we should have like n/a or something that shows we are explicitly setting it to be this way, not that some operation was missed or something (not sure i am making sense). or even "success" as some of these we know are success in this function (like processing_reason)

metrics/release.go
```diff
@@ -17,128 +17,197 @@ limitations under the License.
...
-		"target":    "",
-	}).Inc()
+func RegisterCompletedReleaseProcessing(release *v1alpha1.Release) {
+	if !release.HasProcessingFinished() {
+		return
+	}
+
+	ReleaseAttemptProcessingDurationSeconds.
+		With(prometheus.Labels{
+			"reason":           "",
```
johnbieren, line 166: why is reason blank for all of these? If we are never providing a reason, why is it even here?

johnbieren (commented):
I did not review any _test.go files as AIUI those are not ready

controllers/release/adapter.go
```diff
@@ -371,9 +369,13 @@ func (a *Adapter) registerGitOpsDeploymentStatus(binding *applicationapiv1alpha1
 	patch := client.MergeFrom(a.release.DeepCopy())
 
 	if condition.Status == metav1.ConditionTrue {
-		a.release.MarkDeployed(condition.Reason, condition.Message)
+		a.release.MarkDeployed()
```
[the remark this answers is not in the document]
davidmogar: Already discussed offline. We went for no message at all as "deployed" was information enough.

controllers/release/adapter.go
```diff
@@ -371,9 +369,13 @@ func (a *Adapter) registerGitOpsDeploymentStatus(binding *applicationapiv1alpha1
...
 	} else {
-		a.release.MarkDeploying(condition.Status, condition.Reason, condition.Message)
+		if condition.Reason == applicationapiv1alpha1.ComponentDeploymentConditionErrorOccurred {
```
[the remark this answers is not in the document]
davidmogar: Discussed offline.

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
davidmogar: Fixed. Now both files use `Time`.

metrics/release.go
davidmogar: Fixed.

metrics/release.go
davidmogar: I think I missed a push maybe... Fixed though.

metrics/release.go
```diff
@@ -17,128 +17,197 @@ limitations under the License.
...
+	)
+
+	ReleaseAttemptDeploymentDurationSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "release_attempt_deployment_duration_seconds",
+			Help:    "How long in seconds a Release deployment takes to complete",
+			Buckets: []float64{60, 150, 300, 450, 600, 750, 900, 1050, 1200, 1800, 3600},
+		},
+		[]string{
+			"reason",
```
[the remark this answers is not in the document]
davidmogar: There already.

metrics/release.go
```diff
@@ -17,128 +17,197 @@ limitations under the License.
...
 	)
 
-	ReleaseAttemptDeploymentTotal = prometheus.NewCounterVec(
-		prometheus.CounterOpts{
-			Name: "release_attempt_deployment_total",
-			Help: "Total number of deployments released to managed environments by the operator",
+	ReleaseAttemptConcurrentPostActionsExecutionsTotal = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "release_attempt_concurrent_post_actions_executions_total",
+			Help: "Total number of concurrent release processing attempts",
```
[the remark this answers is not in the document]
davidmogar: Fixed.

controllers/release/adapter.go
```diff
@@ -401,21 +407,21 @@ func (a *Adapter) registerReleasePipelineRunStatus(pipelineRun *v1beta1.Pipeline
 	return nil
 }
 
-// registerReleaseStatusData adds all the Release information to its Status.
-func (a *Adapter) registerReleaseStatusData(releasePipelineRun *v1beta1.PipelineRun, releaseStrategy *v1alpha1.ReleaseStrategy) error {
+// registerProcessingData adds all the Release information to its Status and marks it as processing.
```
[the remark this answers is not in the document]
davidmogar: Changed to `// registerProcessingData adds all the Release processing information to its Status and marks it as processing.`.

controllers/release/adapter.go
```diff
@@ -239,31 +168,104 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingExists() (reconciler.Operation
...
+
+		return reconciler.RequeueOnErrorOrContinue(a.registerProcessingData(pipelineRun, resources.ReleaseStrategy))
+	}
+
+	return reconciler.ContinueProcessing()
+}
+
+// EnsureReleaseIsValid is an operation that will ensure that a Release is valid by checking all the resources needed
+// to process it.
+func (a *Adapter) EnsureReleaseIsValid() (reconciler.OperationResult, error) {
```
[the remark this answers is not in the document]
davidmogar: :a: :b: :cake: :man_shrugging:

controllers/release/adapter.go
```diff
@@ -119,99 +115,32 @@ func (a *Adapter) EnsureFinalizerIsAdded() (reconciler.OperationResult, error) {
...
+	return reconciler.RequeueOnErrorOrContinue(a.registerDeploymentStatus(binding))
 }
 
-// EnsureSnapshotEnvironmentBindingExists is an operation that will ensure that a SnapshotEnvironmentBinding
+// EnsureReleaseIsDeployed is an operation that will ensure that a SnapshotEnvironmentBinding
 // associated to the Release being processed exists. Otherwise, it will create a new one.
-func (a *Adapter) EnsureSnapshotEnvironmentBindingExists() (reconciler.OperationResult, error) {
-	if !a.release.HasSucceeded() || a.release.IsDeployed() {
+func (a *Adapter) EnsureReleaseIsDeployed() (reconciler.OperationResult, error) {
+	if !a.release.IsReleased() || a.release.HasDeploymentFinished() {
```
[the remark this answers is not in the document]
davidmogar: Fixed. Those were difficult to track with the refactor...

controllers/release/adapter.go
```diff
@@ -119,99 +115,32 @@ func (a *Adapter) EnsureFinalizerIsAdded() (reconciler.OperationResult, error) {
...
-}
-
-// EnsureReleasePipelineStatusIsTracked is an operation that will ensure that the release PipelineRun status is tracked
-// in the Release being processed.
-func (a *Adapter) EnsureReleasePipelineStatusIsTracked() (reconciler.OperationResult, error) {
-	if !a.release.HasStarted() || a.release.IsDone() {
+// EnsureReleaseDeploymentIsTracked is an operation that will ensure that the SnapshotEnvironmentBinding
+// Deployment status is tracked in the Release being processed.
+func (a *Adapter) EnsureReleaseDeploymentIsTracked() (reconciler.OperationResult, error) {
+	if !a.release.IsReleased() || a.release.Status.Deployment.SnapshotEnvironmentBinding == "" || a.release.HasDeploymentFinished() {
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -17,10 +17,9 @@ limitations under the License.
 package v1alpha1
```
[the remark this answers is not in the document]
davidmogar: You will have to explain this to me. Why this is not good enough:

- HasDeploymentFinished checks whether the Release deployment has finished.
- IsDeployed checks whether the Release was successfully deployed.
- IsDeploying checks whether the Release deployment is in progress.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
-// SetCondition creates a new condition with the given conditionType, status, reason and message. Then, it sets this new condition,
-// unsetting previous conditions with the same type as necessary.
-func (r *Release) setStatusConditionWithMessage(conditionType string, status metav1.ConditionStatus, reason ReleaseReason, message string) {
-	meta.SetStatusCondition(&r.Status.Conditions, metav1.Condition{
-		Type:    conditionType,
-		Status:  status,
-		Reason:  reason.String(),
-		Message: message,
-	})
+// isPhaseProgressing checks whether a Release phase is progressing.
```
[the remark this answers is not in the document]
davidmogar: Same fix.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
-		r.Status.StartTime, r.Status.CompletionTime, true)
+	r.Status.Validation.Time = &metav1.Time{Time: time.Now()}
+	conditions.SetConditionWithMessage(&r.Status.Conditions, validatedConditionType, metav1.ConditionFalse, FailedReason, message)
 }
 
-// SetCondition creates a new condition with the given conditionType, status and reason. Then, it sets this new condition,
-// unsetting previous conditions with the same type as necessary.
-func (r *Release) setStatusCondition(conditionType string, status metav1.ConditionStatus, reason ReleaseReason) {
-	r.setStatusConditionWithMessage(conditionType, status, reason, "")
+// hasPhaseFinished checks whether a Release phase has finished.
```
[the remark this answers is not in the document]
davidmogar: Changed to `// hasPhaseFinished checks whether a Release phase (e.g. deployment or processing) has finished.`.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+	if !r.IsProcessing() {
+		r.Status.Processing.StartTime = &metav1.Time{Time: time.Now()}
+	}
+
+	conditions.SetConditionWithMessage(&r.Status.Conditions, processedConditionType, metav1.ConditionFalse, ProgressingReason, message)
+}
+
+// MarkPostActionsExecutionFailed marks the Release post-actions execution as failed.
+func (r *Release) MarkPostActionsExecutionFailed(message string) {
+	if !r.IsProcessing() || r.HasProcessingFinished() {
```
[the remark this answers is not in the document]
davidmogar: No idea :thinking:

Fixed :stuck_out_tongue_closed_eyes:

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
-	if r.HasStarted() && r.Status.StartTime != nil {
+// MarkPostActionsExecuting marks the Release post-actions as executing.
+func (r *Release) MarkPostActionsExecuting(message string) {
+	if r.HasProcessingFinished() {
 		return
 	}
 
-	r.Status.StartTime = &metav1.Time{Time: time.Now()}
-	r.setStatusCondition(releaseConditionType, metav1.ConditionUnknown, ReleaseReasonRunning)
+	if !r.IsProcessing() {
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+	r.Status.Processing.CompletionTime = &metav1.Time{Time: time.Now()}
+	conditions.SetCondition(&r.Status.Conditions, processedConditionType, metav1.ConditionTrue, SucceededReason)
 }
 
-// MarkRunning registers the start time and changes the Succeeded condition to Unknown.
-func (r *Release) MarkRunning() {
-	if r.HasStarted() && r.Status.StartTime != nil {
+// MarkPostActionsExecuting marks the Release post-actions as executing.
+func (r *Release) MarkPostActionsExecuting(message string) {
+	if r.HasProcessingFinished() {
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
 
-	go metrics.RegisterInvalidRelease(reason.String())
+// MarkPostActionsExecuted marks the Release post-actions as executed.
+func (r *Release) MarkPostActionsExecuted() {
+	if !r.IsProcessing() || r.HasProcessingFinished() {
+		return
+	}
+
+	r.Status.Processing.CompletionTime = &metav1.Time{Time: time.Now()}
+	conditions.SetCondition(&r.Status.Conditions, processedConditionType, metav1.ConditionTrue, SucceededReason)
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+}
 
-	go metrics.RegisterInvalidRelease(reason.String())
+// MarkPostActionsExecuted marks the Release post-actions as executed.
+func (r *Release) MarkPostActionsExecuted() {
+	if !r.IsProcessing() || r.HasProcessingFinished() {
+		return
+	}
+
+	r.Status.Processing.CompletionTime = &metav1.Time{Time: time.Now()}
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
 
-	r.setStatusConditionWithMessage(releaseConditionType, metav1.ConditionFalse, reason, message)
+	r.Status.Processing.CompletionTime = &metav1.Time{Time: time.Now()}
+	conditions.SetConditionWithMessage(&r.Status.Conditions, processedConditionType, metav1.ConditionFalse, FailedReason, message)
+}
 
-	go metrics.RegisterInvalidRelease(reason.String())
+// MarkPostActionsExecuted marks the Release post-actions as executed.
+func (r *Release) MarkPostActionsExecuted() {
+	if !r.IsProcessing() || r.HasProcessingFinished() {
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
-	return r.Status.DeploymentStartTime != nil && !r.Status.DeploymentStartTime.IsZero()
+	return r.isPhaseProgressing(deployedConditionType)
+}
+
+// IsEveryPostActionExecuted checks whether the Release post-actions were successfully executed.
+func (r *Release) IsEveryPostActionExecuted() bool {
+	return meta.IsStatusConditionTrue(r.Status.Conditions, postActionsExecutedConditionType.String())
+}
+
+// IsEachPostActionExecuting checks whether the Release post-actions execution is progressing.
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+func (r *Release) IsProcessing() bool {
+	return r.isPhaseProgressing(processedConditionType)
+}
+
+// IsReleased checks whether the Release has finished successfully.
+func (r *Release) IsReleased() bool {
+	return meta.IsStatusConditionTrue(r.Status.Conditions, releasedConditionType.String())
+}
+
+// IsReleasing checks whether the Release is progressing.
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+func (r *Release) IsEachPostActionExecuting() bool {
+	return r.isPhaseProgressing(postActionsExecutedConditionType)
+}
+
+// IsProcessed checks whether the Release was successfully processed..
+func (r *Release) IsProcessed() bool {
+	return meta.IsStatusConditionTrue(r.Status.Conditions, processedConditionType.String())
+}
+
+// IsProcessing checks whether the Release processing is progressing.
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
```diff
@@ -138,128 +144,238 @@ type Release struct {
...
+func (r *Release) IsEveryPostActionExecuted() bool {
+	return meta.IsStatusConditionTrue(r.Status.Conditions, postActionsExecutedConditionType.String())
+}
+
+// IsEachPostActionExecuting checks whether the Release post-actions execution is progressing.
+func (r *Release) IsEachPostActionExecuting() bool {
+	return r.isPhaseProgressing(postActionsExecutedConditionType)
+}
+
+// IsProcessed checks whether the Release was successfully processed..
```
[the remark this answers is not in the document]
davidmogar: Fixed.

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
johnbieren: Not fixed

controllers/release/adapter.go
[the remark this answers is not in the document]
johnbieren: abc order --

controllers/release/adapter.go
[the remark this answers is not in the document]
johnbieren: If a release is not already deployed, continue processing ... that sounds opposite of what we want to me.

controllers/release/adapter.go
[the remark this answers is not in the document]
johnbieren: I think this if condition needs a revisit. I don't think it is right still

metrics/release.go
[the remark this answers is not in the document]
johnbieren: it wasn't lol but looks good now

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
johnbieren: I think it would be nice to say the Has functions check if it is finished, regardless of successful or not successful

controllers/release/adapter.go
```diff
@@ -355,9 +373,31 @@ func (a *Adapter) finalizeRelease() error {
 	return nil
 }
 
-// registerGitOpsDeploymentStatus updates the status of the Release being processed by monitoring the status of the
+// registerDeploymentData adds all the Release information to its Status and marks it as processing.
```
johnbieren, line 376: just like in a similar function (that you fixed the comment), it isn't all the Release information, just the deployment information

controllers/release/adapter.go
```diff
@@ -119,99 +115,51 @@ func (a *Adapter) EnsureFinalizerIsAdded() (reconciler.OperationResult, error) {
...
-}
-
-// EnsureReleasePipelineRunExists is an operation that will ensure that a release PipelineRun associated to the Release
-// being processed exists. Otherwise, it will create a new release PipelineRun.
-func (a *Adapter) EnsureReleasePipelineRunExists() (reconciler.OperationResult, error) {
-	pipelineRun, err := a.loader.GetReleasePipelineRun(a.ctx, a.client, a.release)
-	if err != nil && !errors.IsNotFound(err) {
-		return reconciler.RequeueWithError(err)
+// EnsureReleaseIsCompleted is an operation that will ensure that a Release is completed (marked as released) when
+// all the phases (e.g. deployment or processing) have been completed if required.
```
johnbieren, line 119: reads better to say `required phases` instead of `phases .. if required`

controllers/release/adapter.go
```diff
@@ -119,99 +115,51 @@ func (a *Adapter) EnsureFinalizerIsAdded() (reconciler.OperationResult, error) {
...
-	return reconciler.ContinueProcessing()
+	return reconciler.RequeueOnErrorOrContinue(a.registerDeploymentStatus(binding))
 }
 
-// EnsureSnapshotEnvironmentBindingExists is an operation that will ensure that a SnapshotEnvironmentBinding
+// EnsureReleaseIsDeployed is an operation that will ensure that a SnapshotEnvironmentBinding
 // associated to the Release being processed exists. Otherwise, it will create a new one.
-func (a *Adapter) EnsureSnapshotEnvironmentBindingExists() (reconciler.OperationResult, error) {
-	if !a.release.HasSucceeded() || a.release.IsDeployed() {
+func (a *Adapter) EnsureReleaseIsDeployed() (reconciler.OperationResult, error) {
```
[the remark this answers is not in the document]
davidmogar: We are missing the code to set the deployment as outdated. Didn't we said that we would do it in a separate PR? I can do it now if you want.

controllers/release/adapter.go
davidmogar: Fixed.

controllers/release/adapter.go
davidmogar: Changed.

controllers/release/adapter.go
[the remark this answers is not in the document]
johnbieren: It can be a separate PR, but I didn't realize we had this bug now until reviewing this pr. If we already have the bug, keep the bug until after this pr then we can fix it. No more adding to this pr

metrics/release.go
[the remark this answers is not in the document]
johnbieren: That's fine by me

controllers/release/adapter.go
johnbieren: I still like the word all in there :slightly_smiling_face: `when all required phases`

controllers/release/adapter_test.go
```diff
@@ -656,58 +432,150 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
 		})
 
-		It("contains a parameter with the json representation of the Snapshot", func() {
-			jsonSpec, _ := json.Marshal(snapshot.Spec)
-			Expect(pipelineRun.Spec.Params).Should(ContainElement(HaveField("Value.StringVal", Equal(string(jsonSpec)))))
+		It("should do nothing if the Release is already processed", func() {
+			adapter.release.MarkProcessing("")
+			adapter.release.MarkProcessed()
+
+			result, err := adapter.EnsureReleaseIsRunning()
```
johnbieren, line 487: shouldn't this be `EnsureReleaseIsProcessed`?

controllers/release/adapter_test.go
```diff
@@ -656,58 +432,150 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+				},
+			})
+			adapter.release.MarkProcessing("")
+
+			result, err := adapter.EnsureReleaseIsProcessed()
+			Expect(!result.RequeueRequest && !result.CancelRequest).To(BeTrue())
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("should register the processing data if the pipelineRun already exists", func() {
```
johnbieren, line 512: on line 493 you capitalize it as PipelineRun. Let's be consistent

controllers/release/adapter_test.go
```diff
@@ -754,74 +664,101 @@ var _ = Describe("Release Adapter", Ordered, func() {
 			adapter = createReleaseAndAdapter()
 		})
 
-		It("does nothing if there is no PipelineRun", func() {
-			Expect(adapter.registerProcessingStatus(nil)).To(Succeed())
-			Expect(adapter.release.Status.Release.CompletionTime).To(BeNil())
+		It("should continue if the Release processing has not", func() {
```
johnbieren, line 667: missing a word

controllers/release/adapter_test.go
```diff
@@ -935,6 +872,170 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+		AfterEach(func() {
+			_ = adapter.client.Delete(ctx, adapter.release)
+		})
+
+		BeforeEach(func() {
+			adapter = createReleaseAndAdapter()
+		})
+
+		It("does nothing if there is no binding", func() {
+			Expect(adapter.registerDeploymentData(nil, nil)).To(Succeed())
```
johnbieren, line 887: should be passing `nil, releasePlanAdmission`

controllers/release/adapter_test.go
```diff
@@ -935,6 +872,170 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			Expect(adapter.registerDeploymentStatus(nil)).To(Succeed())
+			Expect(adapter.release.IsDeploying()).To(BeFalse())
+		})
+
+		It("does nothing if the binding doesn't have the expected condition", func() {
+			Expect(adapter.registerDeploymentStatus(snapshotEnvironmentBinding)).To(Succeed())
+			Expect(adapter.release.IsDeploying()).To(BeFalse())
+		})
+
+		It("registers the deployment status when the binding has been deployed", func() {
```
johnbieren, line 938: could probably add a test for the 
```
	} else {
		if condition.Reason == applicationapiv1alpha1.ComponentDeploymentConditionErrorOccurred {
			a.release.MarkDeploymentFailed(condition.Message)
			a.release.MarkReleaseFailed("Release deployment failed")
		} else {
			a.release.MarkDeploying(condition.Message)
		}
	}
```
piece of the code

controllers/release/adapter_test.go
```diff
@@ -935,6 +872,170 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+		AfterEach(func() {
+			_ = adapter.client.Delete(ctx, adapter.release)
+		})
+
+		BeforeEach(func() {
+			adapter = createReleaseAndAdapter()
+		})
+
+		It("does nothing if there is no PipelineRun", func() {
+			Expect(adapter.registerProcessingData(nil, nil)).To(Succeed())
```
johnbieren, line 965: should be passing `nil, releaseStrategy`

controllers/release/adapter_test.go
```diff
@@ -935,6 +872,170 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			_ = adapter.client.Delete(ctx, adapter.release)
+		})
+
+		BeforeEach(func() {
+			adapter = createReleaseAndAdapter()
+		})
+
+		It("does nothing if there is no PipelineRun", func() {
+			Expect(adapter.registerProcessingStatus(nil)).To(Succeed())
+			Expect(adapter.release.Status.CompletionTime).To(BeNil())
```
johnbieren, line 1010: shouldn't this be checking the processing completion time?

controllers/release/adapter_test.go
```diff
@@ -935,6 +872,170 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+		})
+
+		It("does nothing if the PipelineRun is not done", func() {
+			pipelineRun := &v1beta1.PipelineRun{}
+			Expect(adapter.registerProcessingStatus(pipelineRun)).To(Succeed())
+			Expect(adapter.release.Status.CompletionTime).To(BeNil())
```
johnbieren, line 1016: shouldn't this be checking the processing completion time?

metrics/metrics_suite_test.go
```diff
@@ -0,0 +1,94 @@
...
+	var metricLabels string
+
+	for _, label := range labels {
+		metricLabels += fmt.Sprintf("%s=\"%s\",", label, label)
+	}
+
+	return metricLabels
+}
+
+// newHistogramReader generates a strings.Reader to be used by Prometheus' testutils. The Reader will contain the
```
johnbieren, line 63: wrong func name

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
-			Name:    "release_deployment_duration_seconds",
-			Help:    "How long in seconds a Release deployment takes to complete",
-			Buckets: []float64{60, 150, 300, 450, 600, 750, 900, 1050, 1200, 1800, 3600},
-		},
-		[]string{
-			"environment",
-			"reason",
-			"target",
-		},
+		releaseDeploymentDurationSecondsOpts,
```
johnbieren, line 59: why can't these just be defined in line?

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
+		releaseDeploymentDurationSecondsLabels,
```
johnbieren, line 60: same as line 59

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
-		[]string{
-			"deployment_reason",
-			"post_actions_reason",
-			"processing_reason",
-			"release_reason",
-			"validation_reason",
-			"release_strategy",
-			"target",
-		},
+		releaseDurationSecondsOpts,
```
johnbieren, line 74: same as l59

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
+		releaseDurationSecondsLabels,
```
johnbieren, line 75: same as l59

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
 	ReleasePostActionsExecutionDurationSeconds = prometheus.NewHistogramVec(
-		prometheus.HistogramOpts{
-			Name:    "release_post_actions_execution_duration_seconds",
-			Help:    "How long in seconds Release post-actions take to complete",
-			Buckets: []float64{60, 150, 300, 450, 600, 750, 900, 1050, 1200, 1800, 3600},
-		},
-		[]string{
-			"reason",
-		},
+		releasePostActionsExecutionDurationSecondsOpts,
```
johnbieren, line 93: same as l59

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
+		releasePostActionsExecutionDurationSecondsLabels,
```
johnbieren, line 94: same as l59

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
-			Name:    "release_processing_duration_seconds",
-			Help:    "How long in seconds a Release processing takes to complete",
-			Buckets: []float64{60, 150, 300, 450, 600, 750, 900, 1050, 1200, 1800, 3600},
-		},
-		[]string{
-			"reason",
-			"release_strategy",
-			"target",
-		},
+		releaseProcessingDurationSecondsOpts,
```
johnbieren, line 106: same as l59

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
+		releaseProcessingDurationSecondsLabels,
```
johnbieren, line 107: same as l59

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
-		[]string{
-			"deployment_reason",
-			"post_actions_reason",
-			"processing_reason",
-			"release_reason",
-			"release_strategy",
-			"target",
-			"validation_reason",
-		},
+		releaseTotalOpts,
```
johnbieren, line 121: same as l59

metrics/release.go
```diff
@@ -23,111 +23,124 @@ import (
...
+		releaseTotalLabels,
```
johnbieren, line 122: same as l59

metrics/release_test.go
```diff
@@ -17,239 +17,225 @@ limitations under the License.
...
 		})
 
-		It("registers a new observation for 'release_attempt_running_seconds' with the elapsed time from the moment"+
-			"the Release was created to when it started (Release marked as 'Running').", func() {
-			// Defined buckets for ReleaseAttemptRunningSeconds
-			timeBuckets := []string{"1", "5", "10", "30"}
-			data := []int{1, 2, 3, 4}
-			readerData := createHistogramReader(AttemptRunningSecondsHeader, timeBuckets, data, "", elapsedSeconds, len(inputSeconds))
-			Expect(testutil.CollectAndCompare(ReleaseAttemptRunningSeconds, strings.NewReader(readerData))).To(Succeed())
+		It("increments ReleaseTotal", func() {
```
johnbieren, line 72: same name as line 66

metrics/release_test.go
```diff
@@ -17,239 +17,225 @@ limitations under the License.
...
+			RegisterCompletedRelease(startTime, completionTime, "", "", "", "", "", "", "")
+			Expect(testutil.ToFloat64(ReleaseConcurrentTotal.WithLabelValues())).To(Equal(float64(-1)))
 		})
 
-		// Input seconds for duration of operations less or equal to the following buckets of 1, 5, 10 and 30 seconds
-		inputSeconds := []float64{1, 3, 8, 15}
-		elapsedSeconds := 0.0
+		It("adds an observation to ReleaseDurationSeconds", func() {
+			RegisterCompletedRelease(startTime, completionTime,
+				releaseDurationSecondsLabels[0],
```
johnbieren, line 50: before you reply to my other comment with "so I can use them in the tests," why don't we just use empty string or something. I don't like making the code less pretty just to help the tests be shorter

metrics/release_test.go
```diff
@@ -17,239 +17,225 @@ limitations under the License.
...
-				prometheus.GaugeOpts{
-					Name: "release_attempt_concurrent_requests",
-					Help: "Total number of concurrent release attempts",
-				},
-			)
-			metrics.Registry.MustRegister(ReleaseAttemptRunningSeconds, ReleaseDurationSeconds, ReleaseConcurrentTotal)
+		BeforeEach(func() {
+			initializeMetrics()
+
+			completionTime = &metav1.Time{}
```
johnbieren, line 96: there should be a test about doing nothing when the completion or start time are nil

metrics/release_test.go
```diff
@@ -17,239 +17,225 @@ limitations under the License.
...
+				))).To(Succeed())
 		})
+	})
 
-		// Input seconds for duration of operations less or equal to the following buckets of 60, 600, 1800 and 3600 seconds
-		inputSeconds := []float64{30, 500, 1500, 3000}
-		elapsedSeconds := 0.0
-		labels := fmt.Sprintf(`reason="%s", strategy="%s", succeeded="true", target="%s",`,
-			validReleaseReason, strategy, defaultNamespace)
+	Context("When RegisterCompletedReleaseProcessing is called", func() {
```
johnbieren, line 121: why is this in a different order than the functions are defined in release.go?

metrics/release_test.go
johnbieren, line 1: to avoid repeating it, I think there should be a test case for each of the RegisterCompletedRelease* tests for when the times are nil

api/v1alpha1/release_types_test.go
```diff
@@ -21,310 +21,1070 @@ import (
...
 	"github.com/redhat-appstudio/release-service/conditions"
-	"k8s.io/apimachinery/pkg/api/meta"
-	"time"
 
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 var _ = Describe("Release type", func() {
-	var release *Release
```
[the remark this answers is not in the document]
davidmogar: Because I don't want releases to be reused for tests in the same context. The alternative would be to reset conditions or keep track of the status of previous tests. We don't want any of those.

api/v1alpha1/release_types_test.go
```diff
@@ -21,310 +21,1070 @@ import (
...
+	Context("When MarkReleaseFailed method is called", func() {
+		var release *Release
+
+		BeforeEach(func() {
+			release = &Release{}
+		})
+
+		It("should do nothing if the Release has not started", func() {
+			release.MarkReleaseFailed("")
+			Expect(release.Status.Deployment.CompletionTime).To(BeNil())
```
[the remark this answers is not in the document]
davidmogar: Good catch.

controllers/release/adapter_test.go
davidmogar: Good catch.

controllers/release/adapter_test.go
davidmogar: Agree.

controllers/release/adapter_test.go
davidmogar: Good catch.

controllers/release/adapter_test.go
davidmogar: Yeah, that's better actually.

controllers/release/adapter_test.go
davidmogar: Done.

controllers/release/adapter_test.go
davidmogar: Yeah, that's better actually.

controllers/release/adapter_test.go
davidmogar: Indeed.

controllers/release/adapter_test.go
davidmogar: Indeed.

metrics/metrics_suite_test.go
davidmogar: Fixed.

metrics/release_test.go
davidmogar: I'm going to answer only here. The reason I'm not doing it inline is because we want to use the real labels and options defined for the metrics. Passing empty strings there would make the test much easier and that's my previous version of the tests, but it wouldn't ensure that the function called is setting the labels when calling the metric and it wouldn't check that each parameter goes to the right label.

With the current code you have that ugly piece call, but then through `getMetricLabelsString` you ensure that everything is used properly in each function.

metrics/release_test.go
davidmogar: Added.

metrics/release_test.go
davidmogar: My bad. Fixed.

metrics/release_test.go
davidmogar: Fixed.

metrics/release_test.go
johnbieren: @davidmogar can you use a different It string? You have `increments ReleaseTotal` twice in a row

johnbieren (changes requested):

metrics/release_test.go
```diff
@@ -17,239 +17,273 @@ limitations under the License.
...
+
+			completionTime = &metav1.Time{}
+			startTime = &metav1.Time{Time: completionTime.Add(-60 * time.Second)}
 		})
 
-		// Input seconds for duration of operations less or equal to the following buckets of 1, 5, 10 and 30 seconds
-		inputSeconds := []float64{1, 3, 8, 15}
-		elapsedSeconds := 0.0
+		It("does nothing if the start time is nil", func() {
+			Expect(testutil.ToFloat64(ReleaseConcurrentTotal.WithLabelValues())).To(Equal(float64(0)))
```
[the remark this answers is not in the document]
davidmogar: Bad copy paste. Good catch. What would I do without you :heart_eyes:

metrics/release_test.go
davidmogar: Actually is the same test. I removed the one that doesn't use labels.

api/v1alpha1/release_types_test.go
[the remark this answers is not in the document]
davidmogar: Yeah, that's what I mean. If you use the higher level `BeforeEach`, that block will be executed once per `Context`, but not once per each test in each context, so whatever you declare there, will be reused within the `Context`.

johnbieren (approved):

johnbieren (approved):

johnbieren (approved):

PR #180, opened 2023-03-08
feat(HACBS-1764): cleanup tls errors from webhook in manager pod

johnbieren:
This commit removes the webhook and cainjection files from the crd kustomization file. This removes the tls cert errors that currently pollute the manager pod log. The webhooks still work properly with this change.


happybhati (approved):
LGTM

PR #185, opened 2023-04-27
feat(HACBS-2125): capture user who creates Release

johnbieren:
This commit modifies the Release webhook to be a low level one. The main reason for this is so that whenever a Release CR is created or modified, a label is set which depicts the username of who created or modified the CR. This value will end up being used in release pipelines.

### DIFF
--- a/api/v1alpha1/release_webhook.go
+++ b/api/v1alpha1/release_webhook.go
@@ -17,38 +17,116 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"fmt"
-	"k8s.io/apimachinery/pkg/runtime"
+	"context"
+	"encoding/json"
 	"reflect"
+	"strings"
+
+	"gomodules.xyz/jsonpatch/v2"
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-func (r *Release) SetupWebhookWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(r).
-		Complete()
+const authorLabel = "release.rhtap.openshift.io/author"
+
+// releaseHandler describes the handler for the Release low level webhook
+type releaseHandler struct {
+	Client client.Client
 }
 
-//+kubebuilder:webhook:path=/validate-appstudio-redhat-com-v1alpha1-release,mutating=false,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releases,verbs=create;update,versions=v1alpha1,name=vrelease.kb.io,admissionReviewVersions=v1
+// +kubebuilder:webhook:path=/mutate-appstudio-redhat-com-v1alpha1-release,mutating=true,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releases,verbs=create;update,versions=v1alpha1,name=mrelease.kb.io,admissionReviewVersions=v1
 
-var _ webhook.Validator = &Release{}
-
-// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
-func (r *Release) ValidateCreate() error {
-	return nil
+// SetupReleaseLowLevelWebhook registers the Release low level webhook with the passed manager.
+func SetupReleaseLowlevelWebhook(mgr ctrl.Manager) {
+	mgr.GetWebhookServer().Register("/mutate-appstudio-redhat-com-v1alpha1-release", &webhook.Admission{Handler: &releaseHandler{Client: mgr.GetClient()}})
 }
 
-// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
-func (r *Release) ValidateUpdate(old runtime.Object) error {
-	if !reflect.DeepEqual(r.Spec, old.(*Release).Spec) {
-		return fmt.Errorf("release resources spec cannot be updated")
+// Handle takes an incoming admission request and returns an admission response. Create requests add
+// an author label with the current user. Update requests are rejected if the spec or author label
+// are being modified. All other requests are accepted without action.
+func (r *releaseHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	switch req.AdmissionRequest.Operation {
+	case admissionv1.Create:
+		return patchAuthorResponse(req)
+	case admissionv1.Update:
+		var releaseOld, releaseNew Release
+		err := json.Unmarshal(req.Object.Raw, &releaseNew)
+		if err != nil {
+			failedResponse(req.UID, "could not unmarshal the raw incoming Release object")
+		}
+		err = json.Unmarshal(req.OldObject.Raw, &releaseOld)
+		if err != nil {
+			failedResponse(req.UID, "could not unmarshal the raw old Release object")
+		}
+		if !reflect.DeepEqual(releaseNew.Spec, releaseOld.Spec) {
+			return failedResponse(req.UID, "release resources spec cannot be updated")
+		}
+		if releaseNew.GetLabels()[authorLabel] != releaseOld.GetLabels()[authorLabel] {
+			return failedResponse(req.UID, "release author label cannnot be updated")
+		}
 	}
-
-	return nil
+	return admission.Response{
+		AdmissionResponse: admissionv1.AdmissionResponse{
+			UID:     req.UID,
+			Allowed: true,
+			Result: &metav1.Status{
+				Status: "Success",
+			},
+		},
+	}
 }
 
-// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
-func (r *Release) ValidateDelete() error {
-	return nil
+// patchAuthorResponse returns a failed admission.Response that blocks the requested operation.
+func failedResponse(uid types.UID, message string) admission.Response {
+	return admission.Response{
+		AdmissionResponse: admissionv1.AdmissionResponse{
+			UID:     uid,
+			Allowed: false,
+			Result: &metav1.Status{
+				Status:  "Failure",
+				Message: message,
+			},
+		},
+	}
+}
+
+// patchAuthorResponse returns an admission.Response that modifies the CR author label to be
+// the user who issued the request.
+func patchAuthorResponse(req admission.Request) admission.Response {
+	author := strings.Replace(req.UserInfo.Username, ":", "_", -1)     // Colons disallowed in labels
+	author = strings.Replace(author, "system_serviceaccount", "sa", 1) // Effort to shorten long usernames
+	if len(author) > 63 {
+		author = string(author)[0:63] // Max length of label value is 63 chars
+	}
+	var patch []jsonpatch.JsonPatchOperation
+	patch = append(patch, jsonpatch.JsonPatchOperation{
+		Operation: "replace",
+		Path:      "/metadata/labels",
+		Value: map[string]string{
+			authorLabel: author,
+		},
+	})
+	patchBytes, err := json.Marshal(patch)
+	if err != nil {
+		return failedResponse(req.UID, "creation of JSON patch adding author label failed")
+	}
+	return admission.Response{
+		AdmissionResponse: admissionv1.AdmissionResponse{
+			UID:     req.UID,
+			Allowed: true,
+			Result: &metav1.Status{
+				Status: "Success",
+			},
+			Patch: patchBytes,
+			PatchType: func() *admissionv1.PatchType {
+				pt := admissionv1.PatchTypeJSONPatch
+				return &pt
+			}(),
+		},
+	}
 }
--- a/api/v1alpha1/release_webhook_test.go
+++ b/api/v1alpha1/release_webhook_test.go
@@ -16,27 +16,41 @@
 package v1alpha1
 
 import (
-	"context"
+	"encoding/json"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"gomodules.xyz/jsonpatch/v2"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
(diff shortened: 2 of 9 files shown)


api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,109 @@ limitations under the License.
...
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-func (r *Release) SetupWebhookWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(r).
-		Complete()
-}
+const authorLabel = "release.rhtap.openshift.io/author"
```
johnbieren, line 35: This will be moved once https://github.com/redhat-appstudio/release-service/pull/128 merges

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,109 @@ limitations under the License.
...
+			failedResponse(req.UID, "could not unmarshal the raw incoming Release object")
+		}
+		err = json.Unmarshal(req.OldObject.Raw, &releaseOld)
+		if err != nil {
+			failedResponse(req.UID, "could not unmarshal the raw old Release object")
+		}
+		if !reflect.DeepEqual(releaseNew.Spec, releaseOld.Spec) {
+			return failedResponse(req.UID, "release resources spec cannot be updated")
+		}
+		return patchAuthorResponse(req)
```
johnbieren, line 69: Open question here about if (as is currently written) any update to metadata is allowed, but you become the marked author OR any update to metadata is allowed EXCEPT updates to the author label, which would cause your request to be rejected

gitops/gitops_suite_test.go
```diff
@@ -85,16 +85,6 @@ var _ = BeforeSuite(func() {
 	})
 	Expect(err).NotTo(HaveOccurred())
 	Expect(k8sClient).NotTo(BeNil())
-
-	k8sManager, err := ctrl.NewManager(cfg, ctrl.Options{
```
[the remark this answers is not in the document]
johnbieren: David said I can remove them safely

api/v1alpha1/release_webhook.go
theflockers: I will have to rebase #128 so better you merge it then I can move this label to `metadata/labels.go` there.

api/v1alpha1/release_webhook.go
johnbieren: Discussed on slack. Resolving

api/v1alpha1/release_webhook.go
johnbieren: Whoever gets approved first :wink:

happybhati (approved):

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 )
 
-func (r *Release) SetupWebhookWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(r).
-		Complete()
-}
+const authorLabel = "release.rhtap.openshift.io/author"
```
davidmogar, line 35: I guess that Leandro's PR was not merged by the time you added this. Could you move it to the right place?

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+}
 
-var _ webhook.Validator = &Release{}
+// +kubebuilder:webhook:path=/mutate-appstudio-redhat-com-v1alpha1-release,mutating=true,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releases,verbs=create;update,versions=v1alpha1,name=mrelease.kb.io,admissionReviewVersions=v1
 
-// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
-func (r *Release) ValidateCreate() error {
-	return nil
+// SetupReleaseLowLevelWebhook registers the Release low level webhook with the passed manager.
+func SetupReleaseLowlevelWebhook(mgr ctrl.Manager) {
```
davidmogar, line 45: Can you rename it to `Register` and pass `mgr ctrl.Manager` and `log *logr.Logger`? Save the logger in the struct you defined so it can be used in the `Handle` function.

This, btw, it's going to cause problems if you create a new webhook with the same function name. Ideally I'd have every crd (and webhooks) into a different sub-package, but I didn't managed to do it yet.

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
-func (r *Release) SetupWebhookWithManager(mgr ctrl.Manager) error {
-	return ctrl.NewWebhookManagedBy(mgr).
-		For(r).
-		Complete()
-}
+const authorLabel = "release.rhtap.openshift.io/author"
 
-//+kubebuilder:webhook:path=/validate-appstudio-redhat-com-v1alpha1-release,mutating=false,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releases,verbs=create;update,versions=v1alpha1,name=vrelease.kb.io,admissionReviewVersions=v1
+// releaseHandler describes the handler for the Release low level webhook
+type releaseHandler struct {
```
davidmogar, line 38: Just call it `webhook` or something like that. It implements more things that only Handler.

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+			Result: &metav1.Status{
+				Status: "Success",
+			},
+		},
+	}
+}
 
-	return nil
+// patchAuthorResponse returns a failed admission.Response that blocks the requested operation.
+func failedResponse(uid types.UID, message string) admission.Response {
```
davidmogar, line 85: This has to be extracted to other place or it will collide with other webhooks redefining it. It can go to operator-goodies maybe... Let's leave it as it is and I'll come up with something.

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
-		return fmt.Errorf("release resources spec cannot be updated")
+// Handle takes an incoming admission request and returns an admission response. Create requests add
+// an author label with the current user. Update requests are rejected if the spec or author label
+// are being modified. All other requests are accepted without action.
+func (r *releaseHandler) Handle(ctx context.Context, req admission.Request) admission.Response {
+	switch req.AdmissionRequest.Operation {
+	case admissionv1.Create:
+		return patchAuthorResponse(req)
+	case admissionv1.Update:
+		var releaseOld, releaseNew Release
```
davidmogar, line 57: A function for this block is maybe not a bad idea...

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+			failedResponse(req.UID, "could not unmarshal the raw old Release object")
+		}
+		if !reflect.DeepEqual(releaseNew.Spec, releaseOld.Spec) {
+			return failedResponse(req.UID, "release resources spec cannot be updated")
+		}
+		if releaseNew.GetLabels()[authorLabel] != releaseOld.GetLabels()[authorLabel] {
+			return failedResponse(req.UID, "release author label cannnot be updated")
+		}
 	}
+	return admission.Response{
```
davidmogar, line 73: It could go into the switch's default case so it's a bit cleaner.

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+		},
+	}
 }
 
-// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
-func (r *Release) ValidateDelete() error {
-	return nil
+// patchAuthorResponse returns an admission.Response that modifies the CR author label to be
+// the user who issued the request.
+func patchAuthorResponse(req admission.Request) admission.Response {
```
davidmogar, line 100: Use a receiver here as well so we can add log lines if we want. Also, it should not be used outside this file that easily so yeah... receiver ;)

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+	author := strings.Replace(req.UserInfo.Username, ":", "_", -1)     // Colons disallowed in labels
+	author = strings.Replace(author, "system_serviceaccount", "sa", 1) // Effort to shorten long usernames
```
davidmogar, line 102: We should do this. Services accounts defined in appstudio should be renamed. We shouldn't rename them ourselves. What if there are two services accounts `system_serviceaccount_foo` and `sa_foo`? You would't never know which one created or modified the resource.

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+	if len(author) > 63 {
```
davidmogar, line 103: Constants for this and for the colon replacement please ;)

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+		author = string(author)[0:63] // Max length of label value is 63 chars
```
davidmogar, line 104: Use the new constant here and don't mention the size in the comment.

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+	}
+	var patch []jsonpatch.JsonPatchOperation
+	patch = append(patch, jsonpatch.JsonPatchOperation{
```
davidmogar, line 107: Could you use `PatchResponseFromRaw`? I think it would be cleaner. Check the test-operator webhook code.

api/v1alpha1/release_webhook_test.go
```diff
@@ -45,42 +59,246 @@ var _ = Describe("Release validation webhook", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(len(jsonPatch)).To(Equal(1))
+			patch := jsonPatch[0]
+			Expect(patch.Operation).To(Equal("replace"))
+			Expect(patch.Path).To(Equal("/metadata/labels"))
+			Expect(patch.Value).To(Equal(map[string]interface{}{
+				authorLabel: "admin",
+			}))
+		})
+		It("should overwrite the author label value when one is provided by user", func() {
```
davidmogar, line 85: We should have empty lines between It blocks. Probably you inherited this. We have tests that need to be rewrote yet.

api/v1alpha1/release_webhook_test.go
```diff
@@ -45,42 +59,246 @@ var _ = Describe("Release validation webhook", func() {
...
+			Expect(patch.Path).To(Equal("/metadata/labels"))
+			Expect(patch.Value).To(Equal(map[string]interface{}{
+				authorLabel: "admin",
+			}))
 		})
+	})
 
-		It("Should not error out when updating the resource metadata", func() {
-			ctx := context.Background()
+	Context("When a Release is updated", func() {
```
davidmogar, line 124: Empty line between contexts.

api/v1alpha1/release_webhook.go
```diff
@@ -17,38 +17,116 @@ limitations under the License.
...
+const authorLabel = "release.rhtap.openshift.io/author"
 
-//+kubebuilder:webhook:path=/validate-appstudio-redhat-com-v1alpha1-release,mutating=false,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releases,verbs=create;update,versions=v1alpha1,name=vrelease.kb.io,admissionReviewVersions=v1
+// releaseHandler describes the handler for the Release low level webhook
+type releaseHandler struct {
+	Client client.Client
+}
 
-var _ webhook.Validator = &Release{}
+// +kubebuilder:webhook:path=/mutate-appstudio-redhat-com-v1alpha1-release,mutating=true,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releases,verbs=create;update,versions=v1alpha1,name=mrelease.kb.io,admissionReviewVersions=v1
```
davidmogar, line 42: It probably can be used for ReleasePlans as well.

api/v1alpha1/release_webhook.go
johnbieren: I merged first so he moved it as part of his pr (which is now merged)

api/v1alpha1/release_webhook.go
johnbieren: I don't necessarily disagree, but 1) I don't think that is something to hold this (or the next) PR up on and 2) I think will take a while and be a program decision. I would expect someone has an argument for keeping the service account names as is

api/v1alpha1/release_webhook.go
johnbieren: done in followup

api/v1alpha1/release_webhook_test.go
johnbieren: done in follow up

api/v1alpha1/release_webhook.go
johnbieren: done in follow up

api/v1alpha1/release_webhook.go
johnbieren: done in follow up

api/v1alpha1/release_webhook.go
johnbieren: let's revisit this idea when you see new version

api/v1alpha1/release_webhook.go
johnbieren: fixed in new version

api/v1alpha1/release_webhook.go
johnbieren: fixed in new version

api/v1alpha1/release_webhook.go
johnbieren: can revisit how things are in the new version

api/v1alpha1/release_webhook.go
johnbieren: can revisit in new version

api/v1alpha1/release_webhook.go
johnbieren: done in followup

api/v1alpha1/release_webhook.go
davidmogar: Agree, but I would just remove that line. Just trim it.

PR #187, opened 2023-05-03
chore: fix ordering

davidmogar:
In this house we follow the rules of ABC!


johnbieren (approved):
someone get this man an IDE that just finds the function for you so abc doesn't matter

mmalina:
@davidmogar I'm wondering, is this a convention in Go language? I've never heard of it in Python world - typically I will order functions based on the order they are called or something similar.

PR #188, opened 2023-05-04
feat(HACBS-2129): add release reference as a Pipeline parameter

davidmogar:
This change adds a new function to declare new Pipeline parameters. This function allows to pass any number of client.Objects. The kind of the object will be used as the parameter name while the value will contain the namespaced name of the object.

In addition, every new ReleasePipelineRun will contain a reference to the Release.


davidmogar:
@johnbieren ping :wink:

johnbieren:
@davidmogar we can't actually use this stuff unless our service account we run the pipeline with has rbac permissions to the resource being passed. Do you know how we do that?

PR #189, opened 2023-05-04
fix: few small issues in release webhook

johnbieren:


api/v1alpha1/release_webhook.go
```diff
@@ -56,11 +56,11 @@ func (r *releaseHandler) Handle(ctx context.Context, req admission.Request) admi
 		var releaseOld, releaseNew Release
 		err := json.Unmarshal(req.Object.Raw, &releaseNew)
 		if err != nil {
-			failedResponse(req.UID, "could not unmarshal the raw incoming Release object")
+			return failedResponse(req.UID, "could not unmarshal the raw incoming Release object")
```
theflockers, line 59: I guess we can´t mimic the Handle in a way we can make codecov happy here , right?

johnbieren: I'm not sure that I can create a struct that wouldn't marshal to pass to the json.unmarshal error and cause an error, so I don't think so

davidmogar (approved):

PR #190, opened 2023-05-09
feat(HACBS-2126): add author webhook

johnbieren:


api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const MAX_LABEL_LENGTH = 63
+
+// authorWebhook describes the data structure for the author webhook
+type authorWebhook struct {
+	Client client.Client
```
davidmogar, line 39: Don't export it.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+		var releaseOld Release
+		err = json.Unmarshal(req.OldObject.Raw, &releaseOld)
+		if err != nil {
+			return admission.Errored(http.StatusInternalServerError, errors.Wrap(err, "error decoding object"))
+		}
+		if releaseNew.GetLabels()[metadata.AuthorLabel] != releaseOld.GetLabels()[metadata.AuthorLabel] {
+			return admission.Errored(http.StatusInternalServerError, errors.New("release author label cannnot be updated"))
+		}
+	}
+	return admission.Allowed("Success")
```
davidmogar, line 89: Add it to the switch default.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+// handleRelease takes an incoming admission request and returns an admission response. Create requests
+// add an author label with the current user. Update requests are rejected if the spec or author label
+// are being modified. All other requests are accepted without action.
+func (a *authorWebhook) handleRelease(req admission.Request) admission.Response {
+	var releaseNew Release
+	err := json.Unmarshal(req.Object.Raw, &releaseNew)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, errors.Wrap(err, "error decoding object"))
+	}
+	switch req.AdmissionRequest.Operation {
```
davidmogar, line 76: Can we add an empty line before? Too packed.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+	default:
+		return admission.Errored(http.StatusInternalServerError, errors.New("webhook tried to handle an unsupported resource: "+string(req.Kind.Kind)))
+	}
+}
+
+// handleRelease takes an incoming admission request and returns an admission response. Create requests
+// add an author label with the current user. Update requests are rejected if the spec or author label
+// are being modified. All other requests are accepted without action.
+func (a *authorWebhook) handleRelease(req admission.Request) admission.Response {
+	var releaseNew Release
```
davidmogar, line 71: Just call it `release`. Not a new release, just the one you are handling.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+	err := json.Unmarshal(req.Object.Raw, &releaseNew)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, errors.Wrap(err, "error decoding object"))
+	}
+	switch req.AdmissionRequest.Operation {
+	case admissionv1.Create:
+		return a.patchAuthorResponse(req.UserInfo.Username, &releaseNew)
+	case admissionv1.Update:
+		var releaseOld Release
+		err = json.Unmarshal(req.OldObject.Raw, &releaseOld)
```
davidmogar, line 81: You don't need this, isn't it? You already have it at the beginning of this function.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+
+	newObj, err := json.Marshal(obj)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, errors.Wrap(err, "error encoding response object with author added"))
+	}
+
+	return admission.PatchResponseFromRaw(oldObj, newObj)
+}
+
+// removeAuthorResponse returns an admission.Response that removes the author label from the CR.
```
davidmogar, line 173: I'd change it a bit. Main goal of the function is to remove the label. Returning a response is a side effect.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+	newObj, err := json.Marshal(obj)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, errors.Wrap(err, "error encoding response object with author removed"))
+	}
+
+	return admission.PatchResponseFromRaw(oldObj, newObj)
+}
+
+// authorLabelValue takes a username and returns it in a form appropriate to use as a label value.
+func authorLabelValue(username string) string {
```
davidmogar, line 191: Add a receiver here as well.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+	"github.com/pkg/errors"
+	"github.com/redhat-appstudio/release-service/metadata"
+	admissionv1 "k8s.io/api/admission/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+const MAX_LABEL_LENGTH = 63
```
davidmogar, line 35: Too Pythonic. Rename to `maxLabelLength` as it doesn't need to be exported. It could go also into `metadata/labels.go` and be exported. It can be used in other parts.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+// authorWebhook describes the data structure for the author webhook
+type authorWebhook struct {
+	Client client.Client
+	log    *logr.Logger
+}
+
+// +kubebuilder:webhook:path=/mutate-appstudio-redhat-com-v1alpha1-author,mutating=true,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=releases;releaseplans,verbs=create;update,versions=v1alpha1,name=mauthor.kb.io,admissionReviewVersions=v1
+
+// RegisterAuthorWebhook registers the author webhook with the passed manager and log.
+func RegisterAuthorWebhook(mgr ctrl.Manager, log logr.Logger) {
```
davidmogar, line 46: Better pass a logger reference already.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+			if author != "" && author != authorLabelValue(req.UserInfo.Username) {
+				return admission.Errored(http.StatusInternalServerError, errors.New("[labels] cannot set attribution to true if not setting"+
+					" current user as author. Expected author is: "+authorLabelValue(req.UserInfo.Username)))
+			}
+			return a.patchAuthorResponse(req.UserInfo.Username, &releasePlanNew)
+		}
+		return admission.Allowed("Success")
+	case admissionv1.Update:
+		var releasePlanOld ReleasePlan
+		err = json.Unmarshal(req.OldObject.Raw, &releasePlanOld)
```
davidmogar, line 114: This is exactly the same as the one at the beginning of the function.

api/v1alpha1/author_webhook.go
davidmogar: It also needs a more descriptive name. Just call it `parseLabelValue` or whatever... Something that describes what it does.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	return admission.PatchResponseFromRaw(oldObj, newObj)
+}
+
+// removeAuthorResponse returns an admission.Response that removes the author label from the CR.
+func (a *authorWebhook) removeAuthorResponse(obj client.Object) admission.Response {
+	oldObj, err := json.Marshal(obj)
+	if err != nil {
+		return admission.Errored(http.StatusInternalServerError, errors.Wrap(err, "error encoding response object"))
```
davidmogar, line 177: What for? Why not just try to delete the label and marshal?

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+		return a.patchAuthorResponse(req.UserInfo.Username, &releasePlanNew)
+	default:
+		return admission.Allowed("Success")
+	}
+}
+
+// patchAuthorResponse returns an admission.Response that modifies the CR author label to be
+// the user provided.
+func (a *authorWebhook) patchAuthorResponse(username string, obj client.Object) admission.Response {
+	oldObj, err := json.Marshal(obj)
```
davidmogar, line 153: What for? Try to do the operation and then check if you can marshal it.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,197 @@
...
+	mgr.GetWebhookServer().Register("/mutate-appstudio-redhat-com-v1alpha1-author", &webhook.Admission{
+		Handler: &authorWebhook{
+			Client: mgr.GetClient(),
+			log:    &log,
+		},
+	})
+}
+
+// Handle creates an admission response for Release and ReleasePlan requests.
+func (a *authorWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
```
davidmogar, line 56: Maybe this function could do something like this:
```
func (a *authorWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
        var object client.Object
	err := a.decoder.Decode(req, object)
	if err != nil {
		return admission.Errored(http.StatusInternalServerError, errors.Wrap(err, "error decoding request"))
	}

	switch req.Kind.Kind {
	case "Release":
		return a.handleRelease(req, object.(*v1alpha1.Release))
	case "ReleasePlan":
		return a.handleReleasePlan(req, object.(*v1alpha1.ReleasePlan))
	}
}
```
That way you already have the `req` with the `req.Object.Raw` and the object itself.

api/v1alpha1/author_webhook.go
johnbieren: One is the object, one is the oldobject

api/v1alpha1/author_webhook.go
johnbieren: I can't, the IDE complains about missing a return

api/v1alpha1/author_webhook.go
johnbieren: No it is not. One is the Object and one is the OldObject

api/v1alpha1/author_webhook.go
johnbieren: You want an entire function just for `delete(obj.GetLabels(), metadata.AuthorLabel)`? The response is what removes it, as it returns a patch response

api/v1alpha1/author_webhook.go
davidmogar: Weird! I tested this because I wasn't sure and I didn't get any complain. I checked again and it does complain. I guess the repo I used was kinda broken and it wasn't really checking. My bad.

P.S. An empty line wouldn't kill you :wink:

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// handleRelease takes an incoming admission request and returns an admission response. Create requests
+// add an author label with the current user. Update requests are rejected if the author label is being
+// modified. All other requests are accepted without action.
+func (a *authorWebhook) handleRelease(req admission.Request) admission.Response {
+	release := &Release{}
+	err := json.Unmarshal(req.Object.Raw, release)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, errors.Wrap(err, "error decoding object"))
```
davidmogar, line 74: I don't like this error string dancing around everywhere, but you know that already. Go ahead and dismiss my comment :stuck_out_tongue_winking_eye:

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,167 @@
...
+// during an update and the author value is not modified, the previous author label remains.
+func (a *authorWebhook) handleReleasePlan(req admission.Request) admission.Response {
+	releasePlan := &ReleasePlan{}
+	err := json.Unmarshal(req.Object.Raw, releasePlan)
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, errors.Wrap(err, "error decoding object"))
+	}
+	// Author label should not exist in any case if attribution is not true
+	if releasePlan.GetLabels()[metadata.AttributionLabel] != "true" {
+		delete(releasePlan.GetLabels(), metadata.AuthorLabel)
```
davidmogar, line 108: I think that having a function with a proper name wrapping this line is more readable and maintainable. But you already know that as well.

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,167 @@
...
+			return admission.Errored(http.StatusBadRequest, errors.Wrap(err, "error decoding object"))
+		}
+
+		if releasePlan.GetLabels()[metadata.AttributionLabel] == "true" {
+			author := releasePlan.GetLabels()[metadata.AuthorLabel]
+
+			if oldReleasePlan.GetLabels()[metadata.AttributionLabel] != "true" || author == a.sanitizeLabelValue(req.UserInfo.Username) {
+				a.setAuthorLabel(req.UserInfo.Username, releasePlan)
+			} else {
+				// Preserve previous author if update does not involve the author label
```
davidmogar, line 129: This, at least for me, is so difficult to understand. What about:
```suggestion
				// Preserve previous author if the new author does not match the user making the change
```

api/v1alpha1/author_webhook.go
```diff
@@ -0,0 +1,167 @@
...
+				a.setAuthorLabel(oldReleasePlan.GetLabels()[metadata.AuthorLabel], releasePlan)
+			}
+		}
+	}
+	return a.patchResponse(req.Object.Raw, releasePlan)
```
davidmogar, line 134: Make me happy and add an empty line before, please.

johnbieren: I am about to just change my code to have the double spacing alignment :wink:

davidmogar (approved):

PR #192, opened 2023-05-10
feat(HACBS-2127): add automated to release status

johnbieren:
@davidmogar not sure if you want me to create a `r.markAutomated` function that sets the status or just set the status directly in integration-service. Figured you'd have an opinion on it so only added the field until I hear it

### DIFF
--- a/api/v1alpha1/release_types.go
+++ b/api/v1alpha1/release_types.go
@@ -66,6 +66,9 @@ type ReleaseStatus struct {
 	// +optional
 	Target string `json:"target,omitempty"`
 
+	// Automated indicates whether the Release was created as part of an automated process or manually by an end-user
+	Automated bool `json:"automated,omitempty"`
+
 	// CompletionTime is the time when a Release was completed
 	// +optional
 	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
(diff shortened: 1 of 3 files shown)


api/v1alpha1/release_types.go
```diff
@@ -130,6 +130,9 @@ type ProcessingInfo struct {
 
 // ValidationInfo defines the observed state of the release validation.
 type ValidationInfo struct {
+	// Automated indicates whether the Release was created by the integration service
+	Automated bool `json:"automated,omitempty"`
```
davidmogar, line 134: How is this related to validation info? It shouldn't be here. Instead, this should probably be a top level field.

api/v1alpha1/release_types.go
```diff
@@ -130,6 +130,9 @@ type ProcessingInfo struct {
 
 // ValidationInfo defines the observed state of the release validation.
 type ValidationInfo struct {
+	// Automated indicates whether the Release was created by the integration service
```
davidmogar, line 133: I don't really like to mention services in the comments in case the name changes or those services get split (e.g. has and build services). What about:
```suggestion
	// Automated indicates whether the Release was created as part of an automated process or manually by an end-user
```

api/v1alpha1/release_types.go
johnbieren: Ack

davidmogar (approved):
Answering to your question, the `Automated`  field can be modified directly from the operation that will process the integration label.

davidmogar (approved):

PR #194, opened 2023-05-15
feat(HACBS-2127): set release attribution data in controller

johnbieren:
This commit modifies the validation operation that is run as part of the reconcile loop. Part of the validation will now include ensuring that a proper user is provided for attribution and setting that value in the release status.


controllers/release/adapter.go
```diff
@@ -478,3 +490,42 @@ func (a *Adapter) syncResources() error {
...
+		if author == "" { // this should be impossible
+			return fmt.Errorf("release is manual and has no author")
+		}
+	}
+	a.release.Status.Attribution.Author = author
+
+	return a.client.Status().Patch(a.ctx, a.release, patch)
+}
+
+// validateAuthor attributes the release to a specific user and ensures that the user is valid in SSO.
```
mmalina, line 523: Where is the SSO check done? I cannot see it.

api/v1alpha1/release_types.go
```diff
@@ -78,6 +83,21 @@ type ReleaseStatus struct {
...
+	// +optional
+	Author string `json:"author,omitempty"`
+
+	// StandingAuthorization indicates whether the release is attributed through a ReleasePlan
+	// +optional
+	StandingAuthorization bool `json:"standingAuthorization,omitempty"`
+
+	// Validated specifies whether or not the author has been confirmed to be a valid user
+	// +optional
+	Validated string `json:"validated,omitempty"`
```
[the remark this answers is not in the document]
johnbieren: It was just in preparation for the SSO stuff. I will remove it here and add it back in that PR

controllers/release/adapter.go
```diff
@@ -478,3 +490,42 @@ func (a *Adapter) syncResources() error {
...
+func (a *Adapter) trackReleaseAttribution(releasePlan *v1alpha1.ReleasePlan) error {
+	if a.release.IsAttributed() {
+		return nil
+	}
+
+	var author string
+	patch := client.MergeFrom(a.release.DeepCopy())
+
+	if a.release.Labels[metadata.AutomatedLabel] == "true" {
+		if !a.release.IsAutomated() {
```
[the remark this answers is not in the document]
johnbieren: This function is about properly setting the Attribution.Author field in the release status. Why I duplicate the retrieving of the automated label in both registerAttributionData and validateAuthor instead of just do it all here?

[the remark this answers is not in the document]
davidmogar: Because you are mixing the logic. Also, I'm not saying that you have to duplicate the check. A function to set the author should not care about raising an error which is used to decide how the reconciliation should continue. As we discussed offline, one function to validate and another one to register the data.

johnbieren:
Depends on https://github.com/redhat-appstudio/e2e-tests/pull/546

api/v1alpha1/author_webhook_test.go
```diff
@@ -116,6 +116,36 @@ var _ = Describe("Author webhook", Ordered, func() {
...
+				}
+				admissionRequest.Object.Raw, err = json.Marshal(release)
+				Expect(err).NotTo(HaveOccurred())
+
+				rsp := authWebhook.Handle(ctx, admissionRequest)
+				Expect(rsp.AdmissionResponse.Allowed).To(BeTrue())
+				Expect(len(rsp.Patches)).To(Equal(1))
+				patch := rsp.Patches[0]
+				Expect(patch.Operation).To(Equal("add"))
+				// The json functions replace `/` so checking the entire value does not work
```
theflockers, line 145: I would remove this line. The same comment exists in the test above but not in the first test in the context. Maybe both are unnecessary.

controllers/release/adapter.go
```diff
@@ -478,3 +492,43 @@ func (a *Adapter) syncResources() error {
...
+
+	if a.release.Labels[metadata.AutomatedLabel] == "true" {
+		author = releasePlan.Labels[metadata.AuthorLabel]
+		if author == "" {
+			return fmt.Errorf("no author in the ReleasePlan found for automated release")
+		}
+		a.release.Status.Attribution.StandingAuthorization = true
+	} else {
+		author = a.release.Labels[metadata.AuthorLabel]
+		if author == "" { // this should be impossible
```
theflockers, line 513: This should be or it is actually impossible to match at the moment?  When I was writing tests there was cases in the adapter code (maybe they'd gone after dozen refactorings) that could not never happen, although no comments at all. So maybe this comment is also unnecessary.

theflockers (commented):
please check the comments.

happybhati (dismissed):
LGTM 👍  
Few refinement comments from Leandro.

controllers/release/adapter.go
johnbieren: It should be impossible because webhooks prevent it from happening. However, in instances such as e2e or place where webhooks do not run, it could happen, so I figured it is probably best to check for it

controllers/release/adapter.go
```diff
@@ -478,3 +492,43 @@ func (a *Adapter) syncResources() error {
...
+	}
+	a.release.Status.Attribution.Author = author
+
+	return a.client.Status().Patch(a.ctx, a.release, patch)
+}
+
+// validateAuthor attributes the release to a specific user and ensures that the user is valid in SSO.
+func (a *Adapter) validateAuthor() error {
+	if a.release.Labels[metadata.AutomatedLabel] == "true" && !a.release.IsAutomated() {
+		return fmt.Errorf("automated not set in status for automated release")
```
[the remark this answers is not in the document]
johnbieren: Well, the automated label *is* set in this case. Automated is not set in the status though

api/v1alpha1/author_webhook_test.go
johnbieren: I put the comment to explain why we are comparing substrings instead of the entire string. The comment is not there when we are comparing the entire string because it wouldn't make sense in that case... the comment is about why are are using containsSubstring instead of Equals

theflockers (dismissed):

controllers/release/adapter.go
```diff
@@ -260,6 +263,17 @@ func (a *Adapter) EnsureReleaseIsValid() (reconciler.OperationResult, error) {
...
 	}
 
+	err = a.validateAuthor()
+	if err != nil {
+		if strings.Contains(err.Error(), "automated not set in status") {
+			return reconciler.RequeueWithError(err)
+		}
+
+		a.release.MarkValidationFailed(err.Error())
+		a.release.MarkReleaseFailed("Release validation failed")
```
davidmogar, line 273:
```suggestion
		a.release.MarkReleaseFailed("Author validation failed")
```

controllers/release/adapter.go
davidmogar, line 513: I would rewrite the comment. As you said it could happen. Just say something like `webhooks prevent this to happen but they could be disabled in some scenarios`. Or something like that.

controllers/release/adapter.go
```diff
@@ -478,3 +492,43 @@ func (a *Adapter) syncResources() error {
...
+			return fmt.Errorf("no author in the ReleasePlan found for automated release")
+		}
+		a.release.Status.Attribution.StandingAuthorization = true
+	} else {
+		author = a.release.Labels[metadata.AuthorLabel]
+		if author == "" { // this should be impossible
+			return fmt.Errorf("no author found for manual release")
+		}
+	}
+	a.release.Status.Attribution.Author = author
```
davidmogar, line 517: I think I commented this already, but you don't need to declare an `author` variable. Just use `a.release.Status.Attribution.Author` directly. One less assignment as you don't need this line. If you really want to go this way, add an empty line before this one please.

davidmogar (changes requested):

controllers/release/adapter.go
johnbieren: Not sure I follow. I set `author` in the if statement based on conditions, then set the value in the status to whatever I found in the if statement. I can add an empty line though

controllers/release/adapter.go
johnbieren: done

controllers/release/adapter.go
johnbieren: done

controllers/release/adapter.go
davidmogar: I was just saying that you don need the `author` variable in this block. That you can use the target directly (`a.release.Status.Attribution.Author`) but I'm fine with this.

davidmogar (approved):

happybhati (approved):

PR #202, opened 2023-06-06
feat(HACBS-2224): add support for extraData in ReleasePlanAdmission

mmalina:
extraData is an optional new field in the .spec of ReleasePlanAdmission. Its value is passed to the release pipeline as a parameter.


tekton/pipeline_run.go
```diff
@@ -159,6 +160,21 @@ func (r *ReleasePipelineRun) WithReleaseStrategy(strategy *v1alpha1.ReleaseStrat
 	return r
 }
 
+// WithReleasePlanAdmission adds extraData to the release PipelineRun.
+func (r *ReleasePipelineRun) WithReleasePlanAdmission(planAdmission *v1alpha1.ReleasePlanAdmission) *ReleasePipelineRun {
+	// We ignore the error here because none should be raised when marshalling the spec of a CRD.
+	// If we end up deciding it is useful, we will need to pass the errors trough the chain and
+	// add something like a `Complete` function that returns the final object and error.
+	extraDataString, _ := json.Marshal(planAdmission.Spec.ExtraData)
```
davidmogar, line 168: We are not going to pass the extra data as a parameter as it could be too big (not the parameter itself but the total size of all parameters). We said that we are going to pass a reference to the resource and load it later on in the Tekton Pipeline. For that, we already have a call (`WithObjectReferences`) to which you can add more resources (it accepts a list of resources).

mmalina: Thanks, fixed, please check now. Also, I'm not sure about the `ReleasePlanAdmission` object that I create in the test. Maybe it doesn't even need the extraData field? To be honest, I'm not sure if I populate the field correctly, because when I was passing it as a parameter and wrote a test for it, it seemed like the string was empty.

tekton/pipeline_run.go
```diff
@@ -159,6 +160,13 @@ func (r *ReleasePipelineRun) WithReleaseStrategy(strategy *v1alpha1.ReleaseStrat
 	return r
 }
 
+// WithReleasePlanAdmission a reference to the ReleasePlanAdmission object to the pipeline
+func (r *ReleasePipelineRun) WithReleasePlanAdmission(planAdmission *v1alpha1.ReleasePlanAdmission) *ReleasePipelineRun {
+	r.WithObjectReferences(planAdmission)
```
davidmogar, line 165: Nope. Just remove this whole method and use the existing call in `createReleasePipelineRun`.  Something like `WithObjectReferences(a.release, releasePlanAdmission)`.

tekton/pipeline_run_test.go
```diff
@@ -20,11 +20,13 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"k8s.io/apimachinery/pkg/types"
 	"os"
 	"reflect"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/runtime"
```
davidmogar, line 27: Remove all the changes in this file. No test is needed as you will be adding no new functionality.

davidmogar (changes requested):

tekton/pipeline_run.go
davidmogar: Thing is that this PR is mixing two things:

- Support for the extraData field.
- Sending the ReleasePlanAdmission reference to the Pipeline.

I think those should be two different PRs as they are just slightly related. As alternative, this could be done in two different commits. I'd probably go with the former.

mmalina: OK, I removed the second part from this PR and will create a new PR once this is merged. Please check now.

tekton/pipeline_run.go
mmalina: OK, I fixed it, it will be included in the second PR.

tekton/pipeline_run_test.go
mmalina: ok, done, will be part of the second PR (or, in this case, it _will not be_ part of it ;) ).

davidmogar (approved):
All good. Don't forget to update the PR description.

PR #203, opened 2023-06-07
feat(HACBS-1724): add govulncheck pr check

mmalina:
This will run on each PR and check for vulnerabilities in go dependencies.


mmalina:
We agreed with the team that we don't need this extra check for now, since we already have the dependabot alerts from Github enabled and we will further improve this in [HACBS-2273](https://issues.redhat.com/browse/HACBS-2273)

PR #216, opened 2023-07-20
fix: move dependabot file to proper spot

johnbieren:
Per the documentation, the dependabot.yml file should be in the .github directory. With it in its current spot, I am getting workflow failure emails.


mmalina (approved):
Sounds good. For reference, here's the documentation: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file

PR #226, opened 2023-07-30
fix: Update GrafanaDashboard to match GrafanaOperator v5 (RHTAPWATCH-8)

avi-biton:
as part of this bug fix and the upgrade to `GrafanaOperator` v5, we have to update the `GrafanaDashboard` resource to match the new api

More details:
https://redhat-internal.slack.com/archives/C02CTEB3MMF/p1690371680784569

The dashboard is still presenting data due to a patch we merged to RHTAP, and it will be removed after these changes will be merged


avi-biton:
@davidmogar Can you please review and merge?

davidmogar (approved):

PR #229, opened 2023-08-02
feat(HACBS-2427): update network policy to add additional port to CIDR

happybhati:
This commit will add a additional kubernates CIDR
to allow traffic to as per the requirement from ROSA cluster.

### DIFF
--- a/config/manager/network_policy.yaml
+++ b/config/manager/network_policy.yaml
@@ -20,7 +20,9 @@ spec:
           protocol: TCP
     - to:
       - ipBlock:
-          cidr: 172.20.0.1/32
+          cidr: 172.20.0.1/32  # CIDR of Kubernetes API Endpoint.
       ports:
         - port: 6443
+          protocol: TCP
+        - port: 443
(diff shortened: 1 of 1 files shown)


config/manager/network_policy.yaml
```diff
@@ -19,8 +19,12 @@ spec:
         - port: 6443
           protocol: TCP
     - to:
+      - ipBlock:
```
davidmogar, line 22: Duplicated?

davidmogar (dismissed):

mmalina:
@happybhati can you improve the commit message?

> This commit will add a additional kubernates CIDR to allow traffic to

AFAICS, CIDR refers to an ip range in the form of up/mask, e.g. `172.20.0.1/32` that we have there. This commit does not change the ip range, it just allows an additional port for that range, right?

kasemAlem (approved):

mmalina (dismissed):

scoheb (approved):

PR #231, opened 2023-08-02
chore(ci): add workaround for new osp ver

scoheb:
- Testing workaround that specifies kind=Pipeline for PipelineRuns


happybhati (approved):
Thanks for the fix @scoheb

PR #233, opened 2023-08-09
fix(RHTAPBUGS-580): correct releaseplan controller rbac

johnbieren:


johnbieren:
See https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_controller.go#L72 which shows not to use camel case for CRD names in rbac

scoheb (approved):

PR #236, opened 2023-08-16
feat(HACBS-2422): add finalizer to Release PRs

davidmogar:
This change adds a finalizer to Release PipelineRuns to avoid their deletion by external entities until the Release service has finished processing them.

Notice that tests relying on finalizeRelease don't check if the PipelineRun was successfully deleted anymore. The removal of the finalizer is not instant so to test that, we would have to use Eventually blocks, which we know are flaky.

### DIFF
--- a/controllers/release/adapter.go
+++ b/controllers/release/adapter.go
@@ -54,9 +54,6 @@ type adapter struct {
 	syncer  *syncer.Syncer
 }
 
-// finalizerName is the finalizer name to be added to the Releases
-const finalizerName string = "appstudio.redhat.com/release-finalizer"
-
 // newAdapter creates and returns an adapter instance.
 func newAdapter(ctx context.Context, client client.Client, release *v1alpha1.Release, loader loader.ObjectLoader, logger *logr.Logger) *adapter {
 	return &adapter{
@@ -79,13 +76,13 @@ func (a *adapter) EnsureFinalizersAreCalled() (controller.OperationResult, error
 		return controller.ContinueProcessing()
 	}
 
-	if controllerutil.ContainsFinalizer(a.release, finalizerName) {
+	if controllerutil.ContainsFinalizer(a.release, metadata.ReleaseFinalizer) {
 		if err := a.finalizeRelease(); err != nil {
 			return controller.RequeueWithError(err)
 		}
 
 		patch := client.MergeFrom(a.release.DeepCopy())
-		controllerutil.RemoveFinalizer(a.release, finalizerName)
+		controllerutil.RemoveFinalizer(a.release, metadata.ReleaseFinalizer)
 		err := a.client.Patch(a.ctx, a.release, patch)
 		if err != nil {
 			return controller.RequeueWithError(err)
@@ -100,7 +97,7 @@ func (a *adapter) EnsureFinalizersAreCalled() (controller.OperationResult, error
 func (a *adapter) EnsureFinalizerIsAdded() (controller.OperationResult, error) {
 	var finalizerFound bool
 	for _, finalizer := range a.release.GetFinalizers() {
-		if finalizer == finalizerName {
+		if finalizer == metadata.ReleaseFinalizer {
 			finalizerFound = true
 		}
 	}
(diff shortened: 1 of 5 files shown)


tekton/pipeline_run.go
```diff
@@ -123,9 +124,10 @@ func (r *ReleasePipelineRun) WithObjectReferences(objects ...client.Object) *Rel
 	return r
 }
 
-// WithOwner set's owner annotations to the release PipelineRun.
+// WithOwner set's owner annotations to the release PipelineRun and a finalizer to prevent its deletion.
```
mmalina, line 127: Nit: `s/set's/sets/`

Also, wouldn't it be nicer to have a separate `WithFinalizer` function for this?

johnbieren: Agree with `s/set's/sets/`. I think keeping it all in one function is okay though. I kind of think the finalizer concept fits under the English meaning of ownership. Fine with it being split up too though

johnbieren (dismissed):

mmalina (approved):

controllers/release/adapter_test.go
```diff
@@ -1191,6 +1219,7 @@ var _ = Describe("Release adapter", Ordered, func() {
 			Expect(adapter.release.HasProcessingFinished()).To(BeTrue())
 			Expect(adapter.release.IsProcessed()).To(BeFalse())
 		})
+
```
johnbieren, line 1222: why this extra space?

controllers/release/adapter.go
```diff
@@ -290,7 +287,20 @@ func (a *adapter) EnsureReleaseProcessingIsTracked() (controller.OperationResult
...
 	}
 	if pipelineRun != nil {
-		return controller.RequeueOnErrorOrContinue(a.registerProcessingStatus(pipelineRun))
+		err = a.registerProcessingStatus(pipelineRun)
+		if err != nil {
+			return controller.RequeueWithError(err)
+		}
+	}
+
+	// This condition can only be true if the call to registerProccessingStatus changed the state
```
johnbieren, line 296: 1 c in processing

controllers/release/adapter.go
```diff
@@ -374,6 +384,19 @@ func (a *adapter) finalizeRelease() error {
 	}
 
 	if pipelineRun != nil {
+		// The finalizer could still exist at this point. It should be removed if the Release is finalized
```
johnbieren, line 387: I'm a little confused when we would expect this to happen. Shouldn't the EnsureReleaseProcessingIsTracked operation always remove it?

controllers/release/adapter_test.go
```diff
@@ -782,6 +779,37 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       "pipeline-run",
+					Namespace:  "default",
+					Finalizers: []string{metadata.ReleaseFinalizer},
+				},
+			}
+			// The resource needs to be created as it will get patched
+			Expect(adapter.client.Create(adapter.ctx, pipelineRun)).To(Succeed())
+
+			pipelineRun.Status.MarkSucceeded("", "")
```
johnbieren, line 795: if it is getting patched shouldn't this happen first? As in, before line 793?

controllers/release/adapter_test.go
```diff
@@ -782,6 +779,37 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+				},
+			})
+
+			result, err := adapter.EnsureReleaseProcessingIsTracked()
+			Expect(!result.RequeueRequest && !result.CancelRequest).To(BeTrue())
+			Expect(err).NotTo(HaveOccurred())
+			Expect(pipelineRun.Finalizers).To(BeEmpty())
+
+			// Clean up at the end
+			Expect(adapter.client.Delete(adapter.ctx, pipelineRun)).To(Succeed())
```
johnbieren, line 810: if the test fails once on lines 804-807, this won't be run and next time the test runs the pipeline will already exist. Are we okay with that?

controllers/release/adapter.go
davidmogar: Not if we didn't get the status update or if the operation is disabled.

controllers/release/adapter_test.go
davidmogar: Nope. That only changes the status and a patch is not needed as we mock the return of the loader function and we don't rely on this status later on.

controllers/release/adapter_test.go
davidmogar: We get a new environment every single time we run. Everything is gone by that time. If not, that could be the case for anything in the `createResources` function.

controllers/release/adapter.go
johnbieren: Well, the operation isn't disabled. I am not sure why we would ever miss the status update. Nonetheless, can you at least put that in the comment (what your reply just was)?

controllers/release/adapter_test.go
johnbieren: so this is just so the HasProcessingFinished() statement hits?

davidmogar: Yes.

PR #238, opened 2023-08-17
fix(RHTAPBUGS-714): aggregate clusterroles

ralphbean:
This change should cause these roles to be aggregated into the system wide view, edit, and admin clusterroles.

Notably, it will also cause the roles to be aggregated into the dedicated-admin role, which is what I'm really after.

### DIFF
--- a/config/rbac/release_editor_role.yaml
+++ b/config/rbac/release_editor_role.yaml
@@ -3,6 +3,9 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
   name: release-editor-role
+  labels:
+    rbac.authorization.k8s.io/aggregate-to-admin: 'true'
+    rbac.authorization.k8s.io/aggregate-to-edit: 'true'
 rules:
 - apiGroups:
   - appstudio.redhat.com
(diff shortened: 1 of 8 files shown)


ralphbean:
Replaces https://github.com/redhat-appstudio/release-service/pull/237

mmalina:
@ralphbean thanks for the PR. I'll let David (off today) and Johnny review. But in the meantime, can you sign your commit? Here's how to do it in case you haven't done it before: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

davidmogar (approved):

johnbieren:
@ralphbean can you rebase this? We have our main branch set to require the commits to be rebased in order to merge

johnbieren:
> @ralphbean thanks for the PR. I'll let David (off today) and Johnny review. But in the meantime, can you sign your commit? Here's how to do it in case you haven't done it before: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
> 
> Edit: Now that I've read up on [Aggregated ClusterRoles](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#aggregated-clusterroles) I agree with this change. But I'd still like David to review as well.

@ralphbean still need the commit signed then I can merge

PR #246, opened 2023-08-31
feat(HACBS-2463): create ReleaseServiceConfig CRD

johnbieren:
This commit creates the ReleaseServiceConfig CRD. Through this new API, it is possible to specify whether or not local pipelines can be executed through the ReleaseStrategy. This is done through changes in the release adapter.


mmalina:
I was trying to find out if there is a chance this is considered an anti pattern - having a CRD for a single CR (basically a singleton). It's probably ok. But I found [this](https://www.reddit.com/r/kubernetes/comments/11t8qa9/need_an_operator_that_reads_a_single_resource_is/) interesting:

> OpenShift cluster wide configs (consumed by their operators) are basically singleton CRDs, all named cluster. There is validation to ensure you can only create one named cluster.

It sounds like what they do is allow the CR to be located in any namespace and look it up by name (while ensuring there is just one)? Any thoughts on that? Probably doing it as you suggest (having it in a specific namespace and load from there) is good enough.

johnbieren:
> I was trying to find out if there is a chance this is considered an anti pattern - having a CRD for a single CR (basically a singleton). It's probably ok. But I found [this](https://www.reddit.com/r/kubernetes/comments/11t8qa9/need_an_operator_that_reads_a_single_resource_is/) interesting:
> 
> > OpenShift cluster wide configs (consumed by their operators) are basically singleton CRDs, all named cluster. There is validation to ensure you can only create one named cluster.
> 
> It sounds like what they do is allow the CR to be located in any namespace and look it up by name (while ensuring there is just one)? Any thoughts on that? Probably doing it as you suggest (having it in a specific namespace and load from there) is good enough.

@davidmogar thoughts on ^? The ticket suggests following how internal-services does it, which is the style I went with

davidmogar:
> > I was trying to find out if there is a chance this is considered an anti pattern - having a CRD for a single CR (basically a singleton). It's probably ok. But I found [this](https://www.reddit.com/r/kubernetes/comments/11t8qa9/need_an_operator_that_reads_a_single_resource_is/) interesting:
> > > OpenShift cluster wide configs (consumed by their operators) are basically singleton CRDs, all named cluster. There is validation to ensure you can only create one named cluster.
> > 
> > 
> > It sounds like what they do is allow the CR to be located in any namespace and look it up by name (while ensuring there is just one)? Any thoughts on that? Probably doing it as you suggest (having it in a specific namespace and load from there) is good enough.
> 
> @davidmogar thoughts on ^? The ticket suggests following how internal-services does it, which is the style I went with

I think there's value on providing the config in the same namespace where the service is deployed.

api/v1alpha1/releaseserviceconfig_types.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+const ReleaseServiceConfigResourceName string = "config"
+
+// ReleaseServiceConfigSpec defines the desired state of ReleaseServiceConfig.
+type ReleaseServiceConfigSpec struct {
+	// AllowLocalPipelines is the boolean that specifies whether or not pipelines stored locally
+	// on the cluster may be used
+	// +optional
+	AllowLocalPipelines bool `json:"allowLocalPipelines,omitempty"`
```
davidmogar, line 30: What about to just call it debug and this being a side effect of the debug mode?

api/v1alpha1/releaseserviceconfig_types.go
```diff
@@ -0,0 +1,65 @@
...
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ReleaseServiceConfig `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&ReleaseServiceConfig{}, &ReleaseServiceConfigList{})
+}
+
+// IsBundleLocalAndAllowed checks whether the bundle value is empty if AllowLocalPipelines is False.
+func (r *ReleaseServiceConfig) IsBundleLocalAndAllowed(bundle string) bool {
```
davidmogar, line 63: Probably cleaner to just access the values directly although I don't oppose. If you don't go for the `debug` field name, this function should be renamed to something better. Maybe something like `localPipelinesAllowed` for readability or `AreLocalPipelinesAllowed` to start with  `Is`/`Are` like in the Release type.

controllers/release/adapter.go
```diff
@@ -570,3 +607,31 @@ func (a *adapter) validateProcessingResources() *controller.ValidationResult {
...
+	releasePlanAdmission, err := a.loader.GetActiveReleasePlanAdmissionFromRelease(a.ctx, a.client, a.release)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			a.release.MarkValidationFailed(err.Error())
+			return &controller.ValidationResult{Valid: false}
+		}
+		return &controller.ValidationResult{Err: err}
+	}
+
+	releaseStrategy, err := a.loader.GetReleaseStrategy(a.ctx, a.client, releasePlanAdmission)
```
davidmogar, line 622: I was about to suggest to create a `GetReleaseStrategyFromRelease`, but this resource is meant to disappear anyway.

controllers/release/adapter.go
```diff
@@ -69,11 +71,36 @@ func newAdapter(ctx context.Context, client client.Client, release *v1alpha1.Rel
 	releaseAdapter.validations = []controller.ValidationFunction{
 		releaseAdapter.validateProcessingResources,
 		releaseAdapter.validateAuthor,
+		releaseAdapter.validatePipelineRef,
```
davidmogar, line 74: Beautiful :heart_eyes:

controllers/release/adapter_test.go
```diff
@@ -960,6 +1010,25 @@ var _ = Describe("Release adapter", Ordered, func() {
 		})
 	})
 
+	When("When calling getEmptyReleaseServiceConfig", func() {
```
davidmogar, line 1013:
```suggestion
	When("getEmptyReleaseServiceConfig is called", func() {
```

controllers/release/adapter_test.go
```diff
@@ -1398,6 +1471,131 @@ var _ = Describe("Release adapter", Ordered, func() {
 			result := adapter.validateProcessingResources()
 			Expect(result.Valid).To(BeFalse())
 			Expect(result.Err).To(HaveOccurred())
+			Expect(adapter.release.IsValid()).To(BeFalse())
+		})
+	})
+
+	When("calling validatePipelineRef", func() {
```
davidmogar, line 1478:
```suggestion
	When("validatePipelineRef is called", func() {
```

controllers/release/controller.go
```diff
@@ -74,6 +76,7 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	return controller.ReconcileHandler([]controller.Operation{
 		adapter.EnsureFinalizersAreCalled,
 		adapter.EnsureReleaseIsRunning,
+		adapter.EnsureConfigIsLoaded,
```
davidmogar, line 79: This should be the first one probably. Maybe the second. In addition, if you check the internal-services repo you will see this comment:
```
// This operation sets the config in the adapter to be used in other operations.
```
I actually hate that but I think it's the cleanest. Maybe we should use different names for this operations that update the adapter... Or maybe this should not be an operation at all and happen in the newAdapter. Can we discuss it?

controllers/release/adapter.go
johnbieren: Yeah, I almost did create just that

api/v1alpha1/releaseserviceconfig_types.go
johnbieren: Okay, I switched (not pushed yet) `AllowLocalPipelines` to `Debug`. What should I name this then? Should I have a debug check function with a TBD name that calls this? Since there will be more debug checks than just that? What are your thoughts

johnbieren: @davidmogar

davidmogar: In the internal services repo we access the `Debug` field directly without a method.  Regarding whether we will have more debug checks... Not sure... Probably not. And even if we have them, maybe the `Debug mode` should be consistent and just to a set of things always. In other words, not a configurable `Debug mode`. Wdyt?

johnbieren: I would expect debug mode to ignore the pipelineAllowList too, so it may have more checks. I am not sure I follow the second part of your response about it being a set of things. As in, if debug == true, you get these things (like local bundles allowed, pipelineAllowList ignored), if debug == false, production behvaior. no other option?

davidmogar: Yeah, that's what I meant. Reason for this is that the AllowLocalPipelines is a debug thing. It's not something we will offer to users.

controllers/release/controller.go
johnbieren: comment added and order changed

controllers/release/adapter_test.go
johnbieren: fixed

controllers/release/adapter_test.go
johnbieren: fixed

api/v1alpha1/releaseserviceconfig_types.go
johnbieren: switched to debug

api/v1alpha1/releaseserviceconfig_types.go
johnbieren: modified

api/v1alpha1/releaseserviceconfig_types.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const ReleaseServiceConfigResourceName string = "config"
+
+// ReleaseServiceConfigSpec defines the desired state of ReleaseServiceConfig.
+type ReleaseServiceConfigSpec struct {
+	// Debug is the boolean that specifies whether or not the Release Service should run
```
davidmogar, line 27: The boolean or a boolean?

api/v1alpha1/releaseserviceconfig_types.go
```diff
@@ -0,0 +1,65 @@
...
+	Items           []ReleaseServiceConfig `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&ReleaseServiceConfig{}, &ReleaseServiceConfigList{})
+}
+
+// AreDebugOptionsValid checks whether the bundle value is empty if Debug mode is False.
+func (r *ReleaseServiceConfig) AreDebugOptionsValid(bundle string) bool {
+	return r.Spec.Debug || bundle != ""
```
davidmogar, line 64: I don't think this belongs here. This is adapter/tekton logic.

controllers/release/adapter.go
```diff
@@ -570,3 +607,31 @@ func (a *adapter) validateProcessingResources() *controller.ValidationResult {
...
+	releaseStrategy, err := a.loader.GetReleaseStrategy(a.ctx, a.client, releasePlanAdmission)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			a.release.MarkValidationFailed(err.Error())
+			return &controller.ValidationResult{Valid: false}
+		}
+		return &controller.ValidationResult{Err: err}
+	}
+
+	if !a.releaseServiceConfig.AreDebugOptionsValid(releaseStrategy.Spec.Bundle) {
```
davidmogar, line 631: This should just be `if !a.releaseServiceConfig.Debug && releaseStrategy.Spec.Bundle == ""`. No need for that function. In addition, Martin is modifying this part so we should be able to do something like:
```
<releaseStrategy|rpa|rp>.Spec.PipelineRef.UsesBundleResolver()
<releaseStrategy|rpa|rp>.Spec.PipelineRef.UsesClusterResolver()
<releaseStrategy|rpa|rp>.Spec.PipelineRef.UsesGitResolver()
```
That or something similar, as we will control the PipelineRef (it won't be the Tekton one but one defined in Operator Toolkit).

api/v1alpha1/releaseserviceconfig_types.go
johnbieren: It is "the" boolean that specifies whether or not we run in debug mode. There is no other boolean controlling that. It is the only one

api/v1alpha1/releaseserviceconfig_types.go
johnbieren: Why? Why does the adapter care about how the ReleaseServiceConfig makes decisions? Also your previous comment was 
`this function should be renamed to something better. Maybe something like localPipelinesAllowed for readability or AreLocalPipelinesAllowed to start with Is/Are like in the Release type.`
which made it sound like this function is fine here

api/v1alpha1/releaseserviceconfig_types.go
davidmogar: It's also "a" flag that enables debug features ;)

But you are the native one here so I trust on your judgement.

controllers/release/adapter.go
johnbieren: I don't see why we don't want a helper function for this. We will have bundle == "", is pipeline in the allow list, is pipeline in the disallow list, does it use bundle resolver, does it use cluster resolver, does it use git resolver, and who knows what other checks will come up. I think that warrants a helper function

api/v1alpha1/releaseserviceconfig_types.go
davidmogar: Because before it was just an accesor. Now it's making decisions and a config file/resource should not make decisions. We have the adapter for that.

johnbieren: I'm not following. All that changed from the previous version was a field in the spec was renamed. The logic is the same as when this function was okay

controllers/release/adapter.go
davidmogar: It can be just `IsClusterScoped()`. And I'm not opposing to a helper function. I'm just saying that the config resource should not have that logic.

api/v1alpha1/releaseserviceconfig_types.go
davidmogar: Maybe I didn't reached that far. In any case, I don't think that logic should be in here.

loader/loader.go
```diff
@@ -186,6 +187,13 @@ func (l *loader) GetReleasePlan(ctx context.Context, cli client.Client, release
 	return releasePlan, toolkit.GetObject(release.Spec.ReleasePlan, release.Namespace, cli, ctx, releasePlan)
 }
 
+// GetReleaseServiceConfig returns the Release with the given name and namespace. If the ReleaseServiceConfig is not
```
davidmogar, line 190:
```suggestion
// GetReleaseServiceConfig returns the ReleaseServiceConfig with the given name and namespace. If the ReleaseServiceConfig is not
```

davidmogar (changes requested):

loader/loader.go
johnbieren: fixed

controllers/release/controller.go
```diff
@@ -73,6 +74,7 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	adapter := newAdapter(ctx, c.client, release, loader.NewLoader(), &logger)
 
 	return controller.ReconcileHandler([]controller.Operation{
+		adapter.EnsureConfigIsLoaded, // This operation sets the config in the adapter to be used in other operations.
```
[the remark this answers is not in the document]
johnbieren: fixed

controllers/release/controller.go
```diff
@@ -53,6 +53,7 @@ type Controller struct {
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=applications/finalizers,verbs=update
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=enterprisecontractpolicies,verbs=get;list;watch
 //+kubebuilder:rbac:groups=appstudio.redhat.com,resources=enterprisecontractpolicies/status,verbs=get
+//+kubebuilder:rbac:groups=appstudio.redhat.com,resources=releaseserviceconfigs,verbs=get;list;watch;create;update;patch;delete
```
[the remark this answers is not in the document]
johnbieren: It threw errors without `list` in the e2e test. I removed the others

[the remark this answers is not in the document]
davidmogar: Interesting...

PR #248, opened 2023-09-04
refactor(HACBS-2550): pass git resolver info for verify-ec task

johnbieren:
The EnterpriseContact ConfigMap now contains the information needed for the git resolver of the proper task to use. This commit changes the adapter from passing the task bundle information to passing the task git resolver information.

### DIFF
--- a/tekton/pipeline_run.go
+++ b/tekton/pipeline_run.go
@@ -20,9 +20,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"strings"
 	"unicode"
+
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


johnbieren:
Pairs with https://github.com/redhat-appstudio/e2e-tests/pull/749

tekton/pipeline_run_test.go
```diff
@@ -216,7 +219,24 @@ var _ = Describe("PipelineRun", func() {
 
 		It("can add the EC task bundle parameter to the PipelineRun", func() {
```
mmalina, line 220: The wording should be modified here. bundle -> git resolver information

tekton/pipeline_run_test.go
```diff
@@ -91,7 +91,10 @@ var _ = Describe("PipelineRun", func() {
 				Kind: "ConfigMap",
 			},
 			Data: map[string]string{
-				"verify_ec_task_bundle": "test-bundle",
+				"verify_ec_task_bundle":         "test-bundle",
```
mmalina, line 94: We no longer need this one, no?

tekton/pipeline_run_test.go
```diff
@@ -216,7 +219,24 @@ var _ = Describe("PipelineRun", func() {
 
 		It("can add the EC task bundle parameter to the PipelineRun", func() {
 			releasePipelineRun.WithEnterpriseContractConfigMap(enterpriseContractConfigMap)
-			Expect(releasePipelineRun.Spec.Params).Should(ContainElement(HaveField("Value.StringVal", Equal(string("test-bundle")))))
+			params := releasePipelineRun.Spec.Params
+			checked := 0
+
+			for i := range params {
```
mmalina, line 225: Can't we just check the params directly (relying on the ordering)? We do it in other tests in this file as well, so should be fine. But this is fine too.

tekton/pipeline_run.go
```diff
@@ -70,16 +71,16 @@ func (r *ReleasePipelineRun) AsPipelineRun() *tektonv1beta1.PipelineRun {
 	return &r.PipelineRun
 }
 
-// WithEnterpriseContractConfigMap adds a param providing the verify ec task bundle to the release PipelineRun.
+// WithEnterpriseContractConfigMap adds a param providing the verify ec task git resolver information to the release PipelineRun.
 func (r *ReleasePipelineRun) WithEnterpriseContractConfigMap(configMap *corev1.ConfigMap) *ReleasePipelineRun {
```
davidmogar, line 75: Could we use this PR to give a better name to the parameter? Something like `ecConfig` or similar?

johnbieren: You think changing the function name to replace `EnterpriseContract` with `ec` makes it better? I don't

tekton/pipeline_run_test.go
johnbieren: Considering the order doesn't matter, I'd really not rather rely on the ordering if we don't have to

tekton/pipeline_run_test.go
johnbieren: fixed

tekton/pipeline_run_test.go
johnbieren: removed

kasemAlem (dismissed):

tekton/pipeline_run.go
mmalina: I think David meant the argument of this function: `configMap` -> `ecConfig`

tekton/pipeline_run_test.go
mmalina: ok, fine with me

mmalina (approved):

tekton/pipeline_run.go
johnbieren: got it, thanks for clarifying guys. Updated

tekton/pipeline_run_test.go
davidmogar: I think it would be great to do it as Martin suggest. Having said this, the whole package will require a refactor whenever we introduce dev pipelines. Test are also a bit outdated in this package. So we can do it in a different PR.

davidmogar (approved):

PR #252, opened 2023-09-12
feat(HACBS-2461): add new pipeline_ref type 

mmalina:
This will be used for enabling git resolvers in release strategy CR.

### DIFF
--- a/tekton/utils/pipeline_ref.go
+++ b/tekton/utils/pipeline_ref.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+
+// PipelineRef represents a reference to a Pipeline using a resolver.
+type PipelineRef struct {
+	// Resolver is the name of a Tekton resolver to be used (e.g. git)
+	Resolver string `json:"resolver"`
+
+	// Params is a slice of parameters for a given resolver
+	Params []Param `json:"params"`
+}
+
+// Param defines the parameters for a given resolver in PipelineRef
+type Param struct {
+	// Name is the name of the parameter
+	Name string `json:"name"`
+
+	// Value is the value of the parameter
+	Value string `json:"value"`
+}
+
+// ToTektonPipelineRef converts a PipelineRef object to Tekton's own PipelineRef type and returns it.
+func (pr PipelineRef) ToTektonPipelineRef() tektonv1.PipelineRef {
+	params := tektonv1.Params{}
+
+	for _, p := range pr.Params {
+		params = append(params, tektonv1.Param{
+			Name: p.Name,
+			Value: tektonv1.ParamValue{
+				Type:      tektonv1.ParamTypeString,
+				StringVal: p.Value,
+			},
+		})
+	}
+
+	tektonPipelineRef := tektonv1.PipelineRef{
+		ResolverRef: tektonv1.ResolverRef{
+			Resolver: tektonv1.ResolverName(pr.Resolver),
+			Params:   params,
+		},
+	}
+
+	return tektonPipelineRef
+}
+
+// IsClusterScoped returns whether the PipelineRef uses a cluster resolver or not.
+func (pr PipelineRef) IsClusterScoped() bool {
+	return pr.Resolver == "cluster"
+}
--- a/tekton/utils/pipeline_ref_test.go
+++ b/tekton/utils/pipeline_ref_test.go
@@ -1,0 +1,103 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Release type", func() {
+	var (
+		clusterRef PipelineRef
+		gitRef     PipelineRef
+		bundleRef  PipelineRef
+	)
+
+	BeforeEach(func() {
+		clusterRef = PipelineRef{
+			Resolver: "cluster",
+			Params: []Param{
+				{Name: "kind", Value: "pipeline"},
+				{Name: "name", Value: "my-cluster-pipeline"},
+				{Name: "namespace", Value: "my-namespace"},
+			},
+		}
+		gitRef = PipelineRef{
+			Resolver: "git",
+			Params: []Param{
+				{Name: "url", Value: "my-git-url"},
+				{Name: "revision", Value: "my-revision"},
+				{Name: "pathInRepo", Value: "my-path-in-repo"},
+			},
+		}
+		bundleRef = PipelineRef{
+			Resolver: "bundles",
+			Params: []Param{
+				{Name: "bundle", Value: "my-bundle"},
+				{Name: "name", Value: "my-pipeline"},
+				{Name: "kind", Value: "Pipeline"},
+			},
+		}
+	})
+
+	When("ToTektonPipelineRef method is called", func() {
+		It("should return Tekton PipelineRef representation of the PipelineRef", func() {
+			ref := clusterRef.ToTektonPipelineRef()
+			Expect(string(ref.ResolverRef.Resolver)).To(Equal("cluster"))
+			params := ref.ResolverRef.Params
+			Expect(params[0].Name).To(Equal("kind"))
+			Expect(params[0].Value.StringVal).To(Equal("pipeline"))
+			Expect(params[1].Name).To(Equal("name"))
+			Expect(params[1].Value.StringVal).To(Equal("my-cluster-pipeline"))
+			Expect(params[2].Name).To(Equal("namespace"))
+			Expect(params[2].Value.StringVal).To(Equal("my-namespace"))
+
+			ref = gitRef.ToTektonPipelineRef()
+			Expect(string(ref.ResolverRef.Resolver)).To(Equal("git"))
+			params = ref.ResolverRef.Params
+			Expect(params[0].Name).To(Equal("url"))
+			Expect(params[0].Value.StringVal).To(Equal("my-git-url"))
+			Expect(params[1].Name).To(Equal("revision"))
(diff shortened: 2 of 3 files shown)


api/v1alpha1/releasestrategy_types.go
```diff
@@ -22,15 +22,32 @@ import (
 
 // ReleaseStrategySpec defines the desired state of ReleaseStrategy
 type ReleaseStrategySpec struct {
-	// Release Tekton Pipeline to execute
+	// Release Tekton Pipeline to execute; used with bundle resolver;
```
johnbieren, line 25: this comment is messed up

api/v1alpha1/releasestrategy_types.go
```diff
@@ -22,15 +22,32 @@ import (
...
+	// +optional
+	Pipeline string `json:"pipeline,omitempty"`
 
-	// Bundle is a reference to the Tekton bundle where to find the pipeline
+	// Bundle is a reference to the Tekton bundle where to find the pipeline;
+	// used with bundle resolver
 	// +optional
 	Bundle string `json:"bundle,omitempty"`
 
+	// Git repo URL where the Release Tekton Pipeline to execute is located;
```
johnbieren, line 36: can you add the kubebuilder validations for these?

johnbieren (dismissed):

api/v1alpha1/releasestrategy_types.go
mmalina: I guess I could. But for the bundle url we never had one either, so I wonder if we need it.

What kind of patterns would you suggest? Some generic url regexp for the url? What about the `GitRevision` and `GitPathInRepo`? I guess I could go find out what github allows for revision strings (we will restrict it later on, but right now I should allow anything) and filenames and create something based on that, but is it needed/worth it?

kasemAlem (approved):

api/v1alpha1/releasestrategy_types.go
```diff
@@ -22,15 +22,31 @@ import (
...
+
+	// Revision of git repo where the Release Tekton Pipeline to execute is located;
+	// used with git resolver
+	// +optional
+	GitRevision string `json:"gitRevision,omitempty"`
+
+	// Relative path in git repo where the Release Tekton Pipeline to execute is located;
+	// used with git resolver
+	// +optional
+	GitPathInRepo string `json:"gitPathInRepo,omitempty"`
```
davidmogar, line 48: This should be a struct. Something like `Resolver` which would have a `BundleReference` or a `GitRefeference`. I'd make both of them structs as well so for example, the `GitReference` one would have the fields `Url`, `Revision` and `PathInRepo` while the `BundleReference` would just have a `Bundle` field.

The change to support dev will come with plenty of changes, so I don't really mind doing that here. @johnbieren?

johnbieren: I'm not sure I follow what the question is

mmalina: @davidmogar ok, I will do that. But in that case it will be a breaking change, so it definitely needs to be done together with the controller change.

> the BundleReference would just have a Bundle field.

You also need a name for the bundle resolver. (Currently the `Pipeline` field.)

davidmogar: > I'm not sure I follow what the question is

Whether we should do it here or wait for the change supporting dev pipelines.

johnbieren: It doesn't matter to me whether it is done all in this PR after https://github.com/redhat-appstudio/release-service/pull/246 merges or in a followup. Whatever is easiest

api/v1alpha1/releasestrategy_types.go
```diff
@@ -52,6 +47,43 @@ type ReleaseStrategySpec struct {
...
+}
+
+// BundleReference holds a reference to the Pipeline using bundle resolver
+type BundleReference struct {
+	// Name is the name of Pipeline inside the bundle
+	// +required
+	Name string `json:"name"`
+
+	// Bundle is a reference to the Tekton bundle where to find the pipeline
+	// +required
```
johnbieren, line 68: should this really be required @davidmogar ? What if I am running in debug mode?

mmalina: We will use a different approach. See here: https://github.com/redhat-appstudio/operator-toolkit/pull/18/files

That way we will allow all resolver types (including `cluster` for local Pipelines).

tekton/utils/pipeline_ref_test.go
```diff
@@ -0,0 +1,83 @@
...
+	// 			"Reason":  Equal("conditionReason"),
+	// 			"Message": Equal(""),
+	// 		}))
+	// 	})
+	// })
+
+	When("IsClusterScoped method is called", func() {
+		It("should return true for a cluster pipeline", func() {
+			Expect(clusterPpr.IsClusterScoped()).To(BeTrue())
+		})
```
davidmogar, line 76: Empty line after this one please ;)

tekton/utils/pipeline_ref_test.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+var _ = Describe("Release type", func() {
+	var (
+		clusterPpr PipelineRef
+		gitPpr     PipelineRef
+		bundlePpr  PipelineRef
+	)
+
+	BeforeEach(func() {
```
davidmogar, line 32: I'd declare them in each test in this particular case.

davidmogar (changes requested):

tekton/utils/pipeline_ref_test.go
```diff
@@ -0,0 +1,83 @@
+/*
+Copyright 2022.
```
davidmogar, line 2:
```suggestion
Copyright 2023.
```

tekton/utils/pipeline_ref_test.go
```diff
@@ -0,0 +1,83 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	// . "github.com/onsi/gomega/gstruct"
```
davidmogar, line 22:
```suggestion
```

tekton/utils/suite_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	//+kubebuilder:scaffold:imports
+)
+
+func Test(t *testing.T) {
+	RegisterFailHandler(Fail)
+
+	RunSpecs(t, "Conditions Suite")
```
davidmogar, line 33:
```suggestion
	RunSpecs(t, "Tekton Utils Suite")
```

davidmogar (changes requested):

tekton/utils/pipeline_ref_test.go
mmalina: Done.

tekton/utils/pipeline_ref_test.go
mmalina: Removed. Along with other comments.

tekton/utils/suite_test.go
mmalina: Fixed.

tekton/utils/pipeline_ref_test.go
mmalina: In this case it would be exactly the same definition done twice, so I think it makes to do it just once. Please check.

tekton/utils/pipeline_ref.go
```diff
@@ -0,0 +1,68 @@
...
+			Resolver: tektonv1.ResolverName(pr.Resolver),
+			Params:   params,
+		},
+	}
+
+	return tektonPipelineRef
+}
+
+// IsClusterScoped returns true if the PipelineRef is cluster
+// scoped and false otherwise.
```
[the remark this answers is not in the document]
mmalina: I did think of this, but thought you wouldn't like it as it's not explicit enough about exactly what is returned :) But ok, used it now.

tekton/utils/pipeline_ref.go
```diff
@@ -0,0 +1,68 @@
...
+limitations under the License.
+*/
+
+package utils
+
+import tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+
+// PipelineRef represents a reference to a Pipeline using a resolver.
+type PipelineRef struct {
+	// Resolver is name of the resolver to be used, e.g. git
```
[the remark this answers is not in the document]
mmalina: I disagree. It's a resolver, but once you pick one and want to use it, it's the resolver to be used ;) But neither of us is native and I don't really care :) Fixed.

tekton/utils/pipeline_ref.go
```diff
@@ -0,0 +1,68 @@
...
+type Param struct {
+	// Name is the name of the parameter
+	Name string `json:"name"`
+
+	// Value is the value of the parameter
+	Value string `json:"value"`
+}
+
+// ToTektonPipelineRef converts a PipelineRef object to
+// Tekton's own PipelineRef type and returns it.
```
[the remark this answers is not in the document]
mmalina: Fixed.

api/v1alpha1/releasestrategy_types.go
mmalina: Change of plans. The PipelineRef type will be added to this repo in this PR (and nothing else - the rest will come later).

davidmogar (dismissed):

tekton/utils/pipeline_ref.go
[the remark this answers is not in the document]
davidmogar: If you mean `a` or `the` I'm totally fine with that. I just want you to reference Tekton there ;)

tekton/utils/pipeline_ref.go
```diff
@@ -0,0 +1,66 @@
...
+		ResolverRef: tektonv1.ResolverRef{
+			Resolver: tektonv1.ResolverName(pr.Resolver),
+			Params:   params,
+		},
+	}
+
+	return tektonPipelineRef
+}
+
+// IsClusterScoped returns whether the PipelineRef uses a cluster resolver or not.
```
johnbieren, line 63: I am a little confused why we have this function (and none for bundle/git)

tekton/utils/pipeline_ref.go
johnbieren, line 1: Any reason why this went in utils instead of just `tekton/`? It is a struct, not really a util

johnbieren (approved):
Confused as to why one of the functions is there but otherwise lgtm

tekton/utils/pipeline_ref.go
davidmogar: We only need this one so far for the debug mode. Can be added later if needed.

davidmogar (approved):

tekton/utils/pipeline_ref.go
mmalina: 1. Yes, because there were cyclical dependencies. (Or would be once we use this in the releasestrategy struct.)

2. Yes, I pointed this out, but David said we would refactor that stuff soon anyway, so we shouldn't worry about it.

tekton/utils/pipeline_ref.go
mmalina: Local/cluster scoped pipelines will only be allowed in debug mode. So this function will be used in that logic - if we're not in debug mode and it is cluster scope, we reject it.

tekton/utils/pipeline_ref.go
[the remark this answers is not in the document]
mmalina: Oh! I totally missed that! I thought it was all about the/a :D Fixed.

tekton/utils/pipeline_ref.go
johnbieren: I'll talk to David about it. The only import in here is the upstream tekton api. I do not see the cycle

davidmogar:
@johnbieren please resolve your comments. Anything that needs to be changed will be done in my PR.

happybhati (approved):

PR #255, opened 2023-09-24
fix: update filters and panels

davidmogar:
The Release Grafana dashboard was not working properly. I found out that some filters were not properly defined. I also saw that one of the panels was deprecated, so I updated it.

In any case, there seems to be an unrelated problem. Just editing the panels and clicking on the query was enough for the panel to start showing values (without any changes). Let's see how it behaves after this change.

### DIFF
--- a/config/grafana/dashboards/release_dashboard.json
+++ b/config/grafana/dashboards/release_dashboard.json
@@ -22,13 +22,17 @@
     ]
   },
   "editable": true,
-  "gnetId": null,
+  "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 4,
+  "id": 16,
   "links": [],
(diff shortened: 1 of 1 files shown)


config/grafana/dashboards/release_dashboard.json
```diff
@@ -956,5 +989,6 @@
   "timezone": "",
   "title": "Release Service",
   "uid": "b065ed16540677415b31adb8574bb4d9aa9f74e3",
-  "version": 1
-}
+  "version": 1,
+  "weekStart": ""
+}
```
scoheb, line 994: No new line at end

scoheb (changes requested):
Small change

scoheb (approved):

PR #260, opened 2023-09-26
chore(RHTAPREL-435): convert pipelines for tenants-config

scoheb:
- Cleanup .tekton pipelines such that they can be used in conjunction with tenants-config

### DIFF
--- a/.tekton/push.yaml
+++ b/.tekton/push.yaml
@@ -15,15 +15,11 @@ spec:
       value: "{{ revision }}"
     - name: output-image
       value: 'quay.io/redhat-appstudio/release-service:{{ revision }}'
-    - name: infra-deployment-update-script
-      value: |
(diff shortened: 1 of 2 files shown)


.tekton/push.yaml
```diff
@@ -15,15 +15,11 @@ spec:
       value: "{{ revision }}"
     - name: output-image
       value: 'quay.io/redhat-appstudio/release-service:{{ revision }}'
-    - name: infra-deployment-update-script
```
johnbieren, line 18: We don't want to update infra-deployments with the latest version of release-service anymore?

scoheb: we will. but it will now be taken care of by https://github.com/redhat-appstudio/release-service-catalog/tree/main/pipelines/rhtap-service-push

scoheb:
closing in favour of #261 and #262

PR #271, opened 2023-10-03
feat: remove ReleaseStrategy

davidmogar:


davidmogar:
> @davidmogar this is superseeded by #273 , right? It seems the same?

Yes. I need to pair it with an E2E change and for that I have to create a PR from a fork.

kasemAlem:
should be also more PR for [release-integration-tests](https://github.com/hacbs-release/release-integration-tests) ?

PR #273, opened 2023-10-04
feat(RHTAPREL-298): remove ReleaseStrategy

davidmogar:


davidmogar:
> @davidmogar can you please add the jira reference in PR title (I see it's included in the first commit's title). And also the description. Also, the second commit doesn't conform to gitlint rules.

Taht's just for review purposes. I'll fix the commits before merging.

api/v1alpha1/releaseplan_types.go
```diff
@@ -17,20 +17,34 @@ limitations under the License.
...
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
 	// +required
 	Application string `json:"application"`
 
+	// Data is an unstructured key used for providing data for the release Pipeline
+	// +kubebuilder:pruning:PreserveUnknownFields
+	// +optional
+	Data *runtime.RawExtension `json:"data,omitempty"`
+
+	// PipelineRef is an optional reference to a Pipeline that would be executed
```
mmalina, line 37: This is the first time I hear of this. Where is this defined? I don't see this in the linked Jira. Also, maybe the PR description / commit message could say a bit more about these changes as well?

davidmogar: This is what you wrote but just moved to the shared-utils repo.

.idea/misc.xml
```diff
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
davidmogar: Yeah, in some repos the whole directory was ignored but not in this one. I'll do it in a different PR as it doesn't affect the IDE at all.

api/v1alpha1/releaseplan_types.go
davidmogar: Oh, right. I didn't get your comment. This is a feature that is coming. The code for it will be added with a different ticket but I wanted to do all the API changes on the same PR.

mmalina: ok. but this PR still has no description. could we have at least a brief summary of what is being done here besides just removing ReleaseStrategy?

api/v1alpha1/releaseplanadmission_types.go
```diff
@@ -59,10 +65,7 @@ type ReleasePlanAdmissionStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="Display Name",type=string,priority=1,JSONPath=`.spec.displayName`
-// +kubebuilder:printcolumn:name="Application",type=string,JSONPath=`.spec.application`
```
johnbieren, line 63: Is it possible to print a list here? Having the application names show here would be quite nice

johnbieren (commented):
Biggest thing for me was that I wasn't expecting this PR to be moving to shared-utils. I thought we were revisiting if we were going to use that repo. Not 100% sure I didn't miss anything, but if I did, surely the e2e tests will catch it

api/v1alpha1/releaseplan_types.go
```diff
@@ -17,20 +17,34 @@ limitations under the License.
 package v1alpha1
 
 import (
+	shared "github.com/redhat-appstudio/shared-utils/tekton"
```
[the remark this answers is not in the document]
davidmogar: As I commented with you, the final version of the whole change was already implemented (the one including shared-utils) so I had to revert all of that. This I just simply forget about it. I restored the package and left the named imports so it's easy to go back to shared-utils later on.

api/v1alpha1/releaseplan_types.go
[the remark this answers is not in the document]
davidmogar: I think that it's fine here as we talk about a pipeline to be executed before the release pipeline... Later on, when it's finally implemented, we will have to add a name for sure... In my code I used `dev` and `managed` instead of `release`. But that code is reverted and will come in a separate PR.

api/v1alpha1/releaseplanadmission_types.go
davidmogar: No. This is the info printed by kubectl in a table. We would break the table easily. It will have to be a detailed view.

api/v1alpha1/releaseplan_types.go
davidmogar: As soon as I fix the commits after an initial review.

johnbieren:
Why did you add this `Use fork` commit?

api/v1alpha1/releaseplan_types.go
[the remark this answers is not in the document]
johnbieren: Then if, after review, we decide not to use shared-utils (as is my preference) we have all these imports changed to `shared` and it not making much sense. Why not do keep all the shared-utils stuff in one PR, instead of half now half later

davidmogar:
> Why did you add this `Use fork` commit?

We did it in the previous PR as well to point to the fork. It not I couldn't get the code in E2E. The whole commit will be deleted before merging this.

johnbieren:
> > Why did you add this `Use fork` commit?
> 
> We did it in the previous PR as well to point to the fork. It not I couldn't get the code in E2E. The whole commit will be deleted before merging this.

I've never had to make a commit like that for the pairing. Maybe I am missing something?

davidmogar:
> > > Why did you add this `Use fork` commit?
> > 
> > 
> > We did it in the previous PR as well to point to the fork. It not I couldn't get the code in E2E. The whole commit will be deleted before merging this.
> 
> I've never had to make a commit like that for the pairing. Maybe I am missing something?

Maybe I'm doing the imports in the wrong way, but it could be also because the new PipelineRef stuff. It also affects API changes. If you don't change the API not much of a problem. In any case, not a big deal. It will be removed any way.

johnbieren (commented):
The PR LGTM once the `s/redhat-appstudio/davidmogar/g` switch is removed. When you remove that, can you also remove the `.idea/misc.xml` file from the PR too please? Then I will be good to approve

johnbieren (approved):

ralphbean:
:partying_face:

PR #275, opened 2023-10-10
fix(RHTAPREL-630): address high cves due to ubi-micro

scoheb:
- use ubi-micro 9.2 over 9.1.0


johnbieren (dismissed):

kasemAlem (dismissed):

Dockerfile
```diff
@@ -28,7 +28,8 @@ ENV ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS}
 
 # Use ubi-micro as minimal base image to package the manager binary
 # See https://catalog.redhat.com/software/containers/ubi9/ubi-micro/615bdf943f6014fa45ae1b58
-FROM registry.access.redhat.com/ubi9/ubi-micro:9.1.0
+# use 9.2 over 9.1.0 due to CVEs (glibc-common-2.34-40.el9_1.1 and libcap-2.48-8.el9)
```
mmalina, line 31: Thanks for updating and I definitely agree. But do we need this comment? Ideally, updating image references should be done regularly anyway. I don't think we need to list what bugfixes were included in the currently used version.

Dockerfile
```diff
@@ -28,7 +28,8 @@ ENV ENABLE_WEBHOOKS=${ENABLE_WEBHOOKS}
...
+FROM registry.access.redhat.com/ubi9/ubi-micro:9.2
```
mmalina, line 32: I think we always lock to a particular image - either using a shasum, or by using the full version. But 9.2 is a floating reference, at least I think it is. So this should be `registry.access.redhat.com/ubi9/ubi-micro:9.2-15.1696515526`

Dockerfile
scoheb: I guess we can remove. we can see the fixes via the linked JIRA issue

mmalina (approved):

PR #279, opened 2023-10-18
feat(RHTAPREL-119): make target optional in ReleasePlans

johnbieren:
If a user wants to run a dev release pipeline with no managed release pipeline, it really doesn't make sense to have a target workspace. This commit makes the target workspace optional instead of required in the ReleasePlan Spec.

### DIFF
--- a/api/v1alpha1/releaseplan_types.go
+++ b/api/v1alpha1/releaseplan_types.go
@@ -47,7 +47,7 @@ type ReleasePlanSpec struct {
 
 	// Target references where to send the release requests
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
-	// +required
+	// +optional
 	Target string `json:"target"`
 }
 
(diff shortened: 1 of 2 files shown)


api/v1alpha1/releaseplan_types.go
```diff
@@ -47,7 +47,7 @@ type ReleasePlanSpec struct {
 
 	// Target references where to send the release requests
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
-	// +required
+	// +optional
```
scoheb, line 50: does this require any special handling in the controller since it might expect this is required?

johnbieren: hmm probably? but I'm not sure we need to handle it yet; it can likely be handled in the PR that adds support for the dev pipelines. Maybe this is better off as part of that PR instead. @davidmogar thoughts?

davidmogar: Yeah, probably better to add it there.

PR #292, opened 2023-11-02
feat(RHTAPREL-278): add status fields to RPs and RPAs

johnbieren:
This commit adds conditions and a matched resource to the Status struct of ReleasePlans and ReleasePlanAdmissions.


api/v1alpha1/releaseplan_types.go
```diff
@@ -53,6 +55,15 @@ type ReleasePlanSpec struct {
...
 type ReleasePlanStatus struct {
+	// Conditions represent the latest available observations for the releasePlan
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// ReleasePlanAdmission contains the namespaced name of the releasePlanAdmission this ReleasePlan is
+	// matched to
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?\/[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +optional
+	ReleasePlanAdmission string `json:"releasePlanAdmission,omitempty"`
```
davidmogar, line 66: With our current code we could still have multiple ones, but this looks as it should. Just leaving a note here ;)

api/v1alpha1/releaseplan_types.go
```diff
@@ -53,6 +55,15 @@ type ReleasePlanSpec struct {
 
 // ReleasePlanStatus defines the observed state of ReleasePlan.
 type ReleasePlanStatus struct {
+	// Conditions represent the latest available observations for the releasePlan
+	// +optional
+	Conditions []metav1.Condition `json:"conditions"`
+
+	// ReleasePlanAdmission contains the namespaced name of the releasePlanAdmission this ReleasePlan is
```
davidmogar, line 62: We don't really need the namespaced name as we have a target field, but that's ok to me.

api/v1alpha1/release_conditions.go
```diff
@@ -6,6 +6,10 @@ const (
 	// deployedConditionType is the type used to track the status of a Release deployment
 	deployedConditionType conditions.ConditionType = "Deployed"
 
+	// matchedConditionType is the type used to track the status of the ReleasePlan being matched to a
```
davidmogar, line 9: This should be in a releaseplan_conditions.go or releaseplanadmission_conditions.go file.

davidmogar (changes requested):
We could add a print column to indicate whether an RP or RPA matches any counterpart. In the case of RPs we could show the actual RPAs and in the case of RPAs we could show how many RPs are matched.

api/v1alpha1/releaseplan_types.go
johnbieren: I kind of just copied the style of how we store references to CRs in CRDs. I'll leave it as is unless you want me to change it, which I am okay with

api/v1alpha1/releaseplan_types.go
```diff
@@ -69,6 +80,31 @@ type ReleasePlan struct {
 	Status ReleasePlanStatus `json:"status,omitempty"`
 }
 
+// IsMatched checks whether the ReleasePlan is matched to a ReleasePlanAdmission.
```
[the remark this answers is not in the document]
johnbieren: To make the controller PR smaller and easier to review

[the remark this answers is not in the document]
davidmogar: It doesn't belong here. If it's needed by the controller, add it to the controller PR. This PR is about adding status fields, not logic or methods to support the new logic.

davidmogar (approved):

PR #297, opened 2023-11-08
fix(RHTAPBUGS-964): override pipelinerun timeout via RPA

johnbieren:
This commit adds a field named timeout to the PipelineRef struct. If it is provided in a ReleasePlanAdmission, it is passed to the release PipelienRun to override the default timeout value.

### DIFF
--- a/tekton/utils/pipeline_ref.go
+++ b/tekton/utils/pipeline_ref.go
@@ -26,6 +26,11 @@ type PipelineRef struct {
 
 	// Params is a slice of parameters for a given resolver
 	Params []Param `json:"params"`
+
+	// Timeout is value to use to override the tekton default Pipelinerun timeout
+	// +kubebuilder:default="0"
+	// +optional
+	Timeout string `json:"timeout,omitempty"`
 }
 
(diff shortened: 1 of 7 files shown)


ralphbean (approved):
LGTM, but someone else closer to the team should definitely review.

tekton/utils/pipeline_ref.go
```diff
@@ -26,6 +26,10 @@ type PipelineRef struct {
 
 	// Params is a slice of parameters for a given resolver
 	Params []Param `json:"params"`
+
+	// Timeout is value to use to override the tekton default Pipelinerun timeout
+	// +optional
```
davidmogar, line 31: Add `// +kubebuilder:default="0"` please.

davidmogar (changes requested):
All good. Just add the default and can go in.

tekton/utils/pipeline_ref.go
johnbieren: added

davidmogar (approved):

PR #311, opened 2023-11-30
feat(RHTAPREL-714): add Tekton utilities

davidmogar:
The utilities added in this commit come from the shared-utils repository. In addition, the RPA already references the new Pipeline struct.

Signed-off-by: David Moreno García <damoreno@redhat.com>

Depends on #310

PR #312, opened 2023-11-30
feat(RHTAPREL-716): add WithTimeouts method 

davidmogar:
The Pipeline struct now defines a Timeouts field that can be passed to the new WithTimeouts builder's method.

Depends on #310 and #311


tekton/utils/pipeline.go
```diff
@@ -48,10 +48,17 @@ type Pipeline struct {
 	// +optional
 	ServiceAccount string `json:"serviceAccountName,omitempty"`
 
-	// Timeout is value to use to override the tekton default Pipelinerun timeout
+	// Timeout is a value to use to override the tekton default Pipelinerun timeout
+	//
+	// This field is DEPRECATED and will be replaced by Timeouts in a future change.
```
johnbieren, line 53: Timeouts is added in this change, not a future change. Why are we keeping Timeout?

tekton/utils/suite_test.go
johnbieren, line 1: I am confused as to what change in this PR causes this to be added

tekton/utils/pipeline.go
davidmogar: Because this PR only adds the function but is not used. Timeout is still used here.

johnbieren (approved):

PR #323, opened 2023-12-11
feat(RHTAPREL-764): allow RPs to specify RPAs

johnbieren:
This commit changes the logic for fetching a ReleasePlanAdmission for a given ReleasePlan. If the ReleasePlan contains a label for the ReleasePlanAdmission to use, that ReleasePlanAdmission is searched for directly. This allows operations to continue even if multiple ReleasePlanAdmissions exist in a single managed namespace.


scoheb:
logic-wise lgtm. I'll let others approve.

davidmogar (approved):

PR #332, opened 2024-01-02
chore: enable source builds

scoheb:
- EC policy now requires them


johnbieren (approved):
Thanks. I thought https://github.com/redhat-appstudio/release-service/commit/60cf16914310918bfdeed446d722a17c10bf89ca was all we needed for this

PR #334, opened 2024-01-08
feat(RHTAPREL-790): pass taskGitRevision param to plr

johnbieren:
If a git resolver is used in the RPA, pass the revision value from it to the release pipelineRun as the taskGitRevision parameter.

### DIFF
--- a/controllers/release/adapter_test.go
+++ b/controllers/release/adapter_test.go
@@ -872,14 +872,26 @@ var _ = Describe("Release adapter", Ordered, func() {
 		})
 
 		It("references the pipeline specified in the ReleasePlanAdmission", func() {
-			var pipelineName string
+			var pipelineUrl string
 			resolverParams := pipelineRun.Spec.PipelineRef.ResolverRef.Params
 			for i := range resolverParams {
-				if resolverParams[i].Name == "name" {
-					pipelineName = resolverParams[i].Value.StringVal
+				if resolverParams[i].Name == "url" {
+					pipelineUrl = resolverParams[i].Value.StringVal
 				}
 			}
-			Expect(pipelineName).To(Equal(releasePlanAdmission.Spec.PipelineRef.Params[1].Value))
+			Expect(pipelineUrl).To(Equal(releasePlanAdmission.Spec.PipelineRef.Params[0].Value))
+		})
+
+		It("contains a parameter with the taskGitRevision", func() {
+			Expect(pipelineRun.Spec.Params).Should(ContainElement(HaveField("Name", "taskGitRevision")))
+			var revision string
+			resolverParams := pipelineRun.Spec.PipelineRef.ResolverRef.Params
+			for i := range resolverParams {
+				if resolverParams[i].Name == "revision" {
+					revision = resolverParams[i].Value.StringVal
+				}
+			}
+			Expect(pipelineRun.Spec.Params).Should(ContainElement(HaveField("Value.StringVal", revision)))
 		})
 
 		It("contains the proper timeout value", func() {
@@ -1689,11 +1701,11 @@ var _ = Describe("Release adapter", Ordered, func() {
 				Origin:       "default",
 				Environment:  environment.Name,
 				PipelineRef: &tektonutils.PipelineRef{
-					Resolver: "bundles",
+					Resolver: "git",
 					Params: []tektonutils.Param{
-						{Name: "bundle", Value: "quay.io/some/bundle"},
-						{Name: "name", Value: "release-pipeline"},
-						{Name: "kind", Value: "pipeline"},
+						{Name: "url", Value: "my-url"},
+						{Name: "revision", Value: "my-revision"},
+						{Name: "pathInRepo", Value: "my-path"},
 					},
 					Timeout: "2h0m0s",
 				},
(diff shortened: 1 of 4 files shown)


tekton/pipeline_run.go
```diff
@@ -163,6 +163,21 @@ func (r *ReleasePipelineRun) WithServiceAccount(serviceAccount string) *ReleaseP
 	return r
 }
 
+// WithTaskGitRevisionParameter adds the taskGitRevision parameter to the managed Release PipelineRun with the value of the revision
+// from the pipelineRef.
+func (r *ReleasePipelineRun) WithTaskGitRevisionParameter(pipelineRef *tektonv1.PipelineRef) *ReleasePipelineRun {
```
mmalina, line 168: Shouldn't we also check that the resolvername is "git" first? I understand that it's unlikely that there will be a "revision" used in another resolver (since we don't really use that many resolver types), but it seems like the right thing to do.

Also, what's the benefit of not working with our PipelineRef directly and calling `ToTektonPipelineRef` first? Is it for consistency with the other function? Any other reason?

johnbieren: Totally. Updated to add a check for `git` resolver first

It was just for consistency with the other function, but I changed it to use the PipelineRef directly

controllers/release/adapter_test.go
```diff
@@ -872,14 +872,26 @@ var _ = Describe("Release adapter", Ordered, func() {
...
-			Expect(pipelineName).To(Equal(releasePlanAdmission.Spec.PipelineRef.Params[1].Value))
+			Expect(pipelineUrl).To(Equal(releasePlanAdmission.Spec.PipelineRef.Params[0].Value))
+		})
+
+		It("contains a parameter with the taskGitRevision", func() {
+			Expect(pipelineRun.Spec.Params).Should(ContainElement(HaveField("Name", "taskGitRevision")))
+			var revision string
+			resolverParams := pipelineRun.Spec.PipelineRef.ResolverRef.Params
+			for i := range resolverParams {
+				if resolverParams[i].Name == "revision" {
```
mmalina, line 890: I was trying to figure out if the pipelinerun will actually have it. Can you help me find where it's defined? The ppr is created in the `BeforeEach` function, right? And I guess I need to look at the `releasePlanAdmission` referenced in resources. But where is that one defined? Is `loader_mock` involved here?

tekton/pipeline_run_test.go
```diff
@@ -23,6 +23,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/redhat-appstudio/release-service/tekton/utils"
```
mmalina, line 26: See the warning here - I guess you added this when you switched to using utils.PipelineRef directly, but in fact you can use the existing import?

tekton/pipeline_run.go
mmalina: Cool, thanks.

tekton/pipeline_run_test.go
johnbieren: Thanks, I thought VSCode would handle that for me

controllers/release/adapter_test.go
mmalina: Oh, now I see it - `createResources()` is called in the `BeforeAll` on line 73. So that's where it appears. Thanks for the explanation.

tekton/pipeline_run_test.go
```diff
@@ -212,6 +212,35 @@ var _ = Describe("PipelineRun", func() {
...
+			Expect(releasePipelineRun.Spec.Params[0].Name).To(Equal("taskGitRevision"))
+			Expect(releasePipelineRun.Spec.Params[0].Value.StringVal).To(Equal("my-revision"))
+		})
+
+		It("does not add the taskGitRevision parameter to the PipelineRun object when using a bundles resolver", func() {
+			pipelineRef := &tektonutils.PipelineRef{
+				Resolver: "bundles",
+				Params: []tektonutils.Param{
+					{
+						Name:  "revision",
```
mmalina, line 235: I don't really mind, but maybe it could a more realistic example of a bundle resolver (which typically doesn't have a "revision"). This wouldn't affect the code path taken anyway, since the first `if` that checks if the resolver is git will not pass. Just a thought.

For example:
```go
Params: []tektonutils.Param{
					{
						Name:  "bundle",
						Value: "my-bundle",
					},
					{
						Name:  "name",
						Value: "my-name",
					},
					{
						Name:  "kind",
						Value: "my-kind",
					},
				},
```

Or the params could even be empty.

tekton/pipeline_run_test.go
mmalina: I would have expected that too :)

tekton/pipeline_run_test.go
johnbieren: We're extracting the revision parameter, so my goal was to have a resolver with the revision parameter there to ensure it is not still extracted. Maybe I shouldn't have use `bundles` resolver here. Would it help if I changed it to like `Resolver: "my-custom-resolver"`?

mmalina: OK, makes sense.

> Would it help if I changed it to like Resolver: "my-custom-resolver"?

No, it's probably fine, thanks.

mmalina (approved):

PR #342, opened 2024-01-23
fix(RHTAPBUGS-1113): releasePlans missing Matched condition

johnbieren:
When a new ReleasePlan is created and has no matching ReleasePlanAdmission, the Matched condition should be added and be set to false. Before this commit, the condition would not be added.


johnbieren:
Blocked by https://github.com/redhat-appstudio/e2e-tests/pull/996

PR #343, opened 2024-01-23
fix(RHTAPREL-810): reorder operations in ReleasePlan controller

johnbieren:
This commit reorders the operations in the ReleasePlan controller so that the application owner reference is set last. If the application is not found, the reconcile loop errors. This means that if the application is not yet created, the other operations will not run, and (at this time) no matching information will be set. With this commit, the matching information is set before, as this does not rely on the application CR existing.

### DIFF
--- a/controllers/releaseplan/controller.go
+++ b/controllers/releaseplan/controller.go
@@ -63,8 +63,8 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	adapter := newAdapter(ctx, c.client, releasePlan, loader.NewLoader(), &logger)
 
 	return controller.ReconcileHandler([]controller.Operation{
-		adapter.EnsureOwnerReferenceIsSet,
 		adapter.EnsureMatchingInformationIsSet,
+		adapter.EnsureOwnerReferenceIsSet,
 	})
 }
 


johnbieren:
CI blocked by https://github.com/redhat-appstudio/e2e-tests/pull/996

mmalina:
Is there a bug in Jira for this? Or how was this discovered?

> This commit reorders the operations in the ReleasePlan controller so that the application owner reference is set last. If the application is not found, the reconcile loop errors. This means that if the application is not yet created, the other operations will not run, and (at this time) no matching information will be set. With this commit, the matching information is set before, as this does not rely on the application CR existing.

I was confused when reading this at first as I thought "If the application is not found, the reconcile loop errors." is the result of this fix. So I'd suggest to clarify it a bit, e.g. add "currently" or "until now" before the second sentence:

```
This commit reorders the operations in the ReleasePlan controller so that the application owner reference is set last.

Currently, if the application is not found, the reconcile loop errors. This means that if the application is not yet created, the other operations will not run, and (at this time) no matching information will be set.

With this commit, the matching information is set before, as this does not rely on the application CR existing.
```

controllers/releaseplan/controller.go
```diff
@@ -63,8 +63,8 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	adapter := newAdapter(ctx, c.client, releasePlan, loader.NewLoader(), &logger)
 
 	return controller.ReconcileHandler([]controller.Operation{
-		adapter.EnsureOwnerReferenceIsSet,
 		adapter.EnsureMatchingInformationIsSet,
+		adapter.EnsureOwnerReferenceIsSet,
```
mmalina, line 67: Just checking: This has no effect on tests? Is there a way we could add a test for the scenario we're fixing? If not, it's ok.

theflockers:
> Is there a bug in Jira for this? Or how was this discovered?
> 
> > This commit reorders the operations in the ReleasePlan controller so that the application owner reference is set last. If the application is not found, the reconcile loop errors. This means that if the application is not yet created, the other operations will not run, and (at this time) no matching information will be set. With this commit, the matching information is set before, as this does not rely on the application CR existing.
> 
> I was confused when reading this at first as I thought "If the application is not found, the reconcile loop errors." is the result of this fix. So I'd suggest to clarify it a bit, e.g. add "currently" or "until now" before the second sentence:
> 
> ```
> This commit reorders the operations in the ReleasePlan controller so that the application owner reference is set last.
> 
> Currently, if the application is not found, the reconcile loop errors. This means that if the application is not yet created, the other operations will not run, and (at this time) no matching information will be set.
> 
> With this commit, the matching information is set before, as this does not rely on the application CR existing.
> ```

+1 about the Jira. I think it is important for (eventually) future tracking.

johnbieren:
> > Is there a bug in Jira for this? Or how was this discovered?
> > > This commit reorders the operations in the ReleasePlan controller so that the application owner reference is set last. If the application is not found, the reconcile loop errors. This means that if the application is not yet created, the other operations will not run, and (at this time) no matching information will be set. With this commit, the matching information is set before, as this does not rely on the application CR existing.
> > 
> > 
> > I was confused when reading this at first as I thought "If the application is not found, the reconcile loop errors." is the result of this fix. So I'd suggest to clarify it a bit, e.g. add "currently" or "until now" before the second sentence:
> > ```
> > This commit reorders the operations in the ReleasePlan controller so that the application owner reference is set last.
> > 
> > Currently, if the application is not found, the reconcile loop errors. This means that if the application is not yet created, the other operations will not run, and (at this time) no matching information will be set.
> > 
> > With this commit, the matching information is set before, as this does not rely on the application CR existing.
> > ```
> 
> +1 about the Jira. I think it is important for (eventually) future tracking.

I can create a JIRA for it. It was found when I was working on RHTAPBUGS-1113

mmalina:
> I can create a JIRA for it. It was found when I was working on [RHTAPBUGS-1113](https://issues.redhat.com//browse/RHTAPBUGS-1113)

Yeah, so maybe it's enough to reference that bug here and add a note there that you also found this issue and address it in this PR? A separate bug would work as well of course.

controllers/releaseplan/controller.go
johnbieren: I'm not aware of any way to test this besides creating pretty involved tests per adapter that does everything the adapter does and ensures it happens in the right order. It would pretty much be recreating e2e

johnbieren:
> > I can create a JIRA for it. It was found when I was working on [RHTAPBUGS-1113](https://issues.redhat.com//browse/RHTAPBUGS-1113)
> 
> Yeah, so maybe it's enough to reference that bug here and add a note there that you also found this issue and address it in this PR? A separate bug would work as well of course.

Updated with https://issues.redhat.com/browse/RHTAPREL-810 reference

controllers/releaseplan/controller.go
mmalina: I see. It's fine then.

PR #344, opened 2024-01-25
feat(RHTAPREL-795): remove Releases after a given period of time

theflockers:
This PR add the `ReleaseGracePeriodDays` (defaults to 7 days) parameter to ReleasePlan 
and add the `ExpirationTime` to theRelease. The ReleaseGracePeriodDays then is used to
set the ExpirationTime of a Release.


theflockers:
Example Cronjob/script for cleaning up the Expired Release:

The script works, however the cronjob itself is not tested yet.

```
apiVersion: batch/v1
kind: CronJob
metadata:
  name: ExpiredReleasesCleanUp
  namespace: release-service
spec:
  schedule: "0 3 * * *"
  successfulJobsHistoryLimit: 1
  failedJobsHistoryLimit: 1
  jobTemplate:
    spec:
      template:
        spec:
          containers:
            - name: release-cleanup
              env:
                - name: EXPIRED_RELEASES_FILE
                  value: /tmp/releases-to-be-deleted
                - name: MAX_PROCS
                  value: 5
              image: quay.io/redhat-appstudio/release-service-utils:305541d8b8c2670dea4b50bd8c56858c365ca11e
                #
                NOW=`date +%s`

                oc get release --all-namespaces \
                --sort-by=.status.expirationTime \
                --template '{{range .items}}{{.metadata.name}}{{"\t"}}{{.metadata.namespace}}{{"\t"}}{{.status.expirationTime}}{{"\n"}}{{end}}' \
                | awk  -v now=$NOW \
                '{
                   # parsing the expirationTime and converting it to epoch
                   # so we can calculate easier the expired Releases
                   gsub("[:\\-TZ]", " ", $3)
                   t=mktime($3)
                   expirationTime=strftime("%s", t)
                
                   # expirationTime should be smaller than today in seconds
                   if(now < expirationTime) {
                     args="%s:%s"
                     printf(args, $1, $2)
                   }
                }' > $EXPIRED_RELEASES_FILE
                
                
                # The deleteAndLog will run the Release deletion and save the operation in a structured way that
                # can be read easily by kubectl or journalctl
                function deleteAndLog() {
                  release=${1%:*}
                  namespace=${1#*:}
                  oc delete release $release -n $namespace |while read logLine; do
                    logger -i --id $release $logLine
                  done
                }
                export -f deleteAndLog
                
                # using xargs to enable parallel deletion
                xargs -a $EXPIRED_RELEASES_FILE -i -P $MAX_PROCS bash -c 'deleteAndLog "{}"'
```

theflockers:
Noticed I missed one item of the accepting criteria. Moving back to draft to work on it.

api/v1alpha1/webhooks/releaseplan/webhook_test.go
```diff
@@ -131,4 +131,19 @@ var _ = Describe("ReleasePlan webhook", func() {
 			Expect(webhook.ValidateDelete(ctx, releasePlan)).To(BeNil())
 		})
 	})
+
+	When("a ReleasePlan is created without ReleaseGracePeriodDays value", func() {
+		It("should set the default value for ReleaseGracePeriodDats to metadata.DefaultReleaseGracePeriodDays", func() {
```
johnbieren, line 136: typo Dats

controllers/release/adapter_test.go
```diff
@@ -804,6 +805,55 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+		})
+
+		It("should set the ExpirationTime and continue", func() {
+			result, err := adapter.EnsureReleaseExpirationTimeIsAdded()
+			Expect(adapter.release.Status.ExpirationTime).To(Equal(expectedExpirationTime))
+			Expect(!result.RequeueRequest && !result.CancelRequest).To(BeTrue())
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("should set continue if the ExpirationTime is already set", func() {
```
johnbieren, line 848: I think this test would be better if the release.Status.ExpirationTime is set to something else at the start (line 849) instead. The way it is written, you manually set it to the default, then call the operation which sets it to the default anyways, so you can't really be sure if it was overwritten

controllers/releaseplan/adapter.go
```diff
@@ -102,3 +103,21 @@ func (a *adapter) EnsureMatchingInformationIsSet() (controller.OperationResult,
 
 	return controller.RequeueOnErrorOrContinue(a.client.Status().Patch(a.ctx, a.releasePlan, patch))
 }
+
+// EnsureReleaseGracePeriodDaysIsSet is an operation that will ensure that the ReleasePlan has the
+// ReleaseGracePeriodDays set
+func (a *adapter) EnsureReleaseGracePeriodDaysIsSet() (controller.OperationResult, error) {
```
johnbieren, line 109: I'm not sure what we need this for since this is already done in the webhook

metadata/constants.go
```diff
@@ -0,0 +1,23 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metadata
+
+// Common constants
+const (
+	// DefaultReleaseGracePeriodDays is the default time to live in days of a Release
+	DefaultReleaseGracePeriodDays = 7
```
johnbieren, line 22: @davidmogar do we want this here? Is it going to be in the ReleaseServiceConfig? Can we just use kubebuilder defaults instead of having the constant?

api/v1alpha1/webhooks/releaseplan/webhook.go
```diff
@@ -46,6 +47,10 @@ func (w *Webhook) Default(ctx context.Context, obj runtime.Object) error {
 		}
 	}
 
+	if releasePlan.Spec.ReleaseGracePeriodDays == 0 {
```
johnbieren, line 50: So if you omit the field, it equals 0? If so, that is a little unfortunate... it would be nice if setting it to 0 was a real possibility when a user wants their Release CRs to never expire

api/v1alpha1/releaseplan_types.go
```diff
@@ -51,6 +51,11 @@ type ReleasePlanSpec struct {
 	// +optional
 	ServiceAccount string `json:"serviceAccount,omitempty"`
 
+	// ReleaseGracePeriodDays is the number of days a Release should be kept
+	// This value is used to define the Release ExpirationTime
```
johnbieren, line 55: https://book.kubebuilder.io/reference/markers/crd-validation is what I meant with my comment on the constants file. Maybe we can just add the kubebuilder annotation here and not need the webhook change (or constants file)

controllers/releaseplan/adapter.go
theflockers: Maybe I did something wrong, but In my test the value was added the on creation through the webhook but after reconciling it lost the value. I will test once again without it.

api/v1alpha1/releaseplan_types.go
theflockers: Interesting. 
```
 +kubebuilder:default any
sets the default value for this field.
```

I will try it.

theflockers: The good thing about having the constant is that it can be re-used eventually. But not 100% sure that we would re-use it (besides tests).

api/v1alpha1/webhooks/releaseplan/webhook.go
theflockers: Yes. I initially expected it would be null, but it is indeed zero.

metadata/constants.go
davidmogar: No, it shouldn't be here. As you said, and just to give some more context to Leandro, defaults will be moved in the future (different issue) to the ReleaseServiceConfig resource. For the time being, using a kubebuilder annotation is a much better and cleaner solution. You can find it [here](https://book.kubebuilder.io/reference/markers/crd-validation) in case you didn't know about it.

api/v1alpha1/webhooks/releaseplan/webhook.go
davidmogar: That's always the case for `int`. I know other team "fixed" this by using `*int`, but I don't really like that solution.

api/v1alpha1/webhooks/releaseplan/webhook_test.go
theflockers: Fixed.

api/v1alpha1/releaseplan_types.go
theflockers: Changed to set the default through `kubebuilder`

controllers/release/adapter_test.go
theflockers: fixed.

controllers/releaseplan/adapter.go
theflockers: Removed.

metadata/constants.go
theflockers: file removed.

api/v1alpha1/webhooks/release/webhook.go
```diff
@@ -34,6 +37,29 @@ type Webhook struct {
...
+
+	if release.Spec.GracePeriodDays == 0 {
+		releasePlan := &v1alpha1.ReleasePlan{}
+
+		err := w.client.Get(ctx, types.NamespacedName{
+			Name:      release.Spec.ReleasePlan,
+			Namespace: release.Namespace,
+		}, releasePlan)
+
+		if err != nil && errors.IsNotFound(err) {
```
theflockers, line 51: although this situation would not happen, I think it is better to avoid some eventual panic. Not sure if this is the best approach. Thoughts?

api/v1alpha1/webhooks/release/webhook_test.go
```diff
@@ -84,4 +101,16 @@ var _ = Describe("Release validation webhook", func() {
 			Expect(webhook.ValidateDelete(ctx, release)).To(BeNil())
 		})
 	})
+
+	When("Default method is called", func() {
+		It("should set GracePeriodDays to ReleasePlan's value and return nil", func() {
+			ctx := context.Background()
+
+			Expect(k8sClient.Create(ctx, releasePlan)).Should(Succeed())
```
johnbieren, line 109: This needs to be deleted after the test

api/v1alpha1/webhooks/release/webhook_test.go
```diff
@@ -84,4 +101,16 @@ var _ = Describe("Release validation webhook", func() {
...
+			Expect(k8sClient.Create(ctx, release)).Should(Succeed())
+
+			Expect(webhook.Default(ctx, release)).To(BeNil())
+			Expect(release.Spec.GracePeriodDays).To(Equal(7))
```
johnbieren, line 113: Might be nice to replace 7 with releasePlan.Spec.ReleaseGracePeriodDays

api/v1alpha1/webhooks/release/webhook.go
johnbieren: I think this situation could actually happen (just not in the automated case). We already have people creating ReleasePlans for nonexistent applications :sweat_smile: 
In fact, I'd like for you to add a test for this situation too

api/v1alpha1/webhooks/release/webhook.go
```diff
@@ -34,6 +37,29 @@ type Webhook struct {
...
+			w.log.Info("releasePlan not found to get ReleaseGracePeriodDays")
```
johnbieren, line 52: Maybe add like "Not setting release graceperioddays

johnbieren (changes requested):

api/v1alpha1/webhooks/release/webhook_test.go
theflockers: fixed.

controllers/release/adapter.go
```diff
@@ -278,6 +279,27 @@ func (a *adapter) EnsureReleaseIsProcessed() (controller.OperationResult, error)
...
 }
 
+// EnsureReleaseExpirationTimeIsAdded is an operation that ensures that a Release has the ExpirationTime set.
+func (a *adapter) EnsureReleaseExpirationTimeIsAdded() (controller.OperationResult, error) {
+	releasePlan, err := a.loader.GetReleasePlan(a.ctx, a.client, a.release)
+	if err != nil && !errors.IsNotFound(err) {
+		return controller.RequeueWithError(err)
+	}
+
+	if a.release.Status.ExpirationTime == nil {
```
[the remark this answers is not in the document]
theflockers: makes sense. Done.

api/v1alpha1/webhooks/releaseplan/webhook.go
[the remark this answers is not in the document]
theflockers: indeed. fixed.

controllers/release/adapter_test.go
```diff
@@ -804,6 +805,70 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+			expectedExpirationTime := &metav1.Time{Time: creationTime.Add(time.Hour * 24 * expireDays)}
+
+			result, err := adapter.EnsureReleaseExpirationTimeIsAdded()
+			Expect(adapter.release.Status.ExpirationTime).To(Equal(expectedExpirationTime))
+			Expect(!result.RequeueRequest && !result.CancelRequest).To(BeTrue())
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("should set the ExpirationTime with the value of ReleasePlan's ReleaseGracePeriodDays and then continue", func() {
+			expireDays := time.Duration(6)
```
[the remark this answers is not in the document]
theflockers: yes. updated.

controllers/release/adapter_test.go
```diff
@@ -804,6 +805,70 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+			creationTime := adapter.release.CreationTimestamp
+			expectedExpirationTime := &metav1.Time{Time: creationTime.Add(time.Hour * 24 * expireDays)}
+
+			result, err := adapter.EnsureReleaseExpirationTimeIsAdded()
+			Expect(adapter.release.Status.ExpirationTime).To(Equal(expectedExpirationTime))
+			Expect(!result.RequeueRequest && !result.CancelRequest).To(BeTrue())
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		It("should continue if the ExpirationTime is already set", func() {
```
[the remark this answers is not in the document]
theflockers: done.

api/v1alpha1/webhooks/release/suite_test.go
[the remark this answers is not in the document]
theflockers: Yes. removed.

api/v1alpha1/webhooks/release/webhook_test.go
theflockers: done.

api/v1alpha1/webhooks/release/webhook.go
theflockers: done.

theflockers: * the tests will come in the next commit.

api/v1alpha1/webhooks/release/webhook.go
theflockers: I added one extra test just to make sure the default value for GracePeriodDays actually is 0 when the releasePlan does not exist. Turns out I was redundant by calling "Default" as it is done implicitly every time a release is created, so I just need to check it the results are the ones expected. Please see if you think these tests are enough.

api/v1alpha1/webhooks/release/suite_test.go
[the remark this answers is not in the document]
johnbieren: This still has changes

controllers/release/suite_test.go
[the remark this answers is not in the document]
johnbieren: Still stands

api/v1alpha1/webhooks/release/webhook.go
johnbieren: I think it is sufficient, yep. Thanks!

johnbieren (commented):
There are two more files I think are leftover from old stuff and shouldn't be part of th diff, but otherwise lgtm

api/v1alpha1/webhooks/release/suite_test.go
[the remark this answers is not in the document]
theflockers: I think this comment was about the releaseplan webhook file. This one Is actually required. Otherwise the webhook for release is not executed. If I dont have it loaded, it causes this: 

```
<*errors.StatusError | 0xc0004c8fa0>: 
      Internal error occurred: failed calling webhook "mreleaseplan.kb.io": failed to call webhook: the server could not find the requested resource

```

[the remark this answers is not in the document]
johnbieren: Oh, I guess it is because you create a ReleasePlan in the Release webhook tests now

johnbieren (approved):

PR #366, opened 2024-02-19
feat(RHTAPREL-253): queue Releases

davidmogar:


controllers/release/adapter.go
```diff
@@ -251,6 +251,22 @@ func (a *adapter) EnsureReleaseIsProcessed() (controller.OperationResult, error)
...
 
+	pipelineRuns, err := a.loader.GetActiveManagedReleasePipelineRuns(a.ctx, a.client, a.release)
+	if err != nil {
+		return controller.RequeueWithError(err)
+	}
+
+	// Requeue de Release if other PipelineRun is running for the same Application
+	if len(pipelineRuns.Items) > 0 {
+		patch := client.MergeFrom(a.release.DeepCopy())
+		a.release.MarkReleaseQueued(fmt.Sprintf("%d Release(s) running for the same Application", len(pipelineRuns.Items)))
```
mmalina, line 262: What's running is not releases, but pipelineruns. Maybe "Release PipelineRuns" would work?

johnbieren (commented):
I like the approach. I reviewed most but not all of the code

api/v1alpha1/release_types.go
```diff
@@ -526,7 +541,8 @@ func (r *Release) hasPhaseFinished(conditionType conditions.ConditionType) bool
 	case condition.Status == metav1.ConditionTrue:
 		return true
 	default:
-		return condition.Status == metav1.ConditionFalse && condition.Reason != ProgressingReason.String()
+		return condition.Status == metav1.ConditionFalse && condition.Reason != ProgressingReason.String() &&
```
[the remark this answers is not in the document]
davidmogar: I don't think so. There the reason is already `Progressing`.

api/v1alpha1/release_types.go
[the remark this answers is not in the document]
davidmogar: :man_shrugging:

loader/loader.go
```diff
@@ -44,6 +45,35 @@ func NewLoader() ObjectLoader {
 	return &loader{}
 }
 
+// GetActiveManagedReleasePipelineRuns returns all active managed Release PipelineRuns for the Application being Released.
+// PipelineRuns for the Release passed as an argument are ignored.
+func (l *loader) GetActiveManagedReleasePipelineRuns(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*tektonv1.PipelineRunList, error) {
```
johnbieren, line 50: Can we use unfinished or something instead of active? We already have active for the RP/RPA stuff

controllers/release/adapter.go
```diff
@@ -187,6 +187,22 @@ func (a *adapter) EnsureReleaseIsProcessed() (controller.OperationResult, error)
...
+		patch := client.MergeFrom(a.release.DeepCopy())
+		a.release.MarkReleaseQueued(fmt.Sprintf("%d Release PipelineRun(s) running for the same Application", len(pipelineRuns.Items)))
+		err := a.client.Status().Patch(a.ctx, a.release, patch)
+		if err != nil {
+			return controller.RequeueWithError(err)
+		}
+		return controller.RequeueAfter(time.Minute, nil)
+	}
+
 	pipelineRun, err := a.loader.GetManagedReleasePipelineRun(a.ctx, a.client, a.release)
```
johnbieren, line 206: Won't pipelineRuns.length be 1 if there is already a pipelineRun running for this release? So we don't have to fetch it again?

PR #367, opened 2024-02-19
feat(RHTAPREL-815): pass RSC to managed pipelinerun

johnbieren:
Depends on https://github.com/redhat-appstudio/release-service/pull/313


controllers/release/adapter_test.go
```diff
@@ -149,6 +154,8 @@ var _ = Describe("Release adapter", Ordered, func() {
 					},
 				},
 			})
+			adapter.releaseServiceConfig = releaseServiceConfig
```
johnbieren, line 157: @davidmogar thoughts on these lines? I could change `createReleaseAndAdapter` to do this instead of doing it manually in the tests that require it

PR #374, opened 2024-02-20
feat(RHTAPREL-371): create roleBinding for managed pipelineRun

johnbieren:
Create a RoleBinding that grants the serviceAccount that is used to execute the managed PipelineRun permissions in the origin workspace so that it can read the required CRs.

### DIFF
--- a/controllers/release/adapter.go
+++ b/controllers/release/adapter.go
@@ -35,6 +35,7 @@ import (
 
 	libhandler "github.com/operator-framework/operator-lib/handler"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	rbac "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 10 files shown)


johnbieren:
/test release-service-e2e
No clusters available to be claimed

theflockers (approved):

controllers/release/adapter.go
```diff
@@ -256,13 +257,25 @@ func (a *adapter) EnsureReleaseIsProcessed() (controller.OperationResult, error)
 		return controller.RequeueWithError(err)
 	}
 
+	// Don't check for error as it is expected if no RoleBinding exists yet
+	roleBinding, _ := a.loader.GetRoleBindingFromReleaseStatus(a.ctx, a.client, a.release)
```
davidmogar, line 261: Then use `errors.IsNotFound(err)`.

johnbieren: That won't be the error type if the rolebinding field in the release status is empty

PR #388, opened 2024-02-29
fix: modify gosec PR check to log results

mmalina:
A while ago we modified the workflow to save the results in a SARIF file and then upload to Github to show issues in the UI, but somehow it does not seem to work right now and there is no way to see what's wrong.

This change adds a second run of gosec which will
print the issues in the log directly.

This is related to https://github.com/redhat-appstudio/release-service/pull/363 where we were repeatedly getting failures for gosec, but we couldn't see what the issue was - we just saw that it failed, but there was nothing in the UI. With this change, we can now see the error:

```
Golang errors in file: []:

  > [line 0 : column 0] - error obtaining VCS status: exit status 128
	Use -buildvcs=false to disable VCS stamping.
```

Likely because the error is weird and not related to a specific file, nothing is displayed in Github UI.

### DIFF
--- a/.github/workflows/gosec.yaml
+++ b/.github/workflows/gosec.yaml
@@ -23,6 +23,14 @@ jobs:
         id: gosec
         uses: securego/gosec@master
         with:
+          args: -exclude-generated ./...
+        env:
+          GOROOT: ""
+        continue-on-error: true
+      - name: Run Gosec Security Scanner - save SARIF
+        id: gosec-sarif
+        uses: securego/gosec@master
+        with:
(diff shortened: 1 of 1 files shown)


loader/loader.go
johnbieren, line 1: I am guessing this change was included accidentally by your IDE

johnbieren (dismissed):

loader/loader.go
mmalina: Not really. I needed to a go file change to trigger the gosec workflow. But yeah, we don't want to commit this. Removed.

johnbieren (approved):

PR #397, opened 2024-03-05
fix: move default RSC to manager dir

johnbieren:
The manifests dir isn't included in the default kustomization but manager is, so move default RSC there.


mmalina (dismissed):

happybhati (approved):
/lgtm

PR #411, opened 2024-03-19
chore: add `name` and `com.redhat.component` labels

happybhati:
We need to add the `name` and `com.redhat.component` labels to avoid an EC violation.


johnbieren (commented):
Can you `s/RHTAP/Konflux/g`? Then lgtm

johnbieren (approved):

PR #421, opened 2024-03-28
feat(RHTAPREL-783): do not requeue automated errors

davidmogar:
Our manager logs were being cluttered with automated validations errors as those were requeued forever. Now, if the error appears after 5 minutes of the release being created (time enough for the Integration service to set the Release as automated) the error is reported and the Release marked as failed.

### DIFF
--- a/controllers/release/adapter_test.go
+++ b/controllers/release/adapter_test.go
@@ -1236,16 +1236,40 @@ var _ = Describe("Release adapter", Ordered, func() {
 		})
 
 		When("the release has the automated label", func() {
+			AfterEach(func() {
+				_ = adapter.client.Delete(ctx, adapter.release)
+			})
+
 			BeforeEach(func() {
+				adapter = createReleaseAndAdapter()
 				adapter.release.Labels = map[string]string{
 					metadata.AutomatedLabel: "true",
 				}
+				adapter.ctx = toolkit.GetMockedContext(ctx, []toolkit.MockData{
+					{
+						ContextKey: loader.ReleasePlanContextKey,
+						Resource:   releasePlan,
+					},
+				})
 			})
 
 			It("returns invalid and an error if automated label is present but is not set in release status", func() {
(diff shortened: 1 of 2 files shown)


controllers/release/adapter_test.go
```diff
@@ -1236,13 +1236,24 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+				adapter.ctx = toolkit.GetMockedContext(ctx, []toolkit.MockData{
+					{
+						ContextKey: loader.ReleasePlanContextKey,
+						Resource:   releasePlan,
+					},
+				})
 			})
 
-			It("returns invalid and an error if automated label is present but is not set in release status", func() {
+			It("returns invalid if automated label is present but is not set in release status", func() {
```
johnbieren, line 1256: why change this string? It returns `&controller.ValidationResult{Err: err}` so it has an error

controllers/release/adapter_test.go
```diff
@@ -1254,6 +1265,19 @@ var _ = Describe("Release adapter", Ordered, func() {
 				Expect(conditionMsg).To(Equal("automated not set in status for automated release"))
 			})
 
+			It("returns invalid and an error if the error appears after 5 minutes of being created", func() {
+				adapter.release.SetCreationTimestamp(metav1.Time{Time: time.Now().Add(-7 * time.Minute)})
+				result := adapter.validateAuthor()
+				Expect(result.Valid).To(BeFalse())
+				Expect(result.Err).NotTo(HaveOccurred())
```
johnbieren, line 1272: Right here we check the error did not occur, but the test description says `and an error`

johnbieren (changes requested):
Test names seem wrong, but code seems good

controllers/release/adapter_test.go
davidmogar: Yeah, my bad.

johnbieren (approved):

PR #422, opened 2024-04-01
fix(RHTAPREL-854): add processing cleanup operation to Release adapter

johnbieren:
The Release Processing resources were cleaned up as part of the operation that tracks the Processing status and if an error occurred cleaning them up, it was not retried. This commit breaks the cleanup out into its own operation that will run once the Release is marked as Processed and properly retries if an error occurs during cleanup.

### DIFF
--- a/controllers/release/adapter.go
+++ b/controllers/release/adapter.go
@@ -288,28 +288,48 @@ func (a *adapter) EnsureReleaseProcessingIsTracked() (controller.OperationResult
 		}
 	}
 
-	// This condition can only be true if the call to registerProcessingStatus changed the state
-	if a.release.HasProcessingFinished() {
-		// At this point, the PipelineRun has finished, so it's safe to remove the finalizer and RoleBinding
-		roleBinding, err := a.loader.GetRoleBindingFromReleaseStatus(a.ctx, a.client, a.release)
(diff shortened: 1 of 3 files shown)


controllers/release/adapter.go
```diff
@@ -288,28 +288,48 @@ func (a *adapter) EnsureReleaseProcessingIsTracked() (controller.OperationResult
...
+	if err != nil && !errors.IsNotFound(err) {
+		return controller.RequeueWithError(err)
+	}
+
+	roleBinding, err := a.loader.GetRoleBindingFromReleaseStatus(a.ctx, a.client, a.release)
+	if err != nil && !errors.IsNotFound(err) && !strings.Contains(err.Error(), "valid reference to a RoleBinding") {
+		return controller.RequeueWithError(err)
+	}
+
+	return controller.RequeueOnErrorOrContinue(a.cleanupManagedPipelineRunResources(pipelineRun, roleBinding))
```
davidmogar, line 311: Would you mind to use a name similar to the operation name? `cleanUpProcessingResources` for example?

johnbieren: Much nicer, great idea. Updated

davidmogar (approved):

happybhati (approved):

happybhati (approved):

PR #451, opened 2024-05-02
feat(RELEASE-119): add support  to run a tenant PipelineRun

theflockers:
This commit adds support for running tenant PipelineRuns


metadata/labels.go
```diff
@@ -62,6 +62,9 @@ var (
 	// ManagedPipelineType is the value to be used in the PipelinesTypeLabel for managed Pipelines
 	ManagedPipelineType = "managed"
 
+	// TenantPipelineType is the value to be used in the PipelinesTypeLabel for managed Pipelines
```
johnbieren, line 65: `s/managed/tenant/`

loader/loader.go
```diff
@@ -197,9 +205,9 @@ func (l *loader) GetRoleBindingFromReleaseStatus(ctx context.Context, cli client
 	return roleBinding, nil
 }
 
-// GetManagedReleasePipelineRun returns the managed Release PipelineRun referenced by the given Release or nil if it's not found. In the case
+// GetReleasePipelineRun returns the managed Release PipelineRun referenced by the given Release or nil if it's not found. In the case
```
johnbieren, line 208: probably remove the word managed in this line

johnbieren (commented):
@davidmogar is this how we want it? I was under the understanding that the tenant pipeline would execute before, not in place of, the managed pipeline

metadata/labels.go
theflockers: fixed

theflockers:
Moving it back to draft as changes are required.

api/v1alpha1/releaseplan_types.go
```diff
@@ -58,8 +51,7 @@ type ReleasePlanSpec struct {
 	ReleaseGracePeriodDays int `json:"releaseGracePeriodDays,omitempty"`
 
 	// Target references where to send the release requests
-	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
```
johnbieren, line 61: Can you not have a kubebuilder annotation if it is optional? It looks like you can, so I think we should keep it

tekton/utils/pipeline.go
```diff
@@ -57,7 +57,8 @@ type Pipeline struct {
 // the Pipeline.
 // +kubebuilder:object:generate=true
 type ParameterizedPipeline struct {
-	Pipeline
+	Pipeline `json:",inline"`
```
johnbieren, line 60: what does this change?

tekton/utils/pipeline_run_builder.go
```diff
@@ -111,6 +112,22 @@ func (b *PipelineRunBuilder) WithObjectReferences(objects ...client.Object) *Pip
 	return b
 }
 
+// WithTenantArrayPrams constructs tektonv1.Param entries for each of the provided client.Objects.
+// Each param name is modified to have the prefix "tenant".
```
johnbieren, line 116: Why? These are for tenant defined pipelines. Why would we modify the parameters they pass to their own pipelines, so they have to modify their pipeline definitions to fit?

controllers/release/adapter.go
```diff
@@ -203,22 +205,35 @@ func (a *adapter) EnsureReleaseIsProcessed() (controller.OperationResult, error)
...
+					// This string should probably be a constant somewhere
+					roleBinding, err = a.createRoleBindingForClusterRole("release-pipeline-resource-role", resources.ReleasePlanAdmission)
+					if err != nil {
+						return controller.RequeueWithError(err)
+					}
 				}
 			}
 
-			pipelineRun, err = a.createManagedPipelineRun(resources)
+			if len(resources.ReleasePlan.Spec.Target) == 0 {
```
johnbieren, line 224: I think a better check here is if ReleasePlan has a pipeline in it

controllers/release/adapter.go
```diff
@@ -373,6 +388,45 @@ func (a *adapter) createManagedPipelineRun(resources *loader.ProcessingResources
 	return pipelineRun, nil
 }
 
+// createTenantPipelineRun creates and returns a new tenant Release PipelineRun. The new PipelineRun will include owner
+// annotations, so it triggers Release reconciles whenever it changes. The Pipeline information and the parameters to it
+// will be extracted from the given ReleaseStrategy. The Release's Snapshot will also be passed to the release
```
johnbieren, line 393: We don't have ReleaseStrategy anymore

controllers/releaseplan/adapter.go
```diff
@@ -87,6 +87,11 @@ func (a *adapter) EnsureOwnerReferenceIsSet() (controller.OperationResult, error
 // EnsureMatchingInformationIsSet is an operation that will ensure that the ReleasePlan has updated matching
 // information in its status.
 func (a *adapter) EnsureMatchingInformationIsSet() (controller.OperationResult, error) {
+	// If the releasePlan has no Target set, no ReleasePlanAdmission will be loaded
```
johnbieren, line 90: I don't think you need this change. The releasePlanAdmission will be nil on line 91/96, so then we will set it unmatched already

loader/loader.go
```diff
@@ -248,26 +256,31 @@ type ProcessingResources struct {
...
 	resources.ReleasePlan, err = l.GetReleasePlan(ctx, cli, release)
 	if err != nil {
 		return resources, err
 	}
 
-	resources.ReleasePlanAdmission, err = l.GetActiveReleasePlanAdmissionFromRelease(ctx, cli, release)
-	if err != nil {
-		return resources, err
-	}
+	// Will only load the ReleasePlanAdmission, EnterpriseContractConfigMap and EnterpriseContractPolicy if ReleasePlan.Spec.Target
```
johnbieren, line 267: This function gets 5 resources. Conditionally ignoring 3 of them seems weird to me. I think it may be better to rename this to getmanagedprocessingresources and then create one for the tenant pipelines

johnbieren: Honestly, for just two resources I am not even sure we need a composite function

tekton/utils/pipeline.go
theflockers: without this tag, it the struct does not _inherit_ the fields from the extending struct.

tekton/utils/pipeline_run_builder.go
theflockers: From my understanding, the user could somehow also add some extra parameters in the releasePlan itself. I also added the `tenant` prefix as the parameter name might collide with some previous existent one, *release* for example. This is up for discussion - maybe this is not required at the moment, or some misunderstanding from my side. Can you comment here @davidmogar ?

controllers/release/adapter.go
theflockers: I was actually doing this at first. I switched to the target because it would not make sense having a target for a tenant pipeline, and I ended up using this check as default, but if you think it is more readable using `pipeline` instead of target, also fine.

In my understanding, it only makes sense having target for set for managed pipelines. Could you comment on this statement? Maybe this is a  wrong assumption.

controllers/release/adapter.go
theflockers: copy&paste. This is probably there also for the managed one. I will remove from both.

controllers/releaseplan/adapter.go
theflockers: I will crosscheck, but I added this here because the controller was panic-ing on `GetMatchingReleasePlanAdmission` as the releasePlanAdmission won't be passed for tenant pipelines.

loader/loader.go
theflockers: Good point. I will think what might be better - and handy.

tekton/utils/pipeline.go
johnbieren: Not entirely sure what that means, but if you are sure we need it, sounds good :slightly_smiling_face:

controllers/release/adapter.go
johnbieren: I am not 100% on this, but TMBK this is how things are going to go:
- We add support for tenant pipelines (this PR). You can run either a managed pipeline or a tenant pipeline in your release, but not both
- We allow users to run both. A collector pipeline is pretty much just a specific tenant pipeline

So I don't want to pollute this PR too much with things for supporting both, but for things like this where the two approaches are just about the same, might as well go with the future proof one IMO

controllers/release/adapter.go
johnbieren: Thanks!

controllers/releaseplan/adapter.go
johnbieren: Yeah, please crosscheck because we have an issue if so. We should hit this line https://github.com/konflux-ci/release-service/blob/main/loader/loader.go#L122 with the error being a IsNotFound error so we should return `nil, err` (or even if it returns no err but an empty list on line 119, the loop on line 128 would be skipped and we would hit line 142)

api/v1alpha1/releaseplan_types.go
theflockers: FYI I had to change this a little bit to allow setting an empty value or a input that matches the regex. When `Target` is missing in the releasePlan, it can not be created.

The ReleasePlan "tenant-pipeline-rp" is invalid: spec.target: Invalid value: "": spec.target in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?$'
```

File: tekton/utils/pipeline_run_builder.go (hunk 5 above)
Comment (theflockers) [reply tekton/utils/pipeline_run_builder.go]: I am for now removing the _tenant_ prefix as what you said makes sense - although we might need to have some check to display a clear error message in case the user adds a parameters that collides with a existent one. Anyhow, lets hear from David tomorrow.

Comment (johnbieren) [reply tekton/utils/pipeline_run_builder.go]: Said check can always come in a future PR once a user hits it and raises it

File: api/v1alpha1/releaseplan_types.go (hunk 3 above)
Comment (johnbieren) [reply api/v1alpha1/releaseplan_types.go:61]: It is interesting because serviceAccount is also optional and had the exact same kubebuilder validation as target

File: controllers/releaseplan/adapter.go (hunk 8 above)
Comment (theflockers) [reply controllers/releaseplan/adapter.go:90]: Adding it again, turns out the problem was not `panic` (it probably was fixed in another place), but anyhow without this line, I get the following error, and the pipeline is not created:

```
2024-05-21T16:59:55.886Z	ERROR	Reconciler error	{"controller": "release", "controllerGroup": "appstudio.redhat.com", "controllerKind": "Release", "Release": {"name":"tenant-pipeline-c7x4z","namespace":"dev-release-team-tenant"}, "namespace": "dev-release-team-tenant", "name": "tenant-pipeline-c7x4z", "reconcileID": "4d625aed-a649-4df2-8de6-97892e2bbb21", "error": "no ReleasePlanAdmission found in namespace (dev-release-team-tenant) with the origin (dev-release-team-tenant) for application 'tenant-pipeline'"}
```
The options would be keeping the code to continue processing or change the logic below it to save the unmatched information, although I think this would not make much sense for tenant pipelines. WDY?

Comment (johnbieren) [reply controllers/releaseplan/adapter.go:90]: This error is not from this controller. You can see `{"controller": "release", "controllerGroup": 
"appstudio.redhat.com", "controllerKind": "Release",` it isn't the ReleasePlan controller

Comment (theflockers) [reply controllers/releaseplan/adapter.go:90]: I noticed that. I was wondering if the identification of the logging for any reason could be wrong, because it is intriguing that it only breaks if I don't bypass specifically this block.

**Review by davidmogar — CHANGES_REQUESTED** [review]

File: controllers/release/adapter.go [hunk 10]
```diff
@@ -203,22 +206,46 @@ func (a *adapter) EnsureReleaseIsProcessed() (controller.OperationResult, error)
 			return controller.RequeueWithError(err)
 		}
 
+		if len(resources.ReleasePlan.Spec.Target) == 0 && resources.ReleasePlan.Spec.Pipeline == nil {
```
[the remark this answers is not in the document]
theflockers: Ok. I just don't like to compare with `""`. So I normally check the string length. But ok.

controllers/release/adapter.go
theflockers: I did not get which operation? the `EnsureReleaseIsProcessed`?  I did not get if you want this change or not.

controllers/release/adapter_test.go
```diff
@@ -876,6 +877,78 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+			newReleasePlan.Kind = "ReleasePlan"
+
+			adapter.ctx = toolkit.GetMockedContext(ctx, []toolkit.MockData{
+				{
+					ContextKey: loader.ReleasePlanContextKey,
+					Resource:   newReleasePlan,
+				},
+			})
+
+			/*
```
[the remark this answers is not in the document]
theflockers: True. I removed to make some changes requested by Johnny and forgot to add back.

tekton/utils/pipeline_run_builder.go
```diff
@@ -111,6 +111,20 @@ func (b *PipelineRunBuilder) WithObjectReferences(objects ...client.Object) *Pip
 	return b
 }
 
+// WithTenantArrayPrams constructs tektonv1.Param entries for each of the provided client.Objects.
+func (b *PipelineRunBuilder) WithTenantArrayParams(params []Param) *PipelineRunBuilder {
```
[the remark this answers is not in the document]
theflockers: Fine. I added this because it was not working passing the parameters to WithObjectReferences just like the managed one. I will figure out how to fix it.

controllers/release/adapter.go
```diff
@@ -203,22 +206,46 @@ func (a *adapter) EnsureReleaseIsProcessed() (controller.OperationResult, error)
...
+			return controller.RequeueWithError(fmt.Errorf("tenant PipelineRun must have ReleasePlan.Spec.Pipeline defined"))
```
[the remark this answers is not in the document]
theflockers: I see the suggested message as confusing as the first one, but fine as well.

controllers/release/adapter.go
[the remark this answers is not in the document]
theflockers: fixed. All around.

controllers/release/adapter.go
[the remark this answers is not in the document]
theflockers: changed.

controllers/release/adapter.go
davidmogar: If it was to me, I would have implemented it in two different operations. That way they are shorter, easier to maintain and more idempotent. Letting @johnbieren decide here whether we go in that direction now, in the future or never ;)

controllers/release/adapter.go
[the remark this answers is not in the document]
davidmogar: That's why it was a suggestion. Feel free to adapt it in any way.

tekton/utils/pipeline_run_builder.go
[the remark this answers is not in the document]
davidmogar: As Johnny said, in this case you can (and should) use WithParams directly.

loader/loader.go
davidmogar: It made sense before and it still makes sense when you get the processing resources for the managed Pipeline. I think the problem here is that we are reusing functions that were used for the managed Pipeline for both of them now. Functions are cheap. We could define new ones. And in the case of the tenant Pipeline, no need for a new composite function.

controllers/release/adapter.go
johnbieren: I think the suggestion is instead of modifying EnsureReleaseIsProcessed to support both managed and tenant pipelines, to create a new operation like EnsureTenantPipelinesAreExecuted or something to run the tenant pipeline and leave the managed operation alone (besides perhaps a check to skip it if a tenant pipeline is defined). I do think this is a bit simpler and will pay off as we add more to it in the future

controllers/releaseplan/adapter.go
johnbieren: I will try testing this later and report back

loader/loader.go
johnbieren: Yeah, I am in favor of not modifying GetProcessingResources. I think you have already did this based on the adapter code, but the function is still modified in loader.go

johnbieren (commented):
I will report back about the releaseplan adapter changes. Let me know if something else is fixed and I haven't resolved the thread (if it is one I started)

controllers/releaseplan/adapter.go
johnbieren: For whatever reason, my clusterbot cluster was not firing any controllers when I updated RPs. No idea why. Anyways, I think the issue is https://github.com/konflux-ci/release-service/pull/451/files#diff-8e9af1e483632eac17610237caaa5f879ea7bfb3e1e1ea228e45a17c770b9432R117
That should be `!=` Can you fix that and see if you still get an error when not modifying the releaseplan adapter function?

theflockers: Sorry not replying here before. Just got lost making all changes requested. The code is not the one to be tested anymore. I will move the PR to draft until I get proper version to be reviewed.

controllers/release/adapter.go
theflockers: Please take a look now if the makes sense. I added the new Handler.

controllers/releaseplan/adapter.go
theflockers: you can close this one as there is not change in this function anymore.

loader/loader.go
theflockers: I removed the changes here as well.

tekton/utils/pipeline_run_builder.go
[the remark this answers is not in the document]
theflockers: Ok. Please take a look if this is done correctly now.

controllers/release/adapter.go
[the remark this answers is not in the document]
theflockers: np. should be ok now.

controllers/release/adapter.go
```diff
@@ -180,14 +181,58 @@ func (a *adapter) EnsureReleaseIsRunning() (controller.OperationResult, error) {
 	return controller.ContinueProcessing()
 }
 
+// EnsureTenantReleasePipelineIsProcessed is an operation that will ensure that a Tenant Release PipelineRun associated to the Release
```
johnbieren, line 184: doesn't match function name (I like this better than the function name)

controllers/release/adapter.go
```diff
@@ -180,14 +181,58 @@ func (a *adapter) EnsureReleaseIsRunning() (controller.OperationResult, error) {
...
+	if a.release.HasProcessingFinished() {
+		return controller.ContinueProcessing()
+	}
+
+	pipelineRun, err := a.loader.GetReleasePipelineRun(a.ctx, a.client, a.release)
+	if err != nil && !errors.IsNotFound(err) {
+		return controller.RequeueWithError(err)
+	}
+
+	roleBinding, _ := a.loader.GetRoleBindingFromReleaseStatus(a.ctx, a.client, a.release)
```
johnbieren, line 196: We aren't passing the roleBinding in this operation, so I think we can remove this line and just pass nil in its place on line 222

controllers/release/adapter.go
johnbieren, line 1: I think you are missing some logic in Ensure*PipelineIsProcessed operation to determine which kind to run. As it is currently written, if you defined a pipeline in both RPA and RP, I think it is a race condition which one gets executed. I think in the managed one you should say "if RP.Pipeline = nil, continue processing" and similarly in the tenant one "if RPA.Pipeline = nil, continue processing"
It may also make sense to put an additional validation check for if both RP.Pipeline and RPA.Pipeline are not nil, throw validation error" (as I've mentioned, we would later remove this check for when we support both, but we don't support both yet and can't have both run as they are writing to the same release status fields)

johnbieren (commented):
I reviewed everything besides `controller/release/adapter_test.go` I will review that once I am :heavy_check_mark: with `controller/release/adapter.go`

I believe I have resolved all my old comments. Please ping me on them if I missed any (excluding the ones opened with this review of course)

controllers/release/adapter.go
theflockers: I just noticed this as well. I am adding the condition.

controllers/release/adapter.go
theflockers: fixed.

controllers/release/adapter.go
theflockers: removed.

controllers/release/adapter.go
theflockers: added the two required conditions.

controllers/release/adapter.go
```diff
@@ -373,6 +420,56 @@ func (a *adapter) createManagedPipelineRun(resources *loader.ProcessingResources
 	return pipelineRun, nil
 }
 
+// createTenantPipelineRun creates and returns a new tenant Release PipelineRun. The new PipelineRun will include owner
+// annotations, so it triggers Release reconciles whenever it changes. The Pipeline information and the parameters to it
+// will be extracted from the given ReleasePlan. The Release's Snapshot will also be passed to the release
+// PipelineRun.
+func (a *adapter) createTenantPipelineRun(releasePlan *v1alpha1.ReleasePlan, snapshot *applicationapiv1alpha1.Snapshot) (*tektonv1.PipelineRun, error) {
+	tenantPipelineParams := []tektonv1.Param{}
```
davidmogar, line 428: I'd add this block to a `ToTektonParams` function in `pipeline.go`. I wouldn't block the PR for this. I'm ok with this being refactored later.

controllers/release/adapter.go
```diff
@@ -373,6 +420,56 @@ func (a *adapter) createManagedPipelineRun(resources *loader.ProcessingResources
...
+		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, integrationgitops.PipelinesAsCodePrefix)).
+		WithFinalizer(metadata.ReleaseFinalizer).
+		WithLabels(map[string]string{
+			metadata.ApplicationNameLabel:  releasePlan.Spec.Application,
+			metadata.PipelinesTypeLabel:    metadata.TenantPipelineType,
+			metadata.ReleaseNameLabel:      a.release.Name,
+			metadata.ReleaseNamespaceLabel: a.release.Namespace,
+			metadata.ReleaseSnapshotLabel:  a.release.Spec.Snapshot,
+		}).
+		WithObjectReferences(a.release, releasePlan, a.releaseServiceConfig, snapshot).
```
davidmogar, line 449: I'm not sure we want to pass the RSC to a tenant pipeline.

controllers/release/adapter.go
```diff
@@ -561,20 +658,49 @@ func (a *adapter) validateAuthor() *controller.ValidationResult {
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			a.release.MarkValidationFailed(err.Error())
+			return &controller.ValidationResult{Valid: false}
+		}
+		return &controller.ValidationResult{Err: err}
+	}
+
+	if releasePlan.Spec.Pipeline != nil {
+		if !a.releaseServiceConfig.Spec.Debug && releasePlan.Spec.Pipeline.PipelineRef.IsClusterScoped() {
```
davidmogar, line 696: This is duplicated code that could and should go into a function.

controllers/release/adapter.go
```diff
@@ -591,3 +717,31 @@ func (a *adapter) validatePipelineRef() *controller.ValidationResult {
...
+
+	releasePlanAdmission, err := a.loader.GetActiveReleasePlanAdmissionFromRelease(a.ctx, a.client, a.release)
+	if err != nil && errors.IsNotFound(err) {
+		a.release.MarkValidationFailed(err.Error())
+		return &controller.ValidationResult{Valid: false}
+	}
+
+	if releasePlanAdmission != nil {
+		if releasePlan.Spec.Pipeline != nil && releasePlanAdmission.Spec.Pipeline != nil {
+			a.release.MarkValidationFailed("pipeline should be set only in the ReleasePlan or in the ReleasePlanAdmission")
```
davidmogar, line 741: We actually don't care about this. I think that the proper check, provided that we want to support both in the future, is to check that Target and Pipeline are not defined at the same time in the RP. No need to check the RPA.

controllers/release/controller.go
```diff
@@ -81,7 +81,8 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		adapter.EnsureReleaseIsValid,
 		adapter.EnsureFinalizerIsAdded,
 		adapter.EnsureReleaseExpirationTimeIsAdded,
-		adapter.EnsureReleaseIsProcessed,
+		adapter.EnsureTenantReleasePipelineIsProcessed,
```
davidmogar, line 84: This is become a mouthful. What do you think about renaming to `Ensure{Tenant|Managed}PipelineIsProcessed`?

controllers/release/adapter.go
theflockers: I am pretty sure I was told to remove the previous function from there, that was doing exactly this.

controllers/release/adapter.go
theflockers: Ok. This makes it easier.

controllers/release/controller.go
theflockers: makes sense. I wil do it.

controllers/release/adapter.go
```diff
@@ -180,14 +182,58 @@ func (a *adapter) EnsureReleaseIsRunning() (controller.OperationResult, error) {
...
+}
+
+// EnsureManagedReleasePipelineIsProcessed is an operation that will ensure that a managed Release PipelineRun associated to the Release
+// being processed and a RoleBinding to grant its serviceAccount permissions exist. Otherwise, it will create them.
+func (a *adapter) EnsureManagedReleasePipelineIsProcessed() (controller.OperationResult, error) {
+	if a.release.HasProcessingFinished() {
+		return controller.ContinueProcessing()
+	}
+
+	pipelineRun, err := a.loader.GetReleasePipelineRun(a.ctx, a.client, a.release)
```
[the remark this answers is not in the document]
theflockers: I did not noticed any misbehavior while testing this. I will look deeper it the coding is actually stopping further processing after the tenant pipeline finishes and it gets to this point.

controllers/release/adapter.go
theflockers: Ah, ok. The previous function was in the `pipeline_run_builder.go`. I will do that.

controllers/release/adapter.go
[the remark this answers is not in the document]
johnbieren: A tenant pipeline can't be found in this function because this is the managed operation and we don't let a user define both a tenant and managed pipelinerun

controllers/release/adapter.go
johnbieren: Just don't make it tenant pipeline specific. have it just take a list of params and do the append

controllers/release/adapter.go
johnbieren: I actually disagree with this and think it is a much better check that a pipeline is not defined in both the RP and RPA. Regardless, we all agree this is temporary code (disallowing a pipeline in both spots) so it doesn't matter to me - do whatever is easiest, no need to waste cycles on temporary things

controllers/release/adapter.go
```diff
@@ -180,14 +182,58 @@ func (a *adapter) EnsureReleaseIsRunning() (controller.OperationResult, error) {
 	return controller.ContinueProcessing()
 }
 
-// EnsureReleaseIsProcessed is an operation that will ensure that a managed Release PipelineRun associated to the Release
+// EnsureTenantPipelineIsProcessed is an operation that will ensure that a Tenant Release PipelineRun associated to the Release
 // being processed and a RoleBinding to grant its serviceAccount permissions exist. Otherwise, it will create them.
```
johnbieren, line 186: no rolebinding

controllers/release/adapter.go
```diff
@@ -561,33 +657,82 @@ func (a *adapter) validateAuthor() *controller.ValidationResult {
...
-		if resources == nil || resources.ReleasePlanAdmission == nil || errors.IsNotFound(err) {
+		if errors.IsNotFound(err) {
 			a.release.MarkValidationFailed(err.Error())
 			return &controller.ValidationResult{Valid: false}
 		}
-
 		return &controller.ValidationResult{Err: err}
 	}
+
+	if releasePlan.Spec.Target != "" {
```
johnbieren, line 669: This relates to the other comment about if we use target == nil or pipeline in RP == nil to determine if it is a tenant pipeline setup. Whichever you decide to go with, just make sure this line syncs up please

controllers/release/adapter.go
theflockers: removed.

controllers/release/adapter.go
theflockers: changed to use  the pipeline one.

controllers/release/controller.go
theflockers: fixed

controllers/release/adapter.go
theflockers: I ended up changing it to not use the RPA.

controllers/release/adapter.go
theflockers: fixed.

controllers/release/adapter.go
theflockers: removed.

controllers/release/adapter.go
theflockers: new function added to `utils/pipeline.go`

controllers/release/adapter.go
```diff
@@ -214,11 +260,12 @@ func (a *adapter) EnsureReleaseIsProcessed() (controller.OperationResult, error)
 			}
 
 			pipelineRun, err = a.createManagedPipelineRun(resources)
+
```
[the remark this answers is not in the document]
theflockers: removed.

controllers/release/adapter_test.go
```diff
@@ -532,12 +533,154 @@ var _ = Describe("Release adapter", Ordered, func() {
...
 
-			pipelineRun, err := adapter.loader.GetManagedReleasePipelineRun(adapter.ctx, adapter.client, adapter.release)
+			pipelineRun, err := adapter.loader.GetReleasePipelineRun(adapter.ctx, adapter.client, adapter.release)
+			Expect(pipelineRun).NotTo(BeNil())
+			Expect(err).NotTo(HaveOccurred())
+			Expect(adapter.client.Delete(adapter.ctx, pipelineRun)).To(Succeed())
+		})
+	})
+
+	When("EnsureTenantPipelineIsProcessed is called", func() {
```
johnbieren, line 548: Can you add a test for continue processing occurs when releaseplan.spec.pipeline == nil?

controllers/release/adapter_test.go
```diff
@@ -532,12 +533,154 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+					ContextKey: loader.ReleasePipelineRunContextKey,
+					Resource: &tektonv1.PipelineRun{
+						ObjectMeta: metav1.ObjectMeta{
+							Name:      "pipeline-run",
+							Namespace: "default",
+						},
+					},
+				},
+				{
+					ContextKey: loader.RoleBindingContextKey,
```
johnbieren, line 582: don't think we need this

controllers/release/adapter_test.go
```diff
@@ -876,6 +1019,133 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+
+		AfterEach(func() {
+			_ = adapter.client.Delete(ctx, adapter.release)
+
+			Expect(k8sClient.Delete(ctx, pipelineRun)).To(Succeed())
+		})
+
+		BeforeEach(func() {
+			adapter = createReleaseAndAdapter()
+			adapter.releaseServiceConfig = releaseServiceConfig
```
johnbieren, line 1037: rsc no longer matters here

controllers/release/adapter_test.go
johnbieren: It actually looks like this is tested because you didn't change the the default releaseplan in this file to have a pipeline, but it is not obvious. Can you make it obvious? Basically I am looking for 2 versions of the test on line 570 (should continue if the PipelineRun exists and the release processing has started). One for if the RP has no pipeline, one for if the RP has a pipeline but the pipelinerun already exists

johnbieren (commented):
My only remaining comments are in the test file

controllers/release/adapter.go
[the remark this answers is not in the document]
theflockers: This is working as expected. Something else it not adding the `Released`  condition to it.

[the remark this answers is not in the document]
davidmogar: Ok, so it works because only one kind of Pipeline is allowed. I don't like this way of coding. Not resilient enough. I'm ok to let it go through for the matter of merging this PR as soon as we can, but that needs a refactor IMHO.

tekton/utils/pipeline.go
```diff
@@ -57,7 +57,8 @@ type Pipeline struct {
...
+	// +optional
```
davidmogar, line 61: Missing empty line before this one and missing documentation.

tekton/utils/pipeline.go
```diff
@@ -85,6 +86,23 @@ func (pr *PipelineRef) ToTektonPipelineRef() *tektonv1.PipelineRef {
 	return tektonPipelineRef
 }
 
+// ToTektonParams converts ParameterizedPipeline []Param to []tektonv1.Param and returns it.
+func (prp *ParameterizedPipeline) ToTektonParams() []tektonv1.Param {
```
davidmogar, line 90: Not what I had in mind. I expected an array of Params to be passed. If you go this route, call it `GetTektonParams` as you are not converting a `ParameterizedPipeline` here.

theflockers: Ok. For now I renamed. We can sync later on show this should be properly done, so it goes with the refactor.

tekton/utils/pipeline.go
theflockers: added.

controllers/release/adapter_test.go
theflockers: removed.

controllers/release/adapter_test.go
theflockers: removed.

controllers/release/adapter_test.go
theflockers: test added.

johnbieren (dismissed):
I am good with the changes. Now just to get David's approval :slightly_smiling_face:

controllers/release/adapter.go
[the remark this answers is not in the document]
davidmogar: This way of coding like in assuming that something else prevented the behavior. Operations depending on other operations behavior. Can be fixed later on.

davidmogar (approved):

johnbieren (approved):

PR #540, opened 2024-08-20
feat(RELEASE-913): add GetPreviousRelease loader function

davidmogar:
A new funciton has been added to the loader to be able to get the Release created just before a given Release. This is a key function for the collectors feature.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -3,6 +3,8 @@ package loader
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"os"
 	"strings"
 
@@ -29,6 +31,7 @@ type ObjectLoader interface {
 	GetEnterpriseContractPolicy(ctx context.Context, cli client.Client, releasePlanAdmission *v1alpha1.ReleasePlanAdmission) (*ecapiv1alpha1.EnterpriseContractPolicy, error)
 	GetMatchingReleasePlanAdmission(ctx context.Context, cli client.Client, releasePlan *v1alpha1.ReleasePlan) (*v1alpha1.ReleasePlanAdmission, error)
 	GetMatchingReleasePlans(ctx context.Context, cli client.Client, releasePlanAdmission *v1alpha1.ReleasePlanAdmission) (*v1alpha1.ReleasePlanList, error)
+	GetPreviousRelease(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*v1alpha1.Release, error)
 	GetRelease(ctx context.Context, cli client.Client, name, namespace string) (*v1alpha1.Release, error)
 	GetRoleBindingFromReleaseStatus(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*rbac.RoleBinding, error)
 	GetReleasePipelineRun(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*tektonv1.PipelineRun, error)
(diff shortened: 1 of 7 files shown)


loader/loader.go
```diff
@@ -170,6 +173,42 @@ func (l *loader) GetMatchingReleasePlans(ctx context.Context, cli client.Client,
...
+
+	// Find the previous release
+	for i, possiblePreviousRelease := range releases.Items {
+		// Ignore the release passed as argument and any release created after that one
+		if possiblePreviousRelease.Name == release.Name ||
+			possiblePreviousRelease.CreationTimestamp.After(release.CreationTimestamp.Time) {
+			continue
+		}
+		if previousRelease == nil || possiblePreviousRelease.CreationTimestamp.After(previousRelease.CreationTimestamp.Time) {
+			previousRelease = &releases.Items[i]
```
johnbieren, line 197: isn't `releases.Items[i]` just equal to `possiblePreviousRelease`? So we don't need `i` in the loop, we can use `_`?

loader/loader_test.go
johnbieren, line 1: Do we always have to sleep for 2 seconds? Is 1 second flaky?

davidmogar: To be honest, not sure. Just went a bit safer there... The whole thing sucks and we should be using a fake client but that's hard.

davidmogar: Changed to 1 seconds. Easy to change I we see it failing.

loader/loader.go
```diff
@@ -170,6 +173,42 @@
...
+
+	// Find the previous release
+	for _, possiblePreviousRelease := range releases.Items {
+		// Ignore the release passed as argument and any release created after that one
+		if possiblePreviousRelease.Name == release.Name ||
+			possiblePreviousRelease.CreationTimestamp.After(release.CreationTimestamp.Time) {
+			continue
+		}
+		if previousRelease == nil || possiblePreviousRelease.CreationTimestamp.After(previousRelease.CreationTimestamp.Time) {
+			previousRelease = &possiblePreviousRelease
```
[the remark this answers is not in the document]
johnbieren: yeah, didn't expect that but good to know

loader/loader.go
```diff
@@ -170,6 +173,42 @@ func (l *loader) GetMatchingReleasePlans(ctx context.Context, cli client.Client,
...
+		return nil, err
+	}
+
+	var previousRelease *v1alpha1.Release
+
+	// Find the previous release
+	for i, possiblePreviousRelease := range releases.Items {
+		// Ignore the release passed as argument and any release created after that one
+		if possiblePreviousRelease.Name == release.Name ||
+			possiblePreviousRelease.CreationTimestamp.After(release.CreationTimestamp.Time) {
```
theflockers, line 193: Just a quick question here. I remember happy having trouble with this functionality as eventually you would have more than one release created at the very same second. So IIUC the `.After`  can (somehow) identify which one is the newer one.

loader/loader_test.go
```diff
@@ -238,6 +240,83 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			Eventually(func() error {
+				return k8sClient.Get(ctx, client.ObjectKey{Name: newerRelease.Name, Namespace: newerRelease.Namespace}, newerRelease)
+			}).Should(Succeed())
+
+			returnedObject, err := loader.GetPreviousRelease(ctx, k8sClient, newerRelease)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(returnedObject).ToNot(BeNil())
+			Expect(returnedObject.Name).To(Equal(release.Name))
+		})
+
```
theflockers, line 286: should we also test the scenario I described on top? Two releases created at the same second? Or we are assuming that this in the real scenario should not happen?

theflockers (commented):
Just a two question. Once they are cleared, I can approve.

loader/loader.go
davidmogar: For the same app? Hard to see. In any case nothing we can do there. Any of it would be the winner and that's fine.

loader/loader_test.go
davidmogar: It should not happen and there's not much we can do there.

theflockers: Indeed. while testing there we noticed that if two releases (or more) are created in the same second we can not be sure which one is newer as the createdTimestamp precision is in seconds. In fact, nothing we can do about that.

theflockers (approved):

johnbieren (approved):

happybhati (approved):
LGTM!

PR #541, opened 2024-08-20
feat(RELEASE-908): add collectors field

davidmogar:
A new collectors field is added to ReleasePlans and ReleasePlansAdmissions to drive the execution of collectors as part of the release workflow.


api/v1alpha1/collectors.go
```diff
@@ -0,0 +1,27 @@
...
+type Collector struct {
+	// Name of the collector
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	Name string `json:"name"`
+
+	// Collector is the type of collector to be used
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	Collector string `json:"collector"`
```
johnbieren, line 14: What do you think about naming this type? I assume this will be like `jira` for example. collector.collector seems a little weird to me

api/v1alpha1/collectors.go
```diff
@@ -0,0 +1,27 @@
...
+
+	// Params is a slice of parameters for a given collector
+	Params []Param `json:"params"`
+}
+
+// Param represents a parameter for a collector
+type Param struct {
```
johnbieren, line 21: Can't we use the same Param definition we have in the tekton directory? Instead of defining an identical struct in two places

api/v1alpha1/collectors.go
davidmogar: Sure. Makes sense. I went with what we had in the ticket but it sounded weird to me.

api/v1alpha1/collectors.go
davidmogar: I didn't want to link this with the Tekton stuff...

Probably we could move it to operator-toolkit in the future.

johnbieren: I am fine with it going into operator-toolkit. I don't want two identical structs in two spots in one repo though

api/v1alpha1/collectors.go
johnbieren: The json tag wasn't updated

davidmogar: Right!

api/v1alpha1/collectors.go
johnbieren: Yeah, followup ticket is fine, or even adding it to the AC of the cleanup tekton dir ticket if you think that is related enough. As long as it is written on a ticket somewhere

johnbieren (approved):
approving pending the one open discussion

johnbieren (approved):
approved pending creation of that ticket, but there is an open conversation about that one

PR #549, opened 2024-08-28
feat(RELEASE-1109): allow tenant and managed pipelines to run

johnbieren:
This commit modifies the release controller to execute both a tenant and managed pipelineRun in the same Release if both are defined.


api/v1alpha1/release_types.go
```diff
@@ -232,37 +261,105 @@ func (r *Release) IsValid() bool {
...
 		r.Status.Processing.StartTime,
 		r.Status.Processing.CompletionTime,
 		SucceededReason.String(),
 		r.Status.Target,
 	)
 }
 
-// MarkProcessing marks the Release as processing.
+// MarkManagedProcessed marks the Release managed portion as processed.
+func (r *Release) MarkManagedProcessed() {
```
davidmogar, line 298: Maybe we should add the word `Pipeline` to these functions.

api/v1alpha1/release_types.go
```diff
@@ -232,37 +261,105 @@ func (r *Release) IsValid() bool {
...
+	go metrics.RegisterCompletedReleaseManagedProcessing(
+		r.Status.Processing.StartTime,
+		r.Status.Processing.CompletionTime,
+		SucceededReason.String(),
+		r.Status.Target,
+	)
+}
+
+// MarkProcessing marks the Release overall as processing.
 func (r *Release) MarkProcessing(message string) {
```
davidmogar, line 314: I'd say that this have to go. HasProcessingFinished should be based in the state of each phase (HasTenantPipelineFinished, HasManagedPipelineFinished and HasFinalPipelineFinished).

controllers/release/adapter.go
```diff
@@ -166,6 +166,22 @@ func (a *adapter) EnsureReleaseIsCompleted() (controller.OperationResult, error)
 	return controller.RequeueOnErrorOrContinue(a.client.Status().Patch(a.ctx, a.release, patch))
 }
 
+// EnsureReleaseIsProcessing is an operation that will ensure that a Release  is marked as processing. If the
+// Release is already marked as processing, no action will be taken.
+func (a *adapter) EnsureReleaseIsProcessing() (controller.OperationResult, error) {
```
davidmogar, line 171: I don't think we need this.

controllers/release/adapter.go
```diff
@@ -337,14 +348,50 @@ func (a *adapter) EnsureReleaseProcessingIsTracked() (controller.OperationResult
...
+			return controller.RequeueWithError(err)
+		}
+	}
+
+	return controller.ContinueProcessing()
+}
+
+// EnsureReleaseProcessingTenantResourcesAreCleanedUp is an operation that will ensure that the resources created for the
+// Release Tenant Processing step are cleaned up once processing is finished.
+func (a *adapter) EnsureReleaseTenantProcessingResourcesAreCleanedUp() (controller.OperationResult, error) {
```
davidmogar, line 374: All of them can go together. We should search all the pipelines for the release (we have a label for that) and delete them in the same function.

davidmogar (commented):
I just read it in diagonal and added some comments. We can discuss them offline (and maybe address them) before I do a proper review.

api/v1alpha1/release_types.go
```diff
@@ -83,6 +96,11 @@ type ReleaseStatus struct {
 	// +optional
 	Target string `json:"target,omitempty"`
 
+	// Type indicates if this is a tenant release, managed release, or both
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +optional
+	Type string `json:"type,omitempty"`
```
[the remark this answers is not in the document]
johnbieren: Because I need a way to quickly determine if the release has only a tenant or only a managed pipeline. Otherwise I have to do like 5 lines of logic in a bunch of places

controllers/release/controller.go
```diff
@@ -82,9 +82,9 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		adapter.EnsureFinalizerIsAdded,
 		adapter.EnsureReleaseExpirationTimeIsAdded,
 		adapter.EnsureTenantPipelineIsProcessed,
+		adapter.EnsureReleaseTenantProcessingIsTracked,
 		adapter.EnsureManagedPipelineIsProcessed,
-		adapter.EnsureReleaseProcessingIsTracked,
-		adapter.EnsureReleaseProcessingResourcesAreCleanedUp,
```
johnbieren, line 87: @davidmogar FYI I removed this operation. Previously, this operation just removed the finalizer, then the finalize release operation deleted the pipelinerun. We talked about changing it so they used the same code. I tested this, and it broke things because the pipelinerun was being deleted ~as soon as it finished, instead of waiting for the release to be deleted. So, now everything is being deleted when the release is finalized. The only change here is that the roleBinding is now kept around until everything else is deleted (the release and subsequently the two pipelineruns). If this isn't okay, we can talk about it

johnbieren: Disregard this. I forgot about the requirement to remove the finalizer from the pipelineRun even before the releaes is called to be deleted. Will update tomorrow

metrics/release.go
```diff
@@ -39,10 +39,18 @@ var (
 		[]string{},
```
theflockers, line 39: writing here, to make it easier. I am wondering it you could just keep the same metrics functions and adding the extra info as labels instead of duplicating. For instance, the`ReleaseProcessingDurationSeconds` we only need a new label saying it it is managed or tenant, or later final. That is enough to classify in the dashboard.

theflockers: We don´t track the CR name (release or pipelinerun, which I think we should add to the metrics), so we can't know if the managed and the tenant metric belongs to the same CR. Anyhow, in the current implementation, if the idea is having a metric to track specifically each pipelinerun duration, I would suggest instead of changing the ReleaseProcessingDurationSeconds (so we still can know how long the whole release took - and the current graphs does not require changes), to add a new one ie. ReleasePipelinerunsDurationSeconds that we could track all pipelines durations. - Of course, if we have had the release name and the pipelinerun name maybe only the second one would be necessary, because it would look like this:

release_pipelineruns_duration_seconds{name="release-abc", pipelinerun="tenant-abc",  type="tenant", target="...tenant", ...}

This would be easy to query and graph.

controllers/release/controller.go
```diff
@@ -82,8 +82,9 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		adapter.EnsureFinalizerIsAdded,
 		adapter.EnsureReleaseExpirationTimeIsAdded,
 		adapter.EnsureTenantPipelineIsProcessed,
+		adapter.EnsureReleaseTenantProcessingIsTracked,
```
davidmogar, line 85: I'd rename these functions to `EnsureTenantPipelineProcessingIsTracked` and `EnsureManagedProcessingIsTracked`.

loader/loader.go
```diff
@@ -119,6 +120,11 @@ func (l *loader) GetMatchingReleasePlanAdmission(ctx context.Context, cli client
 		return releasePlanAdmission, toolkit.GetObject(designatedReleasePlanAdmissionName, releasePlan.Spec.Target, cli, ctx, releasePlanAdmission)
 	}
 
+	// This is a temporary hack to work around RELEASE-1150. It should improved when that ticket is complete.
```
davidmogar, line 123: Is this temporary? Wouldn't this be the outcome of that ticket?

loader/loader.go
```diff
@@ -220,10 +226,10 @@ func (l *loader) GetRelease(ctx context.Context, cli client.Client, name, namesp
 // by the namespaced name stored in the Release's status.
 func (l *loader) GetRoleBindingFromReleaseStatus(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*rbac.RoleBinding, error) {
 	roleBinding := &rbac.RoleBinding{}
-	roleBindingNamespacedName := strings.Split(release.Status.Processing.RoleBinding, string(types.Separator))
+	roleBindingNamespacedName := strings.Split(release.Status.ManagedProcessing.RoleBinding, string(types.Separator))
```
davidmogar, line 229: I'm a bit confused. The RoleBinding only makes sense in the context of the Managed Pipeline but now as we reuse the PipelineInfo, every single pipeline has that field. Should we move it to a different place? We have 4 pipeline types and only one requires that field.

loader/loader.go
```diff
@@ -237,15 +243,20 @@ func (l *loader) GetRoleBindingFromReleaseStatus(ctx context.Context, cli client
 	return roleBinding, nil
 }
 
-// GetReleasePipelineRun returns the Release PipelineRun referenced by the given Release or nil if it's not found. In the case
-// the List operation fails, an error will be returned.
-func (l *loader) GetReleasePipelineRun(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*tektonv1.PipelineRun, error) {
+// GetReleasePipelineRun returns the Release PipelineRun of the specified type referenced by the given Release
+// or nil if it's not found. In the case the List operation fails, an error will be returned.
+func (l *loader) GetReleasePipelineRun(ctx context.Context, cli client.Client, release *v1alpha1.Release, pipelineType string) (*tektonv1.PipelineRun, error) {
```
davidmogar, line 248: We could think about using an enum for the pipelines types as is less error prone. Maybe create a ticket for the future?

metrics/release.go
davidmogar: I'm ok with adding the type but we should never add the name of the release in my opinion. It doesn't repeat over time and would create performance issues.

controllers/release/adapter.go
```diff
@@ -67,10 +67,10 @@ func newAdapter(ctx context.Context, client client.Client, release *v1alpha1.Rel
 	}
 
 	releaseAdapter.validations = []controller.ValidationFunction{
+		releaseAdapter.validatePipelineDefined,
 		releaseAdapter.validateProcessingResources,
 		releaseAdapter.validateAuthor,
 		releaseAdapter.validatePipelineRef,
```
davidmogar, line 73: It would be great to rename this so it doesn't sound like it's similar to the new one.

controllers/release/adapter.go
```diff
@@ -337,28 +353,40 @@ func (a *adapter) EnsureReleaseProcessingIsTracked() (controller.OperationResult
...
+// EnsureReleaseProcessingResourcesAreCleanedUp is an operation that will ensure that the resources created for the Release
+// Processing step are cleaned up once processing is finished. This exists in conjunction with EnsureFinalizersAreCalled because
+// the finalizers should be removed from the pipelineRuns even if the Release is not marked for deletion for quota reasons.
+func (a *adapter) EnsureReleaseProcessingResourcesAreCleanedUp() (controller.OperationResult, error) {
+	if !a.release.HasTenantPipelineProcessingFinished() || !a.release.HasManagedPipelineProcessingFinished() {
+		return controller.ContinueProcessing()
 	}
 
-	return controller.RequeueOnErrorOrContinue(a.cleanupProcessingResources(pipelineRun, roleBinding))
+	return controller.RequeueOnErrorOrContinue(a.finalizeRelease(false))
```
davidmogar, line 385: Why not to delete them here as well?

controllers/release/adapter.go
```diff
@@ -495,29 +529,47 @@ func (a *adapter) createRoleBindingForClusterRole(clusterRole string, releasePla
...
+		return err
+	}
+
+	err = a.cleanupProcessingResources(managedPipelineRun, roleBinding)
+	if err != nil {
+		return err
+	}
+
+	if delete && managedPipelineRun != nil {
+		err = a.client.Delete(a.ctx, managedPipelineRun)
```
davidmogar, line 572: It would be cleaner if the blocks deleting the pipelines would be inside the `cleanUpProcessingResource`. You can even move there the piece getting the pipeline.

davidmogar (changes requested):

loader/loader.go
johnbieren: I am fine with moving it, but do you have any ideas of how to do it cleanly? It would kind of suck to have a `PipelineInfoWithRolebinding` struct but that is kind of the cleanest way I can think to do it

loader/loader.go
johnbieren: https://issues.redhat.com/browse/RELEASE-1180

controllers/release/adapter.go
johnbieren: This *is* the new one. Which one do you not want to sound similar to? Then I can adjust

controllers/release/adapter.go
johnbieren: I added it to the function comment why it is this way, but I will explain in the other comment about deleting them

controllers/release/adapter.go
johnbieren: 2 reasons. The first is that it is nice to have them around before they are pruned for debugging and not rely on tekton results. The second is that when I tried this, I was having issues with the tracking information (fwiw, this is how we currently have things. Clean up removes finalizers, finalizeRelease deletes it)

metrics/release.go
johnbieren: Okay, I am changing it per your suggestion. I am doing it in a separate commit. If you two like how it looks, I will squash them together, but I wanted to do it separately in case I change it wrong and we want to go back to how I had it.
I am just working on fixing one thing in the test and it will be updated

johnbieren: Updated

api/v1alpha1/release_conditions.go
```diff
@@ -6,8 +6,11 @@ const (
 	// postActionsExecutedConditionType is the type used to track the status of Release post-actions
 	postActionsExecutedConditionType conditions.ConditionType = "PostActionsExecuted"
 
-	// processedConditionType is the type used to track the status of a Release processing
-	processedConditionType conditions.ConditionType = "Processed"
+	// managedProcessedConditionType is the type used to track the status of a Release Managed Pipeline processing
+	managedProcessedConditionType conditions.ConditionType = "ManagedPipelineProcessed"
```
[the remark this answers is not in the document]
johnbieren: fixed

api/v1alpha1/release_conditions.go
```diff
@@ -20,9 +23,12 @@ const (
...
 
-	// ProgressingReason is the reason set when an action is progressing
+	// ProgressingReason is the reason set when a phase is progressing
 	ProgressingReason conditions.ConditionReason = "Progressing"
 
-	// SucceededReason is the reason set when an action succeeds
+	// SucceededReason is the reason set when a phase succeeds
 	SucceededReason conditions.ConditionReason = "Succeeded"
+
+	// SkippedReason is the reason set when a phase is skipped
```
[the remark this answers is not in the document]
johnbieren: fixed

api/v1alpha1/release_types.go
```diff
@@ -68,11 +68,15 @@ type ReleaseStatus struct {
...
 	// PostActionsExecution contains information about the post-actions execution
 	// +optional
-	PostActionsExecution PostActionsExecutionInfo `json:"postActionsExecution,omitempty"`
+	PostActionsExecution PipelineInfo `json:"postActionsExecution,omitempty"`
 
-	// Processing contains information about the release processing
+	// ManagedProcessing contains information about the release managed processing
 	// +optional
-	Processing ProcessingInfo `json:"processing,omitempty"`
+	ManagedProcessing PipelineInfo `json:"managedProcessing,omitempty"`
```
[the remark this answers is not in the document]
johnbieren: fixed

api/v1alpha1/release_types.go
```diff
@@ -111,19 +115,8 @@ type AttributionInfo struct {
...
-	CompletionTime *metav1.Time `json:"completionTime,omitempty"`
-
-	// StartTime is the time when the Release post-actions execution started
-	// +optional
-	StartTime *metav1.Time `json:"startTime,omitempty"`
-}
-
-// ProcessingInfo defines the observed state of the release processing.
-type ProcessingInfo struct {
+// PipelineInfo defines the observed state of the release pipeline processing.
```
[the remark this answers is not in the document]
johnbieren: fixed

controllers/release/controller.go
johnbieren: done

loader/loader.go
johnbieren: When I wrote that, I thought we would come up with something more clever than an if statement. Removed the comment

controllers/release/adapter.go
davidmogar: `validatePipelineRef` and `validatePipelineDefined`.

loader/loader.go
davidmogar: Let's keep it this way and maybe go for a general clean up of the service code when we run up of tickets? Pretty sure we can improve some things now that we have more experience.

api/v1alpha1/release_types.go
```diff
@@ -66,13 +67,25 @@ type ReleaseStatus struct {
 	// +optional
 	Conditions []metav1.Condition `json:"conditions"`
 
+	// ProcessingStartTime is the time when the Processing portion of the release began
+	// +optional
+	ProcessingStartTime *metav1.Time `json:"processingStartTime,omitempty"`
```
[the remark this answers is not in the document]
davidmogar: Yes.

davidmogar (commented):
Looking ok to me once the open discussions are resolved.

controllers/release/adapter.go
johnbieren: Changed `validatePipelineRef` to `validatePipelineSource` as it is checking that the pipeline is from a bundle or git repo vs cluster scoped

metrics/release.go
theflockers: It works for me.

theflockers:
lgtm. Maybe link the new PR here (or the new ticket)  for the work on the metrics will be done.

theflockers (approved):

johnbieren:
> lgtm. Maybe link the new PR here (or the new ticket) of the work on the metrics will be done.

Okay, I squashed the commits down then. Can you clarify? What new PR or ticket? What metrics work needs to be done? It sounds like this is something for a new ticket, but I will need some more details before filing it

theflockers:
> > lgtm. Maybe link the new PR here (or the new ticket) of the work on the metrics will be done.
> 
> Okay, I squashed the commits down then. Can you clarify? What new PR or ticket? What metrics work needs to be done? It sounds like this is something for a new ticket, but I will need some more details before filing it

Sorry. I misunderstood what you said. I thought you wanted to do more changes besides the one I requested (which lgtm to me). So please ignore the comment!

PR #572, opened 2024-09-24
Draft: feat(RELEASE-1053): add support for final pipeline run

seanconroy2021:


api/v1alpha1/release_conditions.go
```diff
@@ -6,8 +6,14 @@ const (
 	// postActionsExecutedConditionType is the type used to track the status of Release post-actions
 	postActionsExecutedConditionType conditions.ConditionType = "PostActionsExecuted"
```
johnbieren, line 7: This line can be removed. PostActions is the old name for final pipeline

api/v1alpha1/release_types.go
```diff
@@ -68,11 +68,19 @@ type ReleaseStatus struct {
 
 	// PostActionsExecution contains information about the post-actions execution
 	// +optional
-	PostActionsExecution PostActionsExecutionInfo `json:"postActionsExecution,omitempty"`
+	PostActionsExecution PipelineInfo `json:"postActionsExecution,omitempty"`
```
johnbieren, line 71: This line can be removed

controllers/release/adapter.go
```diff
@@ -337,28 +399,61 @@ func (a *adapter) EnsureReleaseProcessingIsTracked() (controller.OperationResult
...
+		}
+	}
+
+	return controller.ContinueProcessing()
+}
+
+// EnsureReleaseProcessingResourcesAreCleanedUp is an operation that will ensure that the resources created for the Release
+// Processing step are cleaned up once processing is finished. This exists in conjunction with EnsureFinalizersAreCalled because
+// the finalizers should be removed from the pipelineRuns even if the Release is not marked for deletion for quota reasons.
+func (a *adapter) EnsureReleaseProcessingResourcesAreCleanedUp() (controller.OperationResult, error) {
```
johnbieren, line 447: This one should probably also check for final pipeline not being finished yet

controllers/release/adapter.go
```diff
@@ -681,8 +933,8 @@ func (a *adapter) validatePipelineRef() *controller.ValidationResult {
 	return &controller.ValidationResult{Valid: true}
 }
 
-// validateSinglePipeline checks that the Pipeline is defined exclusively in the ReleasePlan or in the ReleasePlanAdmission.
-func (a *adapter) validateSinglePipeline() *controller.ValidationResult {
+// validatePipelineDefined checks that a Pipeline is defined in either the ReleasePlan or in the ReleasePlanAdmission.
+func (a *adapter) validatePipelineDefined() *controller.ValidationResult {
```
johnbieren, line 937: I don't think this function should consider the final pipeline. Final pipeline is called final because it is like a finally block in code. You wouldn't have it be the only pipeline defined

johnbieren (commented):
PR looks quite good for a draft. I didn't review too closely as it is too hard to tell what your changes are since mine are also being shown until my PR is merged. One thing I will say though is that the PostActions thing is the finalPipeline. So anywhere you see PostActionsExecuted in like release_types.go or something, you can remove it

api/v1alpha1/release_types.go
seanconroy2021: Fixed

controllers/release/adapter.go
seanconroy2021: Add in the final pipeline

metrics/release.go
johnbieren, line 1: You're missing a few things in here. Search the file for "tenant" to point them out

controllers/releaseplan/adapter_test.go
```diff
@@ -214,7 +214,8 @@ var _ = Describe("ReleasePlan adapter", Ordered, func() {
 			Spec: v1alpha1.ReleasePlanSpec{
 				Application:            application.Name,
 				Target:                 "default",
-				Pipeline:               parameterizedPipeline,
+				TenantPipeline:         parameterizedPipeline,
+				FinalPipeline:          parameterizedPipeline,
```
johnbieren, line 218: The final pipeline is optional, I don't think you need this addition

controllers/release/adapter.go
```diff
@@ -729,7 +890,7 @@ func (a *adapter) validateProcessingResources() *controller.ValidationResult {
 		return &controller.ValidationResult{Err: err}
 	}
 
-	if releasePlan.Spec.Pipeline == nil {
+	if releasePlan.Spec.TenantPipeline == nil && releasePlan.Spec.FinalPipeline == nil {
```
johnbieren, line 893: I don't think this line should have the && releasePlan.Spec.FinalPipeline == nil

This check is basically saying if the RP does not define a tenant pipeline, it must define a managed one, so make sure everything we need to run a managed pipelineRun exists. The final pipeline doesn't really factor into this logic if we are going by "you have to run a tenant *or* managed pipeline per release, just a final is not good enough"

controllers/release/adapter.go
```diff
@@ -751,8 +912,10 @@ func (a *adapter) validatePipelineSource() *controller.ValidationResult {
 		return a.validationError(err)
 	}
 
-	if releasePlan.Spec.Pipeline != nil {
-		pipelineRef = releasePlan.Spec.Pipeline.PipelineRef
+	if releasePlan.Spec.TenantPipeline != nil {
+		pipelineRef = releasePlan.Spec.TenantPipeline.PipelineRef
+	} else if releasePlan.Spec.FinalPipeline != nil { // TODO: Refactor to remove the else if
```
johnbieren, line 917: I don't think we want this else if, maybe we should talk about it in standup. Honestly, this function should probably make sure each of the three pipelineRefs (or whichever are defined) is not cluster scoped. Currently, it does not check the managed one if a tenant pipeline is defined, but it should IMO

controllers/release/adapter.go
```diff
@@ -804,7 +967,7 @@ func (a *adapter) validatePipelineDefined() *controller.ValidationResult {
 			return &controller.ValidationResult{Err: err}
 		}
 		if releasePlanAdmission.Spec.Pipeline == nil {
-			errString := "releasePlan and releasePlanAdmission both have no pipeline. Each Release should define a tenant pipeline, managed pipeline, or both"
+			errString := "releasePlan has no pipeline or target. Each Release should define a tenant pipeline, managed pipeline or both"
```
johnbieren, line 970: This string should have remained as is

controllers/release/adapter.go
```diff
@@ -788,9 +951,9 @@ func (a *adapter) validatePipelineDefined() *controller.ValidationResult {
 		a.release.Status.Target = releasePlan.Spec.Target
 	}
 
-	if releasePlan.Spec.Pipeline == nil {
+	if releasePlan.Spec.TenantPipeline == nil {
 		if releasePlan.Spec.Target == "" {
-			errString := "releasePlan has no pipeline or target. Each Release should define a tenant pipeline, managed pipeline, or both"
+			errString := "releasePlan has no pipeline or target. Each Release should define a tenant pipeline, managed pipeline or both"
```
johnbieren, line 956: Let's leave the comma in there

johnbieren (commented):
I didn't look at the release_types_test.go or adapter_test.go files until we get the code nailed down

controllers/release/adapter.go
seanconroy2021: Fixed

controllers/release/adapter.go
seanconroy2021: Fixed

controllers/releaseplan/adapter_test.go
seanconroy2021: Removed

controllers/release/adapter.go
johnbieren: I believe you switched the errStrings here

seanconroy2021: Fixed

controllers/release/adapter.go
seanconroy2021: I changed it like you suggested in the call

metrics/release.go
seanconroy2021: I have added them now

seanconroy2021:
> I didn't look at the release_types_test.go or adapter_test.go files until we get the code nailed down

It all working now

controllers/release/adapter_test.go
```diff
@@ -262,7 +270,7 @@ var _ = Describe("Release adapter", Ordered, func() {
 			Expect(adapter.release.IsReleased()).To(BeFalse())
 		})
 
-		It("should do nothing if the managed processing has not completed", func() {
+		It("should do nothing if the final processing has not completed", func() {
```
johnbieren, line 273: The managed stuff on the It before this needs to be updated

controllers/release/adapter_test.go
```diff
@@ -322,6 +330,192 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+		It("should do nothing if the Release managed pipeline processing has not yet completed", func() {
+			adapter.release.MarkManagedPipelineProcessing()
+
+			result, err := adapter.EnsureFinalPipelineIsProcessed()
+			Expect(!result.RequeueRequest && !result.CancelRequest).To(BeTrue())
+			Expect(err).NotTo(HaveOccurred())
+			Expect(adapter.release.IsFinalPipelineProcessing()).To(BeFalse())
+		})
+
+		It("should do nothing if the Release tenant pipeline processing has not yet completed", func() {
```
johnbieren, line 364: The tenant pipeline status has no bearing on the final pipeline being processed

controllers/release/adapter_test.go
```diff
@@ -322,6 +330,192 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+			})
+			adapter.release.MarkManagedPipelineProcessingSkipped()
+
+			result, err := adapter.EnsureFinalPipelineIsProcessed()
+			Expect(result.RequeueRequest && !result.CancelRequest).To(BeTrue())
+			Expect(err).To(HaveOccurred())
+			Expect(adapter.release.IsFinalPipelineProcessing()).To(BeFalse())
+		})
+
+		It("should mark the Final Pipeline Processing as Skipped if the ReleasePlan isn't found", func() {
```
johnbieren, line 388: I would expect this test case to be like this one
```
		It("should requeue the Release if any of the resources is not found", func() {
			adapter.ctx = toolkit.GetMockedContext(ctx, []toolkit.MockData{
				{
					ContextKey: loader.ReleasePlanContextKey,
					Err:        fmt.Errorf("Not Found"),
					Resource:   nil,
				},
			})

			result, err := adapter.EnsureTenantPipelineIsProcessed()
			Expect(result.RequeueRequest && !result.CancelRequest).To(BeTrue())
			Expect(err).To(HaveOccurred())
		})
```

johnbieren: which you actually already have below, so this one can probably be deleted. The releasePlan has to be found... it shouldn't set it to skipped if it can't find it

controllers/release/adapter_test.go
```diff
@@ -913,7 +1108,7 @@ var _ = Describe("Release adapter", Ordered, func() {
 		})
 	})
 
-	When("EnsureManagedPipelineProcessingIsTracked is called", func() {
+	When("EnsureManagedProcessingIsTracked is called", func() {
```
johnbieren, line 1111: errant change. Should remain EnsureManagedPipelineProcessingIsTracked

controllers/release/adapter_test.go
```diff
@@ -974,6 +1169,65 @@ var _ = Describe("Release adapter", Ordered, func() {
 		})
 	})
 
+	When("EnsureFinalProcessingIsTracked is called", func() {
```
johnbieren, line 1172: EnsureFinalPipelineProcessingIsTracked

seanconroy2021: Fixed

controllers/release/adapter_test.go
seanconroy2021: I have removed it

controllers/release/adapter_test.go
seanconroy2021: Changed the above it

controllers/release/adapter_test.go
seanconroy2021: I have removed it

controllers/release/adapter_test.go
```diff
@@ -1064,6 +1292,13 @@ var _ = Describe("Release adapter", Ordered, func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		It("Should continure if the Release final processing has not finished", func() {
```
johnbieren, line 1295: s/continure/continue/

johnbieren (commented):
One very small typo then lgtm. I'd prefer to see a demo of the scenarios before we merge though. It would be nice to see
tenant pipeline only still works
managed pipeline only still works
tenant + managed still works
(the above wouldn't be part of a demo for the Thursday call, just for us)
tenant + final
managed + final
tenant + managed + final
only final - fails

controllers/release/adapter_test.go
seanconroy2021: FIxed

seanconroy2021:
> One very small typo then lgtm. I'd prefer to see a demo of the scenarios before we merge though. It would be nice to see tenant pipeline only still works managed pipeline only still works tenant + managed still works (the above wouldn't be part of a demo for the Thursday call, just for us) tenant + final managed + final tenant + managed + final only final - fails

I will show it in standup in the parking lot

controllers/release/adapter.go
```diff
@@ -751,20 +912,32 @@ func (a *adapter) validatePipelineSource() *controller.ValidationResult {
...
+
+	if releasePlan.Spec.FinalPipeline != nil {
+		pipelineRef = releasePlan.Spec.FinalPipeline.PipelineRef
+		if !a.releaseServiceConfig.Spec.Debug && pipelineRef.IsClusterScoped() {
+			a.release.MarkValidationFailed("tried using debug only options while debug mode is disabled in the ReleaseServiceConfig")
+			return &controller.ValidationResult{Valid: false}
+		}
+	}
+
+	if releasePlan.Spec.TenantPipeline == nil {
```
johnbieren, line 931: Confused why you added this. Using a tenant pipeline doesn't mean you can use any managed pipeline you want. Whether or not the tenant pipeline is nil has no bearing on the validation of the managed pipeline. If a managed pipeline is passed, it should be validated

johnbieren (changes requested):

controllers/release/adapter.go
seanconroy2021: I was writing the comment :)
So, the reason for the change is when you run a tenant pipeline, only the function will still be checking for a RPA, causing an error. Previously, we had the if-else which would only check one or the other .

seanconroy2021: Yeah that would make more sense

seanconroy2021: That works updated it

controllers/release/adapter.go
```diff
@@ -751,20 +912,32 @@ func (a *adapter) validatePipelineSource() *controller.ValidationResult {
...
+		}
+	}
+
+	if releasePlan.Spec.Target != "" {
 		releasePlanAdmission, err := a.loader.GetActiveReleasePlanAdmissionFromRelease(a.ctx, a.client, a.release)
 		if err != nil {
 			return a.validationError(err)
 		}
-
 		pipelineRef = releasePlanAdmission.Spec.Pipeline.PipelineRef
```
johnbieren, line 936: I think you still need a if pipeline != nil check before this line. I am now wondering what would happen if a ReleasePlan has a target and a matching RPA, but the RPA has no pipeline in it. I think that should fail somewhere, but no idea if we catch it anywhere. Probably not in scope of this PR though

seanconroy2021: Ok, I have added in the if condition.

seanconroy2021: I am curious about what you say above going to test it out

johnbieren: I created a ticket for it here https://issues.redhat.com/browse/RELEASE-1212

johnbieren (dismissed):

theflockers (commented):
Logic wise is LGTM but it has too many "codecov" complains about uncovered code. Some of them are indeed unreacheable, but some others are probably not. Could you crosscheck if you increase the test coverage? Then I review it again.

seanconroy2021:
> Logic wise is LGTM but it has too many "codecov" complains about uncovered code. Some of them are indeed unreacheable, but some others are probably not. Could you crosscheck if you increase the test coverage? Then I review it again.

Will do

theflockers:
I am only missing a Pipeline specific test, which is that the _final_ Pipeline should not be passed alone. In case it is, it should fail. 

Besides that, I see that the Tenant/Managed/Final pipeline testing might need improvement but this can be done later.

seanconroy2021:
> I am only missing a Pipeline specific test, which is that the _final_ Pipeline should not be passed alone. In case it is, it should fail.
> 
> Besides that, I see that the Tenant/Managed/Final pipeline testing might need improvement but this can be done later.

Ok, so I have added the test in :
```go
		It("should return false if only the Final Pipeline is set", func() {
			adapter.ctx = toolkit.GetMockedContext(ctx, []toolkit.MockData{
				{
					ContextKey: loader.ReleasePlanContextKey,
					Resource: &v1alpha1.ReleasePlan{
						ObjectMeta: metav1.ObjectMeta{
							Name:      "release-plan",
							Namespace: "default",
						},
						Spec: v1alpha1.ReleasePlanSpec{
							Application:            application.Name,
							ReleaseGracePeriodDays: 6,
							FinalPipeline:          parameterizedPipeline,
						},
					},
				},
			})
			result := adapter.validatePipelineDefined()
			Expect(result.Valid).To(BeFalse())
			Expect(result.Err).NotTo(HaveOccurred())
		})
```

api/v1alpha1/release_types.go
```diff
@@ -170,10 +170,9 @@ type Release struct {
 	Status ReleaseStatus `json:"status,omitempty"`
 }
 
-// HasEveryPostActionExecutionFinished checks whether the Release post-actions execution has finished,
-// regardless of the result.
-func (r *Release) HasEveryPostActionExecutionFinished() bool {
-	return r.hasPhaseFinished(postActionsExecutedConditionType)
+// Has FinalPipelineProcessingFinished checks whether the Release Final Pipeline processing has finished, regardless of the result.
```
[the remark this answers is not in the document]
seanconroy2021: Fixed ty

api/v1alpha1/release_types_test.go
```diff
@@ -30,35 +30,31 @@ import (
 
 var _ = Describe("Release type", func() {
 
-	When("HasEveryPostActionExecutionFinished method is called", func() {
+	When("HasFinalPipelineProcessingFinshed method is called", func() {
```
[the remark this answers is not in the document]
seanconroy2021: fixed ty

davidmogar (approved):

controllers/release/adapter_test.go
```diff
@@ -2342,6 +2924,28 @@ var _ = Describe("Release adapter", Ordered, func() {
 			Expect(result.Err).NotTo(HaveOccurred())
 		})
 
+		It("should return false if only the Final Pipeline is set", func() {
```
seanconroy2021, line 2927: @theflockers

johnbieren:
@seanconroy2021 can you rebase this so we can merge it?

johnbieren:
Yes, we have David's approval, and the only things that seem to have changed since mine are little nit picks. So let's merge it once it passes e2e. We'd like to have it in staging by a meeting tomorrow with UI

PR #583, opened 2024-10-09
fix(RELEASE-1150) add checks to fieldselectors

happybhati:
This commit fixes the issue of FieldSelectors retrurning everything when the argument is nill, added checks to each function using field selector.

### DIFF
--- a/loader/loader.go
+++ b/loader/loader.go
@@ -160,6 +160,11 @@ func (l *loader) GetMatchingReleasePlanAdmission(ctx context.Context, cli client
 // are in the namespace specified by the ReleasePlanAdmission's origin. If the List operation fails, an
 // error will be returned.
 func (l *loader) GetMatchingReleasePlans(ctx context.Context, cli client.Client, releasePlanAdmission *v1alpha1.ReleasePlanAdmission) (*v1alpha1.ReleasePlanList, error) {
+
+	if releasePlanAdmission.Spec.Origin == "" {
+		return nil, fmt.Errorf("releasePlanAdmission has no origin, so no ReleasePlans can be found")
+	}
+
 	releasePlans := &v1alpha1.ReleasePlanList{}
 	err := cli.List(ctx, releasePlans,
 		client.InNamespace(releasePlanAdmission.Spec.Origin),
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -250,6 +250,16 @@ var _ = Describe("Release Adapter", Ordered, func() {
 				return returnedObject != &v1alpha1.ReleasePlanList{} && err == nil && contains == false
 			})
 		})
+
+		It("fails to return release plans if origin is empty", func() {
+			modifiedReleasePlanAdmission := releasePlanAdmission.DeepCopy()
+			modifiedReleasePlanAdmission.Spec.Origin = ""
+
+			returnedObject, err := loader.GetMatchingReleasePlans(ctx, k8sClient, modifiedReleasePlanAdmission)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("releasePlanAdmission has no origin, so no ReleasePlans can be found"))
+			Expect(returnedObject).To(BeNil())
+		})
 	})
 
 	When("calling GetPreviousRelease", func() {


loader/loader.go
```diff
@@ -120,16 +120,16 @@ func (l *loader) GetMatchingReleasePlanAdmission(ctx context.Context, cli client
...
-		return nil, fmt.Errorf("releasePlan has no target, so no ReleasePlanAdmissions can be found")
-	}
-
 	releasePlanAdmissions := &v1alpha1.ReleasePlanAdmissionList{}
-	err := cli.List(ctx, releasePlanAdmissions,
-		client.InNamespace(releasePlan.Spec.Target),
-		client.MatchingFields{"spec.origin": releasePlan.Namespace})
-	if err != nil {
-		return nil, err
+	if releasePlan.Spec.Target != "" {
```
davidmogar, line 124: I think it's better to exit earlier just by checking for `Target == ""`. That way you avoid the else block altogether.

loader/loader.go
```diff
@@ -161,11 +161,15 @@ func (l *loader) GetMatchingReleasePlanAdmission(ctx context.Context, cli client
 // error will be returned.
 func (l *loader) GetMatchingReleasePlans(ctx context.Context, cli client.Client, releasePlanAdmission *v1alpha1.ReleasePlanAdmission) (*v1alpha1.ReleasePlanList, error) {
 	releasePlans := &v1alpha1.ReleasePlanList{}
-	err := cli.List(ctx, releasePlans,
-		client.InNamespace(releasePlanAdmission.Spec.Origin),
-		client.MatchingFields{"spec.target": releasePlanAdmission.Namespace})
-	if err != nil {
-		return nil, err
+	if releasePlanAdmission.Spec.Origin != "" {
```
davidmogar, line 164: Same as above.

loader/loader.go
```diff
@@ -182,11 +186,15 @@ func (l *loader) GetMatchingReleasePlans(ctx context.Context, cli client.Client,
 // If no previous Release is found, a NotFound error is returned.
 func (l *loader) GetPreviousRelease(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*v1alpha1.Release, error) {
 	releases := &v1alpha1.ReleaseList{}
-	err := cli.List(ctx, releases,
-		client.InNamespace(release.Namespace),
-		client.MatchingFields{"spec.releasePlan": release.Spec.ReleasePlan})
-	if err != nil {
-		return nil, err
+	if release.Namespace != "" {
```
davidmogar, line 189: Same as above.

loader/loader.go
happybhati: Ah, I probably misunderstood the ask in the acceptance criteria.

@davidmogar  do you mean same as we already have  [Line 123](https://github.com/konflux-ci/release-service/pull/583/files#diff-8e9af1e483632eac17610237caaa5f879ea7bfb3e1e1ea228e45a17c770b9432L123)?
Just need to add checks to other calls same way as done already for above.

johnbieren: I'm confused.This function seemingly already exits early if target is `""`
```
	if releasePlan.Spec.Target == "" {
		return nil, fmt.Errorf("releasePlan has no target, so no ReleasePlanAdmissions can be found")
	}

	releasePlanAdmissions := &v1alpha1.ReleasePlanAdmissionList{}
	err := cli.List(ctx, releasePlanAdmissions,
		client.InNamespace(releasePlan.Spec.Target),
		client.MatchingFields{"spec.origin": releasePlan.Namespace})
	if err != nil {
		return nil, err
	}
```
not sure there is anything to fix in this one?

loader/loader.go
johnbieren: I'm not sure this function really needs a change. Can namespace really be nil?

loader/loader.go
johnbieren: It fits the rest of the code to follow David's suggestion. Mostly just commenting here to say I think this is the only function that needs to change in this PR (got lucky I guess)

happybhati: ACK, on it.

johnbieren (approved):

loader/loader.go
happybhati: You are right. It's unnecessary. Thanks for pointing out.

happybhati:
Thanks, @johnbieren, for your approval.

@davidmogar Since you reviewed it earlier, please take a look whenever you get a chance.
Ready to be merged!

davidmogar (approved):

PR #615, opened 2024-11-22
fix: metrics still in wrong order

theflockers:
this PR fixes the order of metrics labels and values, and in addition modifies tests to use arbitrary strings intead of values from the label array, which was causing the test to not identify when the value was misplaced.


johnbieren (approved):
/retest

PR #632, opened 2024-12-10
fix: build pipeline task sha was removed

johnbieren:
One of the images used in the build pipeline was deleted, so this commit switches us to a different sha of the task.


seanconroy2021 (approved):
LGTM

PR #638, opened 2024-12-19
fix: update go.mod to fix cve CVE-2024-45338

johnbieren:
Non-linear parsing of case-insensitive content in golang.org/x/net/html in version < 0.33.0.


seanconroy2021 (approved):
lgtm

PR #655, opened 2025-01-24
fix(RELEASE-1277): release controller crashes

theflockers:
this PR tries to fix/decrease the daily crashes of the release controller by filtering objects to cache and increasing the lease renew time.

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"flag"
 	"os"
+	"time"
 
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	crwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -27,6 +28,7 @@ import (
 	"github.com/konflux-ci/operator-toolkit/controller"
 	"github.com/konflux-ci/operator-toolkit/webhook"
 	"github.com/konflux-ci/release-service/api/v1alpha1/webhooks"
+	"github.com/konflux-ci/release-service/metadata"
 
 	"go.uber.org/zap/zapcore"
 
@@ -37,10 +39,13 @@ import (
 	ecapiv1alpha1 "github.com/enterprise-contract/enterprise-contract-controller/api/v1alpha1"
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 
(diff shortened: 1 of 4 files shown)


theflockers:
this PR replaces: https://github.com/konflux-ci/release-service/pull/545 and https://github.com/konflux-ci/release-service/pull/617

metadata/labels.go
```diff
@@ -41,6 +44,9 @@ var (
 	// AutomatedLabel is the label name for marking a Release as automated
 	AutomatedLabel = fmt.Sprintf("release.%s/automated", rhtapDomain)
 
+	// ServiceNameLabel is the label used to specify the service associated with an object
+	ServiceNameLabel = fmt.Sprintf("%s/%s", rhtapDomain, "service")
```
johnbieren, line 48: This would be `appstudio.openshift.io/service` how does that help us?

main.go
```diff
@@ -87,9 +105,19 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&tektonv1.PipelineRun{}: cache.ByObject{
+					Label: labels.SelectorFromSet(labels.Set{metadata.ServiceNameLabel: metadata.ServiceName}),
```
johnbieren, line 111: Is it possible to put an OR in here and then just reuse the managed, tenant, and final pipeline labels without having to add a new one?

theflockers: It allows a list of Labels. So your idea would be we filter the value by `metadata.TenantPipelineType`,  `metadata.ManagedPipelineType` or `metadata.FinalPipelineType` ? Is that correct?

metadata/labels.go
theflockers: It would (from what I understood) add the label as `appstudio.openshift.io/service: release` to Pipelines created by the operator (adapter lines 494, 534 and 576), so the cache would only consider those ones. I did not put much brain on the label name as I basically copied this from another PR as this looked enough to "categorize" what should be cached.

johnbieren: Maybe you meant `s/"service"/ServiceName/` then? You are just combining rhtapDomain variable with the string "service" right now

theflockers: If the suggested replacement is made, the label would look like in the end `appstudio.openshift.io/service: release`, as what adds it to the PipelineRun is:
```
WithLabels(map[string]string{
...
metadata.ServiceNameLabel: metadata.ServiceName,
}).
```
so it adds the label as expected, correct? I understood that the  ServiceNameLabel" is the "key" and the ServiceName (`release`) is the value - and the WithLabel makes it `appstudio.openshift.io/service: release`

https://github.com/konflux-ci/release-service/pull/655/files#diff-6f40aa8049566b39f0c38b10d839c98d6b9298c98d783c4b82d8e6ba368067d4R494

johnbieren: `fmt.Sprintf("%s/%s", rhtapDomain, "service")` you are combining the variable `rhtapDomain` with the *string* "service" look at the quotes. Nothing is adding `release` to the label

johnbieren: Gotcha. I thought your intent was to have `release` in the label name, not just the value

johnbieren (dismissed):

main.go
theflockers: I don´t understand much about the collectors, but sounds like a single label would make it easier. I don´t see (at least so far) a real need for different labels for the cache. Unless we decide in the future we want to cache only  managed, for instance. 

So far it seems enough a single one, even though we are having to add it.  For the moment I would just leave it as it looks now and change it if we understand it is required.

main.go
```diff
@@ -87,9 +105,19 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
```
davidmogar, line 109: The reason I never went for this change is that I wasn't sure of the implication of it as this would be a partial cache. More resources apart from PipelineRuns are cached (eg. Releases) and I'm not sure if this would cause only PipelineRuns to be cached or it's just a filter. It was easier to understand in previous versions of controller-runtime for sure...

Probably this is fine though and this is just a selector prior caching PipelineRuns.

theflockers: Nice one. I found this in the docs:
```
IMPORTANT NOTE Requests to a client backed by a filtered cache for objects that do not match the filter will never return anything. In other words, filtered caches make the filtered-out objects invisible to the client.
```

IIUC we need to add all other objects we want to cache here as well. I think the PipelineRun is the only that needs an actual filter, while the other ones should be here, but (somehow) without a filter. I will check how to do it.

theflockers: Anyhow, I am waiting to see how e2e will behave with this change.

davidmogar: We should implement filters for Releases (not now but in the future). You will also need to list ReleasePlans and ReleasePlanAdmissions and probably ReleaseServiceConfigs, Applications and Snapshots as I think those get cached as well. Not sure about the latter. Resources involved in controllers are cached for sure.

theflockers: Apparently, only objects that are watched by the controllers needs to be added to the manager cache, but I will test both cases (with all objects we need vs all objects we watch) and see how it behaves.

main.go
```diff
@@ -87,9 +105,29 @@ func main() {
...
+					Label: labels.SelectorFromSet(labels.Set{metadata.ServiceNameLabel: metadata.ServiceName}),
+				},
+				// also cache other watched objects, but no filter is required.
+				&appstudiov1alpha1.Release{}:              {},
+				&appstudiov1alpha1.ReleasePlan{}:          {},
+				&appstudiov1alpha1.ReleasePlanAdmission{}: {},
+				// objects that the operator does not watch, but are used by it.
+				&appstudiov1alpha1.ReleaseServiceConfig{}: {},
+				&applicationapiv1alpha1.Snapshot{}:        {},
+				&applicationapiv1alpha1.Component{}:       {},
```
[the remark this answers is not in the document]
theflockers: removed.

johnbieren (commented):
I see E2E passed, so we are good with this now? Can you squash the commits?

theflockers:
> I see E2E passed, so we are good with this now? Can you squash the commits?

I will merge. But lets put it only on stage for now, so I can look for some numbers.

johnbieren (approved):

PR #666, opened 2025-02-07
feat(RELEASE-1434): allow to override volume type

davidmogar:
This is a provisional change that will be used during the transition to using Trusted Artificts in the
release-service-catalog pipelines.

Trusted Artifacts require to use an emptyDir in opposition to the PVCs used today. This change allow to override the volume type used by a given pipeline by defining a map in the ReleaseServiceConfig which maps a string formed like "<url>/<pathInRepo>" to a volume type. Those values are only found in pipeline refs using the git resolver. Any other resolver gets ignored.


config/crd/bases/appstudio.redhat.com_releaseserviceconfigs.yaml
```diff
@@ -66,6 +66,16 @@ spec:
                       tasks
                     type: string
                 type: object
+              volumeOverrides:
+                additionalProperties:
+                  description: VolumeType is the kind of volume to be used for the
+                    managed release PipelineRun referencing this Pipeline.
+                  enum:
+                  - emptyDir
```
mmalina, line 74: We should have a value for the other type as well, no? Otherwise why have a value at all - it could just be a list of refs that use emptyDir.

tekton/utils/pipeline.go
```diff
@@ -78,6 +78,30 @@ func (pr *PipelineRef) GetRevision() (string, error) {
...
+		case "pathInRepo":
+			pathInRepo = param.Value
+		}
+	}
+
+	if url == "" || pathInRepo == "" {
+		return "", fmt.Errorf("unexpected git ref params")
+	}
+
+	return fmt.Sprintf("%s/%s", url, pathInRepo), nil
```
mmalina, line 102: Shouldn't this be based on just the `pathInRepo`? Because repo url could be somebody's fork, for example for e2e runs, or occasionally when people point to a fork repo in their RPA (but ideally that shouldn't happen and is pretty rare). So if `url` is included, we would need an entry in the config for each fork.

scoheb:
can you please include a snippet of a `ReleaseServiceConfig` for context?

config/crd/bases/appstudio.redhat.com_releaseserviceconfigs.yaml
davidmogar: The other value is the default and it's not checked in code. Having a single value is still better than nothing as we can use a map instead of an array. It's faster. If in the future we want to support both, we could expand it.

tekton/utils/pipeline.go
scoheb: I am not sure we need `url`. but now that I think about it...we'll have an issue when we merge to **development** if we only look at `pathInRepo` ... staging nor production will not have been converted. Which means we can only update `ReleaseServiceConfig` when the change has been promoted all the way to production. Not sure that works either.

johnbieren: We talked about this and are hoping to do something like this
```
Stage
*/release-service-catalog.git 	*	pipelines/managed/e2e/e2e.yaml

Prod
https://github.com/konflux-ci/release-service-catalog.git	development	pipelines/managed/e2e/e2e.yaml
https://github.com/konflux-ci/release-service-catalog.git	staging		pipelines/managed/e2e/e2e.yaml
https://github.com/konflux-ci/release-service-catalog.git	development	pipelines/managed/fbc-release/fbc-release.yaml
https://github.com/konflux-ci/release-service-catalog.git	*		pipelines/managed/rh-advisories/rh-advisories.yaml
```
where if a pipeline is listed, it uses emptyDir. We need wildcards, especially in stage, for forks and branches for e2e

davidmogar: Ok, this is changed to a list of structs which allows to set the values for url, revision and pathInRepo. All the values are required but both url and revision allows you to use regex expressions so you can set both to `.*` and it will match any value.

tekton/utils/pipeline.go
```diff
@@ -78,6 +78,30 @@ func (pr *PipelineRef) GetRevision() (string, error) {
 	return "", fmt.Errorf("no revision found")
 }
 
+// GetPipelineGitRef returns a string containing the url to a Git repo and the path to a Pipeline within that repo.
+// If the git resolver is not used or no url and a pathInRepo params are supplied, the function will fail.
+func (pr *PipelineRef) GetPipelineGitRef() (string, error) {
```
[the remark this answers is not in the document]
davidmogar: Added a test for the function replacing this one.

davidmogar:
> can you please include a snippet of a `ReleaseServiceConfig` for context?

You mean an example?

```
apiVersion: appstudio.redhat.com/v1alpha1
kind: ReleaseServiceConfig
metadata:
  name: releaseserviceconfig-sample
spec:
  emptyDirOverrides:
     - url: ".*"
       revision: ".*"
       pathInRepo: "pipelines/my-pipeline.yaml"
```

api/v1alpha1/releaseserviceconfig_types.go
```diff
@@ -52,6 +72,27 @@ type ReleaseServiceConfig struct {
 	Status ReleaseServiceConfigStatus `json:"status,omitempty"`
 }
 
+// IsPipelineOverridden checks whether there is a EmptyDirOverride matching the given url, revision and pathInRepo.
+func (rsc *ReleaseServiceConfig) IsPipelineOverridden(url, revision, pathInRepo string) bool {
```
johnbieren, line 76: Should probably have a test for this

johnbieren (commented):
LGTM besides missing test

api/v1alpha1/releaseserviceconfig_types.go
davidmogar: Added.

johnbieren (approved):

config/crd/bases/appstudio.redhat.com_releaseserviceconfigs.yaml
davidmogar: Resolving this as Martin is off and the PR is already approved.

scoheb:
provision error

/retest

scoheb:
[INFO] Updating existing secret 'sealights-token' in namespace 'release-service'.
secret "sealights-token" deleted
error: failed to create secret Post "https://api.kx-2c7c3f2aed.ajj5.p3.openshiftapps.com:443/api/v1/namespaces/release-service/secrets?fieldManager=kubectl-create&fieldValidation=Strict": dial tcp: lookup api.kx-2c7c3f2aed.ajj5.p3.openshiftapps.com on 172.30.0.10:53: no such host
Error: reached maximum number of attempts (2). error: exit status 1
make: *** [Makefile:25: ci/test/e2e] Error 1

scoheb:
> > can you please include a snippet of a `ReleaseServiceConfig` for context?
> 
> You mean an example?
> 
> ```
> apiVersion: appstudio.redhat.com/v1alpha1
> kind: ReleaseServiceConfig
> metadata:
>   name: releaseserviceconfig-sample
> spec:
>   emptyDirOverrides:
>      - url: ".*"
>        revision: ".*"
>        pathInRepo: "pipelines/my-pipeline.yaml"
> ```

needs to be `EmptyDirOverrides` with a capital `E`

johnbieren:
> > > can you please include a snippet of a `ReleaseServiceConfig` for context?
> > 
> > 
> > You mean an example?
> > ```
> > apiVersion: appstudio.redhat.com/v1alpha1
> > kind: ReleaseServiceConfig
> > metadata:
> >   name: releaseserviceconfig-sample
> > spec:
> >   emptyDirOverrides:
> >      - url: ".*"
> >        revision: ".*"
> >        pathInRepo: "pipelines/my-pipeline.yaml"
> > ```
> 
> needs to be `EmptyDirOverrides` with a capital `E`

Yeah because of `EmptyDirOverrides []EmptyDirOverrides `json:"EmptyDirOverrides,omitempty"`` .. probably should have made the json tag lowercase like we do everywhere else

PR #669, opened 2025-02-10
fix(RELEASE-1441): remove rsc from being managed

scoheb:
- the rsc needs to managed outside of the release service so that clusters can have a specific version for testing.
- A version can be deployed and managed by ArgoCD via infra-deployments for a specific cluster

### DIFF
--- a/config/manager/kustomization.yaml
+++ b/config/manager/kustomization.yaml
@@ -1,7 +1,6 @@
 resources:
 - manager.yaml
 - network_policy.yaml
-- release_service_config.yaml
 
 generatorOptions:
   disableNameSuffixHash: true


controllers/release/adapter.go
```diff
@@ -90,15 +90,24 @@ func (a *adapter) EnsureConfigIsLoaded() (controller.OperationResult, error) {
...
+	a.logger.Info("Getting Release Service Config")
 	a.releaseServiceConfig, err = a.loader.GetReleaseServiceConfig(a.ctx, a.client, v1alpha1.ReleaseServiceConfigResourceName, namespace)
 	if err != nil && !errors.IsNotFound(err) {
 		return controller.RequeueWithError(err)
 	}
 
 	if err != nil {
+		a.logger.Info("Creating Empty Release Service Config")
 		a.releaseServiceConfig = a.getEmptyReleaseServiceConfig(namespace)
+		patch := client.MergeFrom(a.releaseServiceConfig.DeepCopy())
```
davidmogar, line 102: There's no need to persist it. If we really want to persist it maybe that should be done in a different place (whenever the manager is created and not per reconciliation).

scoheb: We need to persist it. The release pipelines want to do a `oc get RSC/release-service-config` in collect-data

scoheb:
still getting `Error from server (NotFound): releaseserviceconfigs.appstudio.redhat.com "release-service-config" not found`

johnbieren (commented):
This PR says it is removing the RSC from being managed. Then I look at the diff and we are adding rbac and updating the adapter

config/rbac/releaseserviceconfig_role_binding.yaml
johnbieren, line 1: Why? We have this https://github.com/redhat-appstudio/infra-deployments/blob/main/components/release/base/release-service-config-rbac.yaml

config/rbac/kustomization.yaml
johnbieren, line 1: What is changing that now we need this and we didn't before?

scoheb:
> This PR says it is removing the RSC from being managed. Then I look at the diff and we are adding rbac and updating the adapter

Essentially I was going down the road of wanting to use volumeTypes on stage. But the RSC is part of the release service bundle, so it can only be changed if you release a new version of the service.
So I removed it from the service and placed it in infra deployments. I got it work, and was able to update the RSC for only the staging env via Argo and infra deployments.
But then E2E was failing. The reason is that without infra deployments, there is no RSC now.
The pipelines were failing with an error "not able to get RSC"
So I tried to persist the empty RSC when one is not found.

controllers/release/adapter.go
johnbieren: The PR says it is moving the RSC to being managed in infra-deployments instead of here. So, we are still creating a CR per env. Why do we now need to create it in the adapter?

johnbieren:
> > This PR says it is removing the RSC from being managed. Then I look at the diff and we are adding rbac and updating the adapter
> 
> Essentially I was going down the road of wanting to use volumeTypes on stage. But the RSC is part of the release service bundle, so it can only be changed if you release a new version of the service. So I removed it from the service and placed it in infra deployments. I got it work, and was able to update the RSC for only the staging env via Argo and infra deployments. But then E2E was failing. The reason is that without infra deployments, there is no RSC now. The pipelines were failing with an error "not able to get RSC" So I tried to persist the empty RSC when one is not found.

E2E isn't using infra-deployments? How are we using `DEPLOY_ONLY` in e2e then? What is it using to install?

davidmogar:
> > E2E isn't using infra-deployments? How are we using `DEPLOY_ONLY` in e2e then? What is it using to install?
> 
> I see! I need to get [redhat-appstudio/infra-deployments#5508](https://github.com/redhat-appstudio/infra-deployments/pull/5508) merged first.

That helps... But yeah, don't persist the RSC in that operation please.

controllers/release/adapter.go
```diff
@@ -90,15 +90,24 @@ func (a *adapter) EnsureConfigIsLoaded() (controller.OperationResult, error) {
 	}
 
 	var err error
+	a.logger.Info("Getting Release Service Config")
```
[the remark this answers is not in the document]
scoheb: removed logging

controllers/release/adapter.go
```diff
@@ -90,15 +90,24 @@ func (a *adapter) EnsureConfigIsLoaded() (controller.OperationResult, error) {
...
 		a.releaseServiceConfig = a.getEmptyReleaseServiceConfig(namespace)
+		patch := client.MergeFrom(a.releaseServiceConfig.DeepCopy())
+		err := a.client.Patch(a.ctx, a.releaseServiceConfig, patch)
+		if err != nil && !errors.IsNotFound(err) {
+			return controller.RequeueWithError(err)
+		}
+		a.logger.Info("Persisted Empty Release Service Config")
 	}
 
+	a.logger.Info("Release Service Config Loaded")
```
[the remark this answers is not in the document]
scoheb: removed logging

controllers/release/adapter.go
scoheb: removed persistance

config/rbac/releaseserviceconfig_role_binding.yaml
scoheb: removed

scoheb:
also created https://github.com/konflux-ci/konflux-ci/pull/1285

PR #681, opened 2025-02-24
feat(RELEASE-1432): migrate to the Trusted artifact version

jinqi7:


johnbieren (approved):
seems sane to me

jinqi7:
Seems we need to force merge the PR since it can't pass EC for untrusted pipelineTask.

johnbieren:
> Seems we need to force merge the PR since it can't pass EC for untrusted pipelineTask.

? No? If we force merge a PR with an untrusted pipeline task, then every single future release-service PR is going to fail for the untrusted pipeline task. The task needs to be listed as trusted by EC before we merge it

.tekton/release-service-pull-request.yaml
```diff
@@ -174,35 +161,34 @@ spec:
...
+        value: $(params.output-image).prefetch
+      - name: ociArtifactExpiresAfter
+        value: $(params.image-expires-after)
       runAfter:
       - clone-repository
       taskRef:
         params:
         - name: name
-          value: prefetch-dependencies
+          value: prefetch-dependencies-oci-ta
```
flacatus, line 181: prefetch-dependencies. Lets keep the name to original to pass the EC checks

jinqi7: Original name?  It's the  task name in the bundle of "quay.io/konflux-ci/tekton-catalog/task-prefetch-dependencies-oci-ta".  And it's being used in the tekton files for integration-service and build-definitions [here](https://github.com/flacatus/build-service/blob/4a98c2e92733bd1d1bc20fe47402ede1c1f34e32/.tekton/build-service-pull-request.yaml#L214C18-L214C46).

jinqi7:
> > Seems we need to force merge the PR since it can't pass EC for untrusted pipelineTask.
> 
> ? No? If we force merge a PR with an untrusted pipeline task, then every single future release-service PR is going to fail for the untrusted pipeline task. The task needs to be listed as trusted by EC before we merge it

I don't understand why the task "prefetch-dependencies-oci-ta" failed in EC check of release service and works in other services.   such as: the [task](https://github.com/konflux-ci/integration-service/blob/00b6164e5bf2316b1fc8b816709c5273e29a21a0/.tekton/integration-service-pull-request.yaml#L207C18-L207C46) in integration service.

.tekton/release-service-pull-request.yaml
jinqi7: Found and fix the issue - the version of the bundle need to change to "0.2" .

jinqi7:
> > Seems we need to force merge the PR since it can't pass EC for untrusted pipelineTask.
> 
> ? No? If we force merge a PR with an untrusted pipeline task, then every single future release-service PR is going to fail for the untrusted pipeline task. The task needs to be listed as trusted by EC before we merge it
 
Fixed the issue. The version of the bundle is updated.

johnbieren (approved):

PR #685, opened 2025-02-26
fix: reduce description length

davidmogar:
The CRDs are growing and this is causing problems when installing them (at least I've seen this problem while using kind). This change reduces the description length which seems to be enough to get rid of the problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd:maxDescLen=0 webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 .PHONY: generate
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.


Makefile
```diff
@@ -100,7 +100,7 @@ help: ## Display this help.
 
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
-	$(CONTROLLER_GEN) rbac:roleName=manager-role crd webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) rbac:roleName=manager-role crd:maxDescLen=0 webhook paths="./..." output:crd:artifacts:config=config/crd/bases
```
mmalina, line 103: What is the effect of this? The description will be removed from the CRDs completely? Meaning if you did `kubectl get crd Release` or similar, you won't see any description at all? That doesn't sound good.

johnbieren: Especially since in the arch call yesterday they were talking about CRD documentation

davidmogar: Pretty much. I have the feeling dev environment is broken right now because of https://github.com/konflux-ci/release-service/commit/0156333cc6f8fa6aa491b9236964eecee0b99720. Yesterday I was testing something in a kind cluster I couldn't even deploy because the RP CRD was too big. I'm investigating the dev issue right now.

davidmogar: So there's not actual problem in development. It was just a problem introduced with Jings PR (that's what I was checking and what lead me to think that could be related with this other problem).

The error I see in Kind is something like (not exactly as I don't have it at hand):
```
The ConfigResourceDefinition "releaseplan" is invalid: metadata.annotations: Too long: must have at most 262144 characters
```
It seems that this can be avoided by running `kubectl create` instead of `apply` as while doing this, the `last-applied-configuration` annotation is not populated. This is actually the one causing the error and not the size of the CRD. I guess that Argo/Kustomize uses create while deploying. That's why I don't see the error in Konflux clusters.

We might close this for now, and for those using kind to testing just remember to use the flag above when running `make install`.

johnbieren: Should we add a new make target with the flag for running on kind?

davidmogar: Maybe. Let's discuss it tomorrow during the Sip & Share maybe. WDYT?

johnbieren: We don't have sip and share tomorrow, it is day of learning. I mean, why wouldn't we want a target that works on kind if the normal ones do not?

PR #700, opened 2025-03-19
fix: change port to 8080 and endpoint to insecure

theflockers:
this PR changes the metrics port to 8080 as the service
in this port is not secured in the pod side

### DIFF
--- a/config/default/manager_metrics_patch.yaml
+++ b/config/default/manager_metrics_patch.yaml
@@ -1,4 +1,4 @@
 # This patch adds the args to allow exposing the metrics endpoint using HTTPS
 - op: add
   path: /spec/template/spec/containers/0/args/0
-  value: --metrics-bind-address=:8443
+  value: --metrics-bind-address=:8080
--- a/config/default/metrics_service.yaml
+++ b/config/default/metrics_service.yaml
@@ -9,9 +9,9 @@ metadata:
   namespace: system
 spec:
   ports:
-  - name: https
-    port: 8443
+  - name: http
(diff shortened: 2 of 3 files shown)


config/default/manager_metrics_patch.yaml
```diff
@@ -1,4 +1,4 @@
 # This patch adds the args to allow exposing the metrics endpoint using HTTPS
 - op: add
   path: /spec/template/spec/containers/0/args/0
-  value: --metrics-bind-address=:8443
+  value: --metrics-bind-address=:8080
```
mmalina, line 4: Why can't it work properly with https and 8443 and without insecureSkipVerify?

theflockers: As mentioned in the standup, the controller is apparently missing the a parameter to set it to secure: https://github.com/konflux-ci/release-service/blob/e4dba7cf889934d0f2559e0e96e762528955e6ce/main.go#L135C18-L135C31

Maybe it uses some pre-defined option but we would need to investigate and in addition, discuss if self-signed certificates are enough or we should request valid certificates for the endpoint. I would suspect the later would be enough.

theflockers: Ah the and also because the "ServiceMonitor" is expecting it to be https, so I think it makes more sense instead of just changing the `scheme` to http, change the port as well for consistency.

mmalina: ok

mmalina (approved):

PR #701, opened 2025-03-19
fix: update task ref for build-sealights-container

scoheb:
- use latest trusted sha for the buildah-oci-ta task


scoheb:
merging since this does not affect PRs ... only pushes.

PR #715, opened 2025-04-02
feat(RELEASE-1514): pass new parameter

scoheb:
- All pipelines will now support the optional parameter `verify_ec_task_git_revision`. It is needed for consuming the verify-conform task via git resolver
- The release controller will now take the value of the existing key from the ec configmap and pass as a parameter


controllers/release/adapter_test.go
johnbieren, line 1: Can you add a test like the one on line 2410 in this file?

scoheb: I updated the existing test.

PR #730, opened 2025-04-16
feat(RELEASE-1464): add RoleBindings for collector secrets

seanconroy2021:
* Introduce a new 'secrets' field in the collector struct as a list of secrets.
* Before creating a collector pipelineRun, automatically create RoleBindings that allow the collector's serviceAccount to access each secret listed.
* After pipelineRun completion, ensure that the corresponding RoleBindings are removed.

_**Notes:**_
The release-service-controller-manager SA has to have permissions to manage roles/rolebindings and also be able to get secrets. Link to working ClusterRole and ClusterRoleBinding that was used for testing [here](https://privatebin.corp.redhat.com/?7d796a1676921708#A3RyPd4C9iPXEWKRiZCvYbwhodRWYn2uPrcR2bawFtfi)

Testing example [here](https://github.com/seanconroy2021/sc-konflux-demos/tree/main/collector-rb)


loader/loader.go
```diff
@@ -230,9 +232,51 @@ func (l *loader) GetRelease(ctx context.Context, cli client.Client, name, namesp
 	return release, toolkit.GetObject(name, namespace, cli, ctx, release)
 }
 
-// GetRoleBindingFromReleaseStatus returns the RoleBinding associated with the given Release. That association is defined
+// GetRoleBindingFromTenantCollectors returns the Tenant Collector RoleBinding associated with the given Release. That association is defined
 // by the namespaced name stored in the Release's status.
-func (l *loader) GetRoleBindingFromReleaseStatus(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*rbac.RoleBinding, error) {
+func (l *loader) GetRoleBindingFromTenantCollectors(ctx context.Context, cli client.Client, release *v1alpha1.Release) (*rbac.RoleBinding, error) {
```
johnbieren, line 237: Do we really need 3 separate functions for this? Wouldn't it reduce duplicated code to instead have this function take
`func (l *loader) GetRoleBindingFromReleaseStatusPipelineInfo(ctx context.Context, cli client.Client, pipelineInfo *v1alpha1.PipelineInfo) (*rbac.RoleBinding, error) {` and then just call it like `roleBinding, err = a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, a.release.Status.[CollectorsProcessing.TenantCollectorsProcessing, CollectorsProcessing.ManagedCollectorsProcessing, ManagedProcessing])`?

seanconroy2021: Yeah, you’re correct refactored it into a single function.

johnbieren (approved):

PR #746, opened 2025-05-19
feat(RELEASE-1474): create RoleBindings for collector pipeline SA

seanconroy2021:
Before a collector pipeline executes, the release service creates RoleBindings for the specified SAs to ensure proper permissions to allow Collectors to run.


controllers/release/adapter.go
```diff
@@ -198,7 +198,17 @@ func (a *adapter) EnsureManagedCollectorsPipelineIsProcessed() (controller.Opera
 		return controller.RequeueWithError(err)
 	}
 
-	roleBinding, _ := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.CollectorsProcessing.ManagedCollectorsProcessing)
+	tenantRoleBinding, _ := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.CollectorsProcessing.ManagedCollectorsProcessing, "tenant")
```
johnbieren, line 201: why does the managed collectors plr care about a tenant rolebinding?

controllers/release/adapter.go
```diff
@@ -416,7 +446,7 @@ func (a *adapter) EnsureManagedPipelineIsProcessed() (controller.OperationResult
 		return controller.RequeueWithError(err)
 	}
 
-	roleBinding, _ := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.ManagedProcessing)
+	tenantRoleBinding, _ := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.ManagedProcessing, "tenant")
```
johnbieren, line 449: can you add a comment explaining that this is so that the managed plr can access the resources in the tenant ns?

controllers/release/adapter.go
```diff
@@ -670,22 +700,37 @@ func (a *adapter) EnsureCollectorsProcessingResourcesAreCleanedUp() (controller.
...
+		return controller.RequeueOnErrorOrContinue(err)
+	}
+
+	secretRoleBinding, err := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.CollectorsProcessing.TenantCollectorsProcessing, "secret")
 	if err != nil && !errors.IsNotFound(err) && !strings.Contains(err.Error(), "valid reference to a RoleBinding") {
 		return controller.RequeueOnErrorOrContinue(err)
 	}
 
-	err = a.cleanupProcessingResources(nil, roleBinding)
+	err = a.cleanupProcessingResources(nil, tenantRoleBinding, secretRoleBinding)
```
johnbieren, line 713: why isn't the managed one cleaned up here?

johnbieren: Oh, I am just now realizing that the difference here is the TenantCollectorsProcessing vs ManagedCollectorsProcessing. Can you make this more clear with the variable names? Using `tenantRoleBinding` twice for two different RoleBindings in the same function is confusing IMO

controllers/release/adapter.go
seanconroy2021: Fixed

controllers/release/adapter.go
seanconroy2021: Fixed and added a comment to make it more clear

controllers/release/adapter.go
```diff
@@ -198,7 +198,17 @@ func (a *adapter) EnsureManagedCollectorsPipelineIsProcessed() (controller.Opera
 		return controller.RequeueWithError(err)
 	}
 
-	roleBinding, _ := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.CollectorsProcessing.ManagedCollectorsProcessing)
+	tenantRoleBinding, _ := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.CollectorsProcessing.ManagedCollectorsProcessing, "tenant")
+	if err != nil && !errors.IsNotFound(err) && !strings.Contains(err.Error(), "valid reference to a RoleBinding") {
```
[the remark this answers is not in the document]
seanconroy2021: Yeah, I get what you're saying and you're right that relying on strings.Contains is a bit fragile. That said, in this case, the error message always comes from loader.GetRoleBindingFromReleaseStatusPipelineInfo, and it's always the same string. We actually use this pattern in quite a few places below as well.

I'm open to changing it to a custom error type, but I think we should ask maybe @johnbieren since it's used in so many places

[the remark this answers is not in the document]
FilipNikolovski: > > Edit: noticed you're also checking the previous error here, and not the error returned from GetRoleBindingFromReleaseStatusPipelineInfo, not sure if that's intended or not.
> 
> The reason for if `err != nil && !errors.IsNotFound(err) && !strings.Contains(err.Error(), "valid reference to a RoleBinding")` is mainly to avoid unhandled errors showing up from a Github Action. We’re not actually using the err value itself later, since we check if tenantRoleBinding is nil further down. So this is intentional and follows the same used in `EnsureManagedPipelineIsProcessed`, which this is based on ... this is from what I know anyway :)

It would be better to just omit the if condition then, since the check will always fail because you're checking the previous error value and not the one from GetRoleBindingFromReleaseStatusPipelineInfo. You could add a lint ignore comment to ignore those errors from the gh action probably? Not sure why we're ignoring this error though, isn't it better to just handle it here?

> Yeah, I get what you're saying and you're right that relying on strings.Contains is a bit fragile. That said, in this case, the error message always comes from loader.GetRoleBindingFromReleaseStatusPipelineInfo, and it's always the same string. We actually use this pattern in quite a few places below as well.
> 
> I'm open to changing it to a custom error type, but I think we should ask maybe @johnbieren since it's used in so many places

This is just a minor nitpick, if it's too much hassle to refactor we could leave it as-is.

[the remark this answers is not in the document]
seanconroy2021: I think we can do [// #nosec G104](https://github.com/securego/gosec/issues/270#issuecomment-454039763)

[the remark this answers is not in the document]
johnbieren: If there is desire to change similar code that is repeated in many places, the best way to approach that is to open a jira for it and we can tackle it as tech debt IMO

johnbieren, line 202: Yeah, I'm confused here. Line 201 doesn't set `err`, it is last set on line 196. `if err != nil && !errors.IsNotFound(err)` is already checked on line 197, so that is redundant here, but also, line 196 gets a plr, of course the error won't contain something about the roleBinding. Am I missing something? It looks like you meant to do `tenantRoleBinding, err` not `_`
(This applies to line 207 and 212 too. I recognize that is how the code was, but it looks like it was wrong)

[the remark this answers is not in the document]
johnbieren: >is mainly to avoid unhandled errors showing up from a Github Action 

huh? what github action?

controllers/release/adapter.go
```diff
@@ -416,7 +446,8 @@ func (a *adapter) EnsureManagedPipelineIsProcessed() (controller.OperationResult
 		return controller.RequeueWithError(err)
 	}
 
-	roleBinding, _ := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.ManagedProcessing)
+	// Get RoleBinding from tenant namespace so the managed PipelineRun can access tenant resources.
+	tenantRoleBinding, _ := a.loader.GetRoleBindingFromReleaseStatusPipelineInfo(a.ctx, a.client, &a.release.Status.ManagedProcessing, "tenant")
 	if err != nil && !errors.IsNotFound(err) && !strings.Contains(err.Error(), "valid reference to a RoleBinding") {
```
johnbieren, line 451: same thing

controllers/release/adapter.go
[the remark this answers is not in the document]
seanconroy2021: https://github.com/konflux-ci/release-service/security/code-scanning/6
It's GoSec Action. It flags unhandled errors, even if we intentionally ignore them with _.

seanconroy2021: Yeah, it looks wrong. I’m also trying to figure out whether we should actually be checking the error here or if that part should just be removed. If we do capture and check the error, it would make sense to reconcile if it failed, but maybe there's a reason why it's not doing that. Going to change it and test it locally.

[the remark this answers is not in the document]
FilipNikolovski: Looks good!

FilipNikolovski (approved):
lgtm :rocket:

controllers/release/adapter.go
seanconroy2021: see comment here: https://github.com/konflux-ci/release-service/pull/746#discussion_r2124355351

theflockers (approved):

PR #759, opened 2025-06-12
fix: use predefined name for kind kubeconfig secret and bump to 0.2

flacatus:
Currently, in Tekton, finally tasks that are dependent on the results of a preceding task will not be scheduled if that preceding task is cancelled or fails before generating the expected result. Basically changing the result to a generated param..

### DIFF
--- a/integration-tests/pipelines/konflux-e2e-tests-pipeline.yaml
+++ b/integration-tests/pipelines/konflux-e2e-tests-pipeline.yaml
@@ -98,10 +98,12 @@ spec:
           - name: revision
             value: main
           - name: pathInRepo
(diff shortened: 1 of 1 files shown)


adrianriobo (commented):
I see this using provision 0.1 is if tine,  AFAIK the  kubeconfig scan issue is sovled on mapt 0.9.2 which is task 0.2 

also decommission should match version too

psturc (approved):

johnbieren:
> @johnbieren I changed recently the laptop and forgot about setting the signature

np! I see it's there now and rebased - once the checks pass again I will merge

johnbieren (approved):

PR #776, opened 2025-07-01
feat(RELEASE-1709): filter RP by RPA label when provided

seanconroy2021:
Update the logic for `GetMatchingReleasePlans` previously if labelled was provided it would still match all RP:

* When a RPA label is provided on RPs, the GetMatchingRP function now only returns RPs that have a matching label. This stop the CR status from showing matches to all RPs.
* If no labels are present it converts to normal matching.
* ReleasePlans with a label will only be included in matching RPA removed from global RPA.
* Updated a bug in tests where it was missing `Should(BeTrue())`

**Testing:** 
So we have two RPs (00, 01)  set with the label 'hello-world-app-rpa-00', while we have two other RPs (02,03) with no label.

RPA: hello-world-app-rpa-00 set labels in RPs so we should see 00, 01 
```
  releasePlans:
  - name: user-ns2/hello-world-app-rp-00
  - name: user-ns2/hello-world-app-rp-01
```

RPA:  hello-world-app-rpa-02 no set labels, so we should see 02, 03
  releasePlans:
  - name: user-ns2/hello-world-app-rp-02
  - name: user-ns2/hello-world-app-rp-03


loader/loader.go
```diff
@@ -180,15 +180,33 @@ func (l *loader) GetMatchingReleasePlans(ctx context.Context, cli client.Client,
...
+
 	for i := len(releasePlans.Items) - 1; i >= 0; i-- {
 		if !slices.Contains(releasePlanAdmission.Spec.Applications, releasePlans.Items[i].Spec.Application) {
 			// Remove ReleasePlans that do not have matching applications from the list
 			releasePlans.Items = append(releasePlans.Items[:i], releasePlans.Items[i+1:]...)
+			continue
+		}
+
+		labelValue, found := releasePlans.Items[i].GetLabels()[metadata.ReleasePlanAdmissionLabel]
+		if found && labelValue != releasePlanAdmission.Name {
```
FilipNikolovski, line 207: Do we expect the releasePlans list to contain items that have different value for this label, even after filtering the list beforehand?

seanconroy2021: The filtering is only applied when ReleasePlans have labels that match to the ReleasePlanAdmission name. Otherwise, the function falls back to retrieving all ReleasePlans. The filtering part is present because during getting all, we may have ReleasePlans that have labels pointing to different ReleasePlanAdmissions. Without filtering, a ReleasePlan labelled for a specific ReleasePlanAdmission would appear in the matched for all other ReleasePlanAdmissions.

PR #779, opened 2025-07-03
fix: update the ref of git-clone-oci-ta

jinqi7:


mmalina:
OK, but I thought you wanted to change the sealights one as well?

PR #787, opened 2025-07-13
fix: schema/go.mod doesn't exist issue

jinqi7:


FilipNikolovski (approved):

johnbieren:
> > > Can we force merge it since the PR doesn't need the e2e check?
> > 
> > 
> > This is exactly why the jira is open to fix that. We also don't have override requirements enabled in this repo right now
> 
> It's different repo. The konflux-e2e-tests check are going to be removed in release-service-catalog repo. After that, the cell expression filter can be removed without affecting anything. For this repo, can we merge the PR first?

Probably because we forgot about this repo for the ticket. This issue has existed for 8 months and I don't think anyone even knew it was an issue. I don't see the need to bypass CI to merge it. The problem should be fixed here too

jinqi7:
> > > > > > Can we force merge it since the PR doesn't need the e2e check?
> > > > > 
> > > > > 
> > > > > This is exactly why the jira is open to fix that. We also don't have override requirements enabled in this repo right now
> > > > 
> > > > 
> > > > It's different repo. The konflux-e2e-tests check are going to be removed in release-service-catalog repo. After that, the cell expression filter can be removed without affecting anything. For this repo, can we merge the PR first?
> > > 
> > > 
> > > Probably because we forgot about this repo for the ticket. This issue has existed for 8 months and I don't think anyone even knew it was an issue. I don't see the need to bypass CI to merge it. The problem should be fixed here too
> > 
> > 
> > If you think it's an issue(it's expected behaviour), I can fix it now by removing the files filtering.
> 
> It's an issue that the checks sit in pending, yeah. I am fine with a build plr still running. The e2e suite shouldn't execute though
It can't work as you described currently. The e2e have to be triggered. The way which is doable in the future is to add some task in the e2e pipeline to stop the pipelinerun in the beginning. But it can not be implemented in this PR.

johnbieren:
> > > > > > > Can we force merge it since the PR doesn't need the e2e check?
> > > > > > 
> > > > > > 
> > > > > > This is exactly why the jira is open to fix that. We also don't have override requirements enabled in this repo right now
> > > > > 
> > > > > 
> > > > > It's different repo. The konflux-e2e-tests check are going to be removed in release-service-catalog repo. After that, the cell expression filter can be removed without affecting anything. For this repo, can we merge the PR first?
> > > > 
> > > > 
> > > > Probably because we forgot about this repo for the ticket. This issue has existed for 8 months and I don't think anyone even knew it was an issue. I don't see the need to bypass CI to merge it. The problem should be fixed here too
> > > 
> > > 
> > > If you think it's an issue(it's expected behaviour), I can fix it now by removing the files filtering.
> > 
> > 
> > It's an issue that the checks sit in pending, yeah. I am fine with a build plr still running. The e2e suite shouldn't execute though
> 
> It can't work as you described currently. The e2e have to be triggered. The way which is doable in the future is to add some task in the e2e pipeline to stop the pipelinerun in the beginning. But it can not be implemented in this PR.

I'm not asking you to implement it in this PR. This PR says `fix: schema/go.mod doesn't exist issue`. It should only fix that. The e2e thing can be fixed in a separate PR then this can be rebased on top of it

johnbieren:
> IMO if there **is a new change introduced to the code or related manifests**, the **build should trigger and integration pipeline too**
> 
> if there's a PR with only **code-unrelated PRs**, like this one, **neither build or integration pipeline should run**.
> 
> I think the reason why you saw these pending checks is that **they are marked as required in the github repository settings**. I don't see a reason for that. If those checks shouldn't run (like in this PR), they should not be required.

If the konflux checks aren't required, the PR will merge when set to auto-merge without waiting for them to pass. We use auto-merge *a lot*. So, they have to be required. I really don't understand why it is impossible to have the e2e tests say "does this PR touch files unrelated to the test? Yes - exit 0"
We literally already have that in the catalog repo, just not for the `.github` files

psturc:
Ah ok so it's all because of the auto-merge functionality that ignores failed checks if they are not required thus it merges the PR when it's not desired.

So there are 2 options

**1. Keep using auto-merge**
  * create a task for evaluating relevance of changed files in the PR that will make all integration pipeline tasks to be skipped
  * keep running build and integration pipeline on every PR

**2. Stop using auto-merge**
  * remove required checks
  * trigger CI (build and integration) based on the on-cel-expression

> I really don't understand why it is impossible to have the e2e tests say "does this PR touch files unrelated to the test? Yes - exit 0"

I'm not saying it's impossible, it's just an extra task to create and maintain, but feel free to create and add it to tekton-integration-catalog. It's definitely a choice of each team how they decide to handle such scenario.

jinqi7:
"auto-merge" is not enabled in this (release-service) repo. Can we remove required check and continue to use cell expression to filter the test unrelated files for this repo?

mmalina:
I would say it's not only about auto-merge - when a check is not required, the github UI will not show Merging is blocked, it will say that all required checks passed and you're free to merge the PR. So then it's easy to accidentally just merge the PR without e2e passing.

jinqi7:
> I would say it's not only about auto-merge - when a check is not required, the github UI will not show Merging is blocked, it will say that all required checks passed and you're free to merge the PR. So then it's easy to accidentally just merge the PR without e2e passing.

But the PR need at least one approval and lgtm tags before it's merged. Before we approve it, don't we need to check if e2e test passed?

mmalina:
> Before we approve it, don't we need to check if e2e test passed?

Generally no. Checking if e2e passed is not part of reviewing a PR. Ensuring e2e is passing is the responsibility of the PR author.

jinqi7:
> > Before we approve it, don't we need to check if e2e test passed?
> 
> Generally no. Checking if e2e passed is not part of reviewing a PR. Ensuring e2e is passing is the responsibility of the PR author.

It's author's responsibility to make the e2e pass. But if the PR didn't pass e2e, the reviewer seems need to add lgtm to the PR instead of an approval.

johnbieren:
I agree with Martin here. I don't think the check should just sit in pending forever, even if it is not a required check

jinqi7:
> I agree with Martin here. I don't think the check should just sit in pending forever, even if it is not a required check

The check will not appear in the check status if it's removed from required checks (for the PRs with only test unrelated files).

mmalina:
> > I agree with Martin here. I don't think the check should just sit in pending forever, even if it is not a required check
> 
> The check will not appear in the check status if it's removed from required checks (for the PRs with only test unrelated files).

The problem with this is that if it's not required, then even for PRs where we want e2e to pass, Github UI will show that it's ok to merge the PR even if e2e is still running or failed. So given this limitation, it's probably better to set it as required, start the workflow always, but just exit 0 if we see that e2e is not needed.

johnbieren:
> > > > I agree with Martin here. I don't think the check should just sit in pending forever, even if it is not a required check
> > > 
> > > 
> > > The check will not appear in the check status if it's removed from required checks (for the PRs with only test unrelated files).
> > 
> > 
> > Yes it will. You can look at the checks on this PR. Many do not have the `Required` bubble
> 
> No. It will not. See the example [konflux-ci/e2e-tests#1658](https://github.com/konflux-ci/e2e-tests/pull/1658) on-pull-request check)

I don't see an `on-pull-request` check on that PR, not sure what you are referring to. But as I said, you can see checks that aren't required right on this PR

johnbieren:
> > > > The problem with this is that if it's not required, then even for PRs where we want e2e to pass, Github UI will show that it's ok to merge the PR even if e2e is still running or failed.
> > > 
> > > 
> > > So it need the approver to take the e2e tests result into account.
> > 
> > 
> > As Martin mentioned, it isn't normal for a reviewer to pay attention to non required github checks. Can I ask why you are pushing back on this? Why can't we just exit 0 here? You already made the change on release-service-catalog to do the same
> 
> They are different solutions for the two repos. The current e2e tests in Release-service-catalog repo have their own filtering conditions. For this repo, we don't have such filtering task. And we don't need it if we remove "Red Hat Konflux / release-service-on-pull-request" and "Red Hat Konflux / konflux-e2e-tests / release-service" from the required checks.

We're going in circles. If those aren't required, we run the risk of merging PRs that legitimately break things that e2e would catch

johnbieren (approved):
Approving since the change looks good, although it won't have passing e2e until the other PR is merged

PR #790, opened 2025-07-16
fix: release service image tag in push events

jinqi7:
The image tags are currently in the "on-pr-<commit-sha>" format for
 both push and pull request events.
This PR changes the tag format for push events to distinguish them
 from those for pull request events.


johnbieren:
Maybe you can update the commit message a little more to explain this more fully because I don't understand how this change does what the commit says

johnbieren (approved):

PR #804, opened 2025-08-04
feat(RELEASE-1764): automate pr assigments

davidmogar:
This commit adds a new workflow to automate PR assigments.

### DIFF
--- a/.github/workflows/pr-assigner.yaml
+++ b/.github/workflows/pr-assigner.yaml
@@ -1,0 +1,23 @@
+---
+name: Assign Reviewers
+
+on:
+  pull_request_target:
+    types: [opened, ready_for_review, review_request_removed]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  assign-reviewers:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Assign reviewers using shared action
+        uses: konflux-ci/release-service-automations/pr-assigner@main
+        with:
+          event-type: ${{ github.event.action }}
+          pr-number: ${{ github.event.pull_request.number }}
+          removed-reviewer: ${{ github.event.requested_reviewer.login || '' }}
+          github-token: ${{ secrets.PR_ASSIGNER_PAT_TOKEN }}
+          slack-webhook: ${{ secrets.PR_ASSIGNER_SLACK_WEBHOOK }}


.github/workflows/pr-assigner.yaml
```diff
@@ -0,0 +1,44 @@
...
+        working-directory: .github/scripts/pr-assigner
+        run: node assign_reviewers.js
+        env:
+          EVENT_TYPE: ${{ github.event.action }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_REPOSITORY: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          REMOVED_REVIEWER: ${{ github.event.requested_reviewer.login }}
+          SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}
+          USER_MAP_URL: https://gist.githubusercontent.com/davidmogar/cbe426b9ca93aefe727d822e24e13e0d/raw/user_map.yaml
```
johnbieren, line 44: Is there a plan to move this file somewhere?

johnbieren (commented):
I mean I don't know javascript, so hard to review this one, but I can approve. The workflow looks good to me at least

.github/workflows/pr-assigner.yaml
davidmogar: Gist are associated to users AFAIK. Ideally yes, but not sure where. Any idea? Maybe a release repo?

johnbieren: Works for me

johnbieren:
> > Looks good, but I would suggest making this a reusable github action, so instead of having the code in each repo. We would have a repo konflux-ci/release-pr-assigner-action. This would allow us to update the code in only one place, instead of, for example, release-service and release-service-catalogue.
> > Here is one I did as an intern: https://github.com/conforma/action-validate-image/blob/main/action.yaml
> > Then, in each repo, we would only need something like this in the workflow.
> > ```
> > - name: Assign Reviewers
> >   uses: konflux-ci/release-pr-assigner-action@v1
> >   with:
> >     EVENT_TYPE: ${{ github.event.action }} ....
> > ```
> 
> I thought this was already raised and the answer was: Let's start with this and consider splitting it out as an action as the next step. I don't mind either way.

It was, just on another of these PRs, not this repo's

davidmogar:
> Some thought PR looks good to me:
> 
> ```
>             while (toAdd.length < needed && pool.length > 0) {
>                 const idx = Math.floor(Math.random() * pool.length);
>                 toAdd.push(pool.splice(idx, 1)[0]);
>             }
> ```
> 
> Currently, reviewers are assigned randomly from the eligible pool. This means the same person could be picked for multiple PRs if luck favors them, as there’s no tracking or balancing of assignments. In the future, we could consider approaches like round-robin or load-balancing to spread review duties more evenly, but these would require additional implementation and state tracking. For now, let’s observe how it works in practice.
> 
> Correct me if I am wrong @davidmogar

This is something we discussed on a call and the agreement was that "that's life". I don't think this is going to happen that often though.

PR #810, opened 2025-08-12
DO NOT MERGE: test OCI artifact viewer

Troy876:


integration-tests/pipelines/konflux-e2e-tests-pipeline.yaml
```diff
@@ -342,3 +342,5 @@ spec:
           value: cluster-provision.log
         - name: enable-test-results-analysis
           value: "true"
+        - name: enable-artifact-browser
+          value: "true"
```
psturc, line 346: I think the param should be something like `artifact-browser-url` (because the current deployment of the service won't be probably used by all teams).

and if the parameter is not specified, the functionality to post the URL to the artifact browser in the PR comment will be ignored

Troy876: Yeah I've changed the name to "artifact-browser-url" and the default value is false, so it shouldn't produce anything unless specified.

psturc: I'd suggest to use
```
        - name: artifact-browser-url
          value: $(params.artifact-browser-url)
```
here and set the default value in the Task to "".

Then check in the Task's script if the value is empty or not. So no need for checking true/false

The URL should be passed to the Task from the integration pipeline's parameters (see description in https://issues.redhat.com/browse/KFLUXDP-195). I've already set this in Konflux UI. The reason is that the URL of the artifact browser can be different for different pipelines, so it should not be hardcoded in the Task.

PR #901, opened 2025-09-28
fix(deps): add new module github.com/go-git/go-git/v5 and dependency

elenagerman:
go-git module required as a part of the [#896](https://github.com/konflux-ci/release-service/pull/896)


mmalina:
> Could this not be done in the same PR :)

I agree. Especially since it seems that the git/references.go file is there just so that the module is actually used.

elenagerman:
@seanconroy2021 @mmalina 
No doubt, it could be.
IMHO, it's even better, but I thought we were working according to some basic standards, sorry. 
I'm closing this one and moving all to the original PR. Thank you!

PR #917, opened 2025-10-01
feat: add required tagging of test clusters

psturc:
### Note: E2E tests don't have to pass to safely merge this PR
See more details in [this PR](https://github.com/konflux-ci/build-service/pull/486#issue-3465218041)

### DIFF
--- a/integration-tests/pipelines/konflux-e2e-tests-pipeline.yaml
+++ b/integration-tests/pipelines/konflux-e2e-tests-pipeline.yaml
@@ -128,7 +128,7 @@ spec:
         - name: id
           value: $(context.pipelineRun.name)
         - name: tags
-          value: env=konflux,user=release-service
+          value: 'owner=konflux-devprod@redhat.com,project=Konflux,created-by=integration-pipeline,component=release-service'
         - name: debug
           value: 'false'
         - name: ownerKind


integration-tests/pipelines/konflux-e2e-tests-pipeline.yaml
```diff
@@ -128,7 +128,9 @@ spec:
         - name: id
           value: $(context.pipelineRun.name)
         - name: tags
-          value: env=konflux,user=release-service
+          value: >
+            'owner=konflux-devprod@redhat.com,project=Konflux,created-by=integration-pipeline,
+            iac=mapt,cluster-name=$(context.pipelineRun.name),k8s-type=kind,component=release-service'
```
elenagerman, line 133: For https://github.com/konflux-ci/build-service/pull/486, I see that it also required the projectName value to be provided. At least this is what I see in the description.
It can be easily added to this list in case the answer is positive:
projectName=$(context.pipelineRun.namespace)

psturc: that's a good point!

what I didn't mention is that the `projectName` is actually a tag that is added automatically by the [mapt cluster provisioning tool](https://github.com/redhat-developer/mapt/blob/d102d1fb7281acb010ca01eea93aa38b8f8d2dd4/pkg/manager/context/context.go#L26), so we don't need to specify it in the integration pipeline.

elenagerman (approved):

PR #982, opened 2025-10-17
feat(RELEASE-1856): force daily renovate updates

davidmogar:
As per the docs, this is the recommended way to change the tekton manager schedule.


mmalina:
Changing to Draft until David is back.

PR #997, opened 2025-10-24
feat: offboard sealights from Internal Services component

flacatus:
Sealights Pilot ended with not meeting expectations. As part of onboard we need to offboard sl tasks to not meet builds and errors in the e2e.


johnbieren:
I think we need to create a PR like this for https://github.com/konflux-ci/internal-services/ too

flacatus:
https://github.com/konflux-ci/internal-services/pull/406

I think we already

PR #1034, opened 2025-11-05
fix(RELEASE-1902): fix gosec CI for merge queues

querti:
The upload of SARIF file fails when the CI runs for merge queues. The
error is "Resource not accessible by integration", and is likely caused
by insufficient permissions when triggered by a merge queue. Resolve it
by skipping this step when triggered by a merge queue. The upload
doesn't have much value in this case, since it was already uploaded on
the pull request event.

### DIFF
--- a/.github/workflows/gosec.yaml
+++ b/.github/workflows/gosec.yaml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
         with:
           fetch-depth: 1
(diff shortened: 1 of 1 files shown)


.github/workflows/gosec.yaml
```diff
@@ -39,6 +39,7 @@ jobs:
           GOROOT: ""
         continue-on-error: true
       - name: Upload SARIF file
+        if: github.event_name != 'merge_group'
```
elenagerman: you could use github.event_name == 'pull_request' instead, which is more explicit about when uploads should occur, but the current approach is equally valid. 
so I leave it to you to decide.

PR #1036, opened 2025-11-05
fix(RELEASE-1902): fix e2e tests on merge queues

querti:
The reason e2e tests fail on merge queue is due to missing image. The image is missing because the event is a "push" and thus test-metadata step expects the image tag to not have the "on-pr-" prefix. Images without this prefix are only created with the push pipeline, which means that this pipeline also has to be triggered on merge queue events.


querti:
I will close this PR for now, I don't believe this is the right direction.

PR #1085, opened 2025-11-19
feat(RELEASE-1965): keep previousRelease when retrying a release

seanconroy2021:
* Update GetPreviousRelease in loader to skip releases with the same snapshot
* Ignore failed releases and only consider successful ones as previousRelease
* Add test to ensure failed releases are not used as previousRelease
* Add test to ensure releases sharing the same snapshot are skipped as previousRelease

Ticket: RELEASE-1965
Has been tested on my local kind cluster


seanconroy2021:
doc PR can be found here https://github.com/konflux-ci/docs/pull/516

PR #1103, opened 2025-11-25
fix: add required conforma labels

happybhati:
Conforma policy checks require explicit
container labels rather than inheriting
them from base images. Added all required
labels including name, description, component,
vendor, version, and CPE to fix CI failures
and ensure Red Hat container certification compliance.

This addresses the buildah-oci-ta 0.6→0.7
migration impact where INHERIT_BASE_IMAGE_LABELS
behavior changed, even though 0.7.1 reverted
the default.

https://github.com/konflux-ci/build-definitions/blob/main/task/buildah-oci-ta/0.7/MIGRATION.md

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -46,6 +46,11 @@ LABEL io.k8s.display-name="release-service"
 LABEL summary="Konflux Release Service"
 LABEL com.redhat.component="release-service"
 LABEL io.openshift.tags="konflux"
+LABEL vendor="Red Hat, Inc."
+LABEL distribution-scope="public"
+LABEL release="1"
+LABEL url="github.com/konflux-ci/release-service"
+LABEL version="1"
 
 USER 65532:65532
 


Dockerfile
```diff
@@ -32,6 +32,12 @@ LABEL io.k8s.display-name="release-service"
 LABEL summary="Konflux Release Service"
 LABEL com.redhat.component="release-service"
 LABEL io.openshift.tags="konflux"
+LABEL vendor="Red Hat, Inc."
+LABEL distribution-scope="public"
+LABEL release="0"
+LABEL url="github.com/konflux-ci/release-service"
+LABEL version="0"
```
seanconroy2021: Could we not pass in the commit timestamp here so it isn't static and doesn't need to be manually updated?

mmalina: As I said on the other PR, I don't think we should hardcode the created label, as we just discussed in today's standup, this one should always be added by the build service.

But I am not sure about release and version - do we even care about those? I wouldn't mind using a hardcoded number, but maybe "1" would be better.

seanconroy2021: I don't think the version makes a big difference, so I'm happy with 1 or 0.

PR #1159, opened 2025-12-12
fix(KONFLUX-10973): remove owner refs from release and releaseplan

FilipNikolovski:
Remove owner references from Release and ReleasePlan to fix ArgoCD pruning. Owner references prevent ArgoCD from correctly pruning resources when using Helm.

Related issues:
- https://github.com/argoproj/argo-cd/issues/4764
- https://github.com/argoproj/argo-cd/issues/11972

Relevant Jira:
- https://issues.redhat.com/browse/KONFLUX-10973

### DIFF
--- a/controllers/releaseplan/controller.go
+++ b/controllers/releaseplan/controller.go
@@ -64,7 +64,6 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	return controller.ReconcileHandler([]controller.Operation{
 		adapter.EnsureMatchingInformationIsSet,
-		adapter.EnsureOwnerReferenceIsSet,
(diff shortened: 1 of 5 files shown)


controllers/releaseplan/controller.go
```diff
@@ -64,7 +64,6 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	return controller.ReconcileHandler([]controller.Operation{
 		adapter.EnsureMatchingInformationIsSet,
-		adapter.EnsureOwnerReferenceIsSet,
```
mmalina: I wanted to check if this should also be removed in the release controller and it's not there. Any idea why it's not there?

Issue #1172, opened 2025-12-18
Expose Snapshot reference on Release objects via label selector


jangel97:
## Summary

Release objects reference their originating Snapshot via `spec.snapshot` and their ReleasePlan via `spec.releasePlan`, but these fields are not indexed and cannot be used with Kubernetes label selectors. As a result, consumers must list all Releases in a namespace and perform client-side filtering to find Releases associated with a given Snapshot or ReleasePlan.

Downstream consumers (dashboards, automation, reporting tools) currently need to map Snapshots and ReleasePlans to Releases and independently implement client-side filtering. Exposing these relationships via labels would simplify those integrations and reduce repeated logic.

This change proposes mirroring both the Snapshot and ReleasePlan references into labels on Release objects (e.g. `appstudio.openshift.io/snapshot` and `appstudio.openshift.io/releasePlan`) to enable server-side filtering using standard Kubernetes label selectors.

## Problem Statement

Today, the only way to determine which Releases correspond to a given Snapshot or ReleasePlan is to:

1. List all `Release` objects in a namespace
2. Filter client-side by `spec.snapshot` or `spec.releasePlan`

Example (Snapshot):

```bash
oc get releases -n <namespace> -o json \
  | jq '.items[] | select(.spec.snapshot=="rhaiis-78c7f")'
```

Example (ReleasePlan):

```bash
oc get releases -n <namespace> -o json \
  | jq '.items[] | select(.spec.releasePlan=="rhaiis-ubi9-prod")'
```

This pattern:

* Requires full list queries
* Forces duplicated client-side logic across tools
* Is inconsistent with other AppStudio resources that expose key relationships via labels (e.g. commit SHA, component)

There is currently no supported way to use a label selector to query Releases by Snapshot or ReleasePlan.

## Proposed Enhancement

Mirror both the Snapshot and ReleasePlan references from the Release spec into labels on Release objects, for example:

```yaml
metadata:
  labels:
    appstudio.openshift.io/snapshot: rhaiis-78c7f
    appstudio.openshift.io/releasePlan: rhaiis-ubi9-prod
```

These labels would be automatically set by the Release Service when creating a Release and enforced as immutable, consistent with the immutability of the corresponding spec fields.

## Benefits

* Enables efficient server-side filtering:

  ```bash
  oc get releases -n <namespace> \
    -l appstudio.openshift.io/snapshot=rhaiis-78c7f
  ```

  ```bash
  oc get releases -n <namespace> \
[...]

PR #1173, opened 2025-12-18
Expose snapshot and releasePlan as labels in Release resource

jangel97:
Expose snapshot and releasePlan as labels in Release resource

Add immutable labels for spec.snapshot and spec.releasePlan to enable label-based querying of Release resources via standard Kubernetes label selectors.

Related to: https://github.com/konflux-ci/release-service/issues/1172


johnbieren:
I don't know why codecov is complaining, but we can ignore it

FilipNikolovski (approved):

PR #1244, opened 2026-01-22
feat(RELEASE-2151): add componentGroup support for rp and rpa

FilipNikolovski:
This change adds spec.componentGroup to ReleasePlan and spec.componentGroups to ReleasePlanAdmission as alternatives to the existing application fields.

The matching logic is backward compatible, allowing a ReleasePlan to match a ReleasePlanAdmission regardless of whether they use applications or componentGroups. All existing resources with application fields continue to work unchanged.

Assisted-by: Claude Code

### DIFF
--- a/api/v1alpha1/releaseplan_types.go
+++ b/api/v1alpha1/releaseplan_types.go
@@ -29,11 +29,21 @@ import (
 )
 
 // ReleasePlanSpec defines the desired state of ReleasePlan.
+// +kubebuilder:validation:XValidation:rule="has(self.application) != has(self.componentGroup)",message="exactly one of application or componentGroup must be specified"
 type ReleasePlanSpec struct {
-	// Application is a reference to the application to be released in the managed namespace
+	// Application is a reference to the application to be released in the managed namespace.
+	// Either application or componentGroup must be specified, but not both.
 	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
-	// +required
-	Application string `json:"application"`
+	// +kubebuilder:validation:MaxLength=63
+	// +optional
+	Application string `json:"application,omitempty"`
+
+	// ComponentGroup is a reference to the component group to be released in the managed namespace.
+	// Either application or componentGroup must be specified, but not both.
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +kubebuilder:validation:MaxLength=63
+	// +optional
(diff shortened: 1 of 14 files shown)


api/v1alpha1/releaseplan_types.go
```diff
@@ -29,11 +29,22 @@ import (
 )
 
 // ReleasePlanSpec defines the desired state of ReleasePlan.
+// +kubebuilder:validation:XValidation:rule="has(self.application) || has(self.componentGroup)",message="either application or componentGroup must be specified"
```
johnbieren: I'm not familiar with this syntax, but double checking - is this correct? To me, this reads "if it has application or componentGroup, tell them you have to have one or the other" (similar question about the next one)

api/v1alpha1/releaseplanadmission_types.go
```diff
@@ -102,6 +113,12 @@ type ReleasePlanAdmission struct {
 	Status ReleasePlanAdmissionStatus `json:"status,omitempty"`
 }
 
+// MatchesReleasePlan returns true if the given name is in applications or componentGroups.
```
johnbieren: Can you add a unit test or two for this?

api/v1alpha1/releaseplan_types.go
```diff
@@ -92,6 +103,7 @@ type ReleasePlanStatus struct {
 // +kubebuilder:resource:shortName=rp
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Application",type=string,JSONPath=`.spec.application`
+// +kubebuilder:printcolumn:name="ComponentGroup",type=string,JSONPath=`.spec.componentGroup`,priority=1
```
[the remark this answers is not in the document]
FilipNikolovski: This was added by claude by mistake, I've removed it. priority > 0 would only print the column in a wide output (-o wide), which we don't want.

api/v1alpha1/releaseplan_types.go
```diff
@@ -103,6 +115,14 @@ type ReleasePlan struct {
 	Status ReleasePlanStatus `json:"status,omitempty"`
 }
 
+// GetGroupName returns the application or componentGroup name.
+func (rp *ReleasePlan) GetGroupName() string {
```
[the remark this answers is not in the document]
FilipNikolovski: done!

api/v1alpha1/releaseplanadmission_types.go
FilipNikolovski: done!

johnbieren (approved):
nicely done

PR #1411, opened 2026-03-17
feat(RELEASE-2092): fall back to KubeArchive for archived Snapshots

querti:
Add a new kubearchive package that provides a client for the KubeArchive API. The client lazily discovers the API URL from a ConfigMap defined in infra-deployments [1], authenticates using the in-cluster service account token, and handles the edge case where multiple archived versions of a resource exist by falling back to the list endpoint and selecting the entry with the highest resourceVersion.

GetSnapshot in the loader now attempts to retrieve the Snapshot from KubeArchive when the cluster returns a NotFound error, enabling release processing for Snapshots that have been garbage-collected from the cluster but are still available in the archive.

[1] https://github.com/redhat-appstudio/infra-deployments/blob/main/components/kubearchive/staging/stone-stg-rh01/kustomization.yaml#L11-L16

Assisted-by: Cursor

### DIFF
--- a/kubearchive/client.go
+++ b/kubearchive/client.go
@@ -1,0 +1,264 @@
+package kubearchive
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"sync"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/rest"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	apiURLConfigMapName = "kubearchive-api-url"
+	apiURLConfigMapKey  = "URL"
+)
+
+const kaNamespace = "product-kubearchive"
+
+// Client provides access to the KubeArchive API for retrieving archived Kubernetes resources.
+// It lazily discovers the KubeArchive API URL on first use and caches it for subsequent calls.
+// Authentication and TLS are derived from the in-cluster REST configuration.
+type Client struct {
+	mu         sync.Mutex
+	apiURL     string
+	resolved   bool
+	httpClient *http.Client
+}
+
+// NewClient creates a new KubeArchive client. The API URL is discovered lazily on the first
+// Get call using the Kubernetes client to locate the KubeArchive service.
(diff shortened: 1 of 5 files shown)


kubearchive/client.go
```diff
@@ -0,0 +1,259 @@
...
+	var bestRaw json.RawMessage
+	bestRV := int64(-1)
+	for _, item := range raw.Items {
+		var meta struct {
+			Metadata struct {
+				Name            string `json:"name"`
+				ResourceVersion string `json:"resourceVersion"`
+			} `json:"metadata"`
+		}
+		if err := json.Unmarshal(item, &meta); err != nil {
```
midnightercz: why is the error ignored here?

midnightercz (dismissed):

midnightercz (dismissed):

ach912 (approved):

johnbieren (commented):
Couple small comments
If it is somewhat easy to set up kubearchive on a local kind cluster, I think it would be great to record a demo of this (start a release with the snapshot not there, show it getting it from kubearchive and still working)

kubearchive/client.go
```diff
@@ -0,0 +1,260 @@
...
+		return apierrors.NewNotFound(gvr.GroupResource(), name)
+	}
+
+	logger.Info("Selected most recent archived version from list",
+		"gvr", gvr.String(), "namespace", namespace, "name", name,
+		"resourceVersion", bestRV)
+	return json.Unmarshal(bestRaw, obj)
+}
+
+func (c *Client) resolve(ctx context.Context, cli client.Client) (string, *http.Client, error) {
```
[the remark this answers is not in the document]
querti: Added.

loader/loader_test.go
```diff
@@ -749,6 +751,53 @@ var _ = Describe("Release Adapter", Ordered, func() {
...
+			missingRelease.Spec.Snapshot = "gone-snapshot"
+
+			kaLoader := newLoader(&fakeKAClient{
+				getFunc: func(_ context.Context, _ client.Client, _ schema.GroupVersionResource, _, _ string, _ client.Object) error {
+					return fmt.Errorf("kubearchive unavailable")
+				},
+			})
+			_, err := kaLoader.GetSnapshot(ctx, k8sClient, missingRelease)
+			Expect(err).To(HaveOccurred())
+			Expect(errors.IsNotFound(err)).To(BeTrue())
```
johnbieren: Can we add like 
`Expect(err.Error()).NotTo(ContainSubstring("kubearchive unavailable"))` so we know it was the right error?

johnbieren (commented):
one last comment then lgtm

johnbieren (approved):

davidmogar (approved):

midnightercz (approved):

PR #1427, opened 2026-03-20
feat(RELEASE-2357): add previousRelease override for collectors

scoheb:
- add collectorDataOverrides.previousRelease field to Release spec
- allows users to specify which Release to use for collector
  operations instead of automatic detection
- the Release is looked up in the same namespace as the current
  Release and must exist and not be failed
- release fails with error if specified previousRelease is not
  found or has failed

Assisted-by: Cursor


happybhati (commented):
Nice feature — this directly addresses the issue we saw with ACS where all prior releases failed and GetPreviousRelease returned nil, making the cve-collector only inspect single commits.

happybhati (approved):

ach912 (approved):

davidmogar (approved):

davidmogar:
Moving to draft and set to not ready as we are waiting for changes.

scoheb:
not currently needed.

PR #1468, opened 2026-04-07
feat(KONFLUX-12768): add MaxConcurrentReconciles support to all controllers

FilipNikolovski:
Add configurable MaxConcurrentReconciles to all controllers via the MAX_CONCURRENT_RECONCILES environment variable, defaulting to 1 if unset. The env var is set to 10 in the manager deployment manifest.

Assisted-by: Claude Code

### DIFF
--- a/controllers/release/controller.go
+++ b/controllers/release/controller.go
@@ -35,13 +35,15 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	sigsctrl "sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 // Controller reconciles a Release object
 type Controller struct {
-	client client.Client
-	log    logr.Logger
+	client                  client.Client
+	log                     logr.Logger
+	MaxConcurrentReconciles int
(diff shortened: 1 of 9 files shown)


elenagerman (approved):

controllers/release/controller.go
```diff
@@ -35,13 +35,15 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 // Controller reconciles a Release object
 type Controller struct {
-	client client.Client
-	log    logr.Logger
+	client                  client.Client
+	log                     logr.Logger
+	MaxConcurrentReconciles int
```
davidmogar, line 46: Why does it need to be exposed? Why not `maxConc...`?

FilipNikolovski: Basically to expose the configuration when we instantiate the Controller, but I can move the setting of MaxConcurrentReconciles in the Register method, I don't have a preference.

FilipNikolovski: I think we need to expose it actually, or alternatively have like a `NewReleaseController` func that accepts this as an argument, otherwise we would need to duplicate the code for fetching the value from the env var in each Register method, wdyt @davidmogar?

theflockers (approved):
lgtm once David's comment is closed.

davidmogar (approved):

johnbieren (approved):

ach912 (approved):

PR #1514, opened 2026-04-21
feat(RELEASE-2119): add func to find current memory & timeout values

seanconroy2021:
Add lookup functions for the memory limits and timeout values currently in use by a failed step or task.

* Add GetStepComputeResources to look up the memory limit for a step, checking PR overrides (StepSpecs then ComputeResources) before falling back to the TR task spec.
* Add GetPipelineRunTimeouts to return the 3 timeout fields.
* Add GetTaskRunTimeout to find the per task timeout by checking TaskRunSpecs overrides, the TR spec, then the global tasks timeout

Jira: https://redhat.atlassian.net/browse/RELEASE-2119
Assisted-by: Claude

### DIFF
--- a/tekton/utils.go
+++ b/tekton/utils.go
@@ -21,6 +21,8 @@ import (
 
 	"github.com/konflux-ci/release-service/metadata"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"knative.dev/pkg/apis"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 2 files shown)


tekton/utils.go
```diff
@@ -84,3 +86,164 @@ func IsPipelineRunDone(pipelineRun *tektonv1.PipelineRun) bool {
...
+func GetStepComputeResources(pipelineRun *tektonv1.PipelineRun, taskRun *tektonv1.TaskRun, taskName, stepName string) *corev1.ResourceRequirements {
+	if pipelineRun != nil {
+		for _, taskRunSpec := range pipelineRun.Spec.TaskRunSpecs {
+			if taskRunSpec.PipelineTaskName != taskName {
+				continue
+			}
+
+			for _, stepSpec := range taskRunSpec.StepSpecs {
+				if stepSpec.Name == stepName {
+					resources := stepSpec.ComputeResources
```
querti, line 194: Is this variable necessary? If yes, should it also be added below in taksRunSpec?

querti (approved):
LGTM other than a minor comment. Also, the PR title and description only seem to describe the second commit, which is kind of confusing when reviewing.

seanconroy2021:
> LGTM other than a minor comment. Also, the PR title and description only seem to describe the second commit, which is kind of confusing when reviewing.

Yeah, sorry, should have added a comment. a278e4e4b650ec94211e28ad40dd7e1676cda938 PR is waiting to be merged, so just cherry-picked to avoid merge conflicts. https://github.com/konflux-ci/release-service/pull/1455

tekton/utils.go
seanconroy2021: Yes the local copy is needed here. Take the address since ComputeResources is a value type, not a pointer. Could also do &stepSpec.ComputeResources directly doesn't seem as clear. 

Not needed down in taskRunSpec.ComputeResources since it is already a pointer, so we return it directly.

querti (dismissed):

querti (dismissed):

tekton/utils.go
```diff
@@ -176,3 +178,72 @@ func getFailedTaskRun(pipelineRun *tektonv1.PipelineRun, taskRuns []tektonv1.Tas
...
+// to the TaskRun's task spec. Returns nil if none are configured.
+func GetStepComputeResources(pipelineRun *tektonv1.PipelineRun, taskRun *tektonv1.TaskRun, taskName, stepName string) *corev1.ResourceRequirements {
+	if pipelineRun != nil {
+		for _, taskRunSpec := range pipelineRun.Spec.TaskRunSpecs {
+			if taskRunSpec.PipelineTaskName != taskName {
+				continue
+			}
+
+			for _, stepSpec := range taskRunSpec.StepSpecs {
+				if stepSpec.Name == stepName {
```
johnbieren, line 195: When would we not find the step other than the step wasn't in the plr? Which I would think would return an error. I am confused by lines 200 and 209

tekton/utils.go
```diff
@@ -176,3 +178,72 @@ func getFailedTaskRun(pipelineRun *tektonv1.PipelineRun, taskRuns []tektonv1.Tas
...
+func GetTaskRunTimeout(pipelineRun *tektonv1.PipelineRun, taskRun *tektonv1.TaskRun, taskName string) *metav1.Duration {
+	if pipelineRun != nil {
+		for _, taskRunSpec := range pipelineRun.Spec.TaskRunSpecs {
+			if taskRunSpec.PipelineTaskName == taskName && taskRunSpec.Timeout != nil {
+				return taskRunSpec.Timeout
+			}
+		}
+	}
+
+	if taskRun != nil && taskRun.Spec.Timeout != nil {
```
johnbieren, line 246: how would this be true and line 234 be false?

seanconroy2021: taskRunSpecs are optional overrides most tasks won't have one so the loop at 234 finds nothing. The TR still has a timeout from the pipeline spec task level timeout when creating the TR so we find that instead..

tekton/utils.go
seanconroy2021: > When would we not find the step other than the step wasn't in the plr?

The step would always exist in the task spec, but it might not have an override in StepSpecs.

Line 200 checks ComputeResources on the TaskRunSpec is a resource override for all steps in the task separate from the StepSpecs which only override the single step. 

Line 209 is the fallback to the step definition in the TR task spec, which is where the task definition sets the resources. I was taking this from looking at the PR/TR.

tekton/utils.go
davidmogar: This code is quite confusing to me. At a very minimum, it needs comments.

seanconroy2021: Will add some :)

seanconroy2021: Ok added

johnbieren (approved):

davidmogar (approved):

ach912 (approved):

PR #1541, opened 2026-04-30
feat(RELEASE-2122): add retry info to ReleasePlanAdmission status

mmalina:
Update ReleasePlanAdmission status with retry configuration info
to help users understand whether their pipelines will be
automatically retried on failure.

* Add RetryInfo struct to ReleasePlanAdmissionStatus with Enabled,
  MaxRetries, and Reason fields
* Add EnsureRetryInformationIsSet operation to RPA controller to
  update retry info based on ReleaseServiceConfig
* RPA controller watches ReleaseServiceConfig from operator
  namespace (SERVICE_NAMESPACE) and enqueues all RPAs for
  reconciliation when it changes
* Create retry matcher utility package with pipeline matching
  logic using auto-anchored regex patterns for URL/revision and
  exact match for pathInRepo
* Add tag extraction from four locations in mapping:
  defaults.tags, components[].tags, components[].componentTags,
  and components[].repositories[].tags
* Support RPA.Spec.Pipeline.MaxRetries override: when set to 0,
  retries are disabled; when >0, retries enabled with that value,
  bypassing RSC matching and tag checks
* Update predicates to trigger on pipeline and data field changes
* Add comprehensive test suite with 26 tests covering all retry
  matching scenarios

When ReleaseServiceConfig changes, all RPAs are enqueued and each
updates its own status. When an individual RPA changes (pipeline
or data), only that RPA updates its status.

Retries are disabled if any tag from RPA.Spec.Data or matched
ReleasePlan.Spec.Data matches the RetryPolicy.DisableOn.Tags list.


swickersh (commented):
The PR describes a deliberate "dual reconciliation approach" — Flow A (RSC changes → all RPAs) handled by the new `ReleaseServiceConfig` controller, and Flow B (individual RPA changes → that RPA) handled by the RPA controller. That's a reasonable design, but the implementation ends up covering Flow A twice:
1. The `ReleaseServiceConfig` controller (`EnsureAllReleasePlanAdmissionsHaveRetryInfo`) loops over all RPAs and patches them directly.
2. The `enqueueAllRPAs` watch added to the RPA controller also enqueues all RPAs when RSC changes, each of which then runs `EnsureRetryInformationIsSet`.

Both are triggered by the same `ReleaseServiceConfigPredicate`, so every qualifying RSC change results in each RPA being patched twice. The patches are idempotent so this won't produce incorrect state, but it doubles the API load for no benefit and makes the reconciliation logic harder to follow.

To match the intended design, I'd suggest picking one path for Flow A and removing the other. The `enqueueAllRPAs` watch in the RPA controller already covers it completely, so the dedicated `ReleaseServiceConfig` controller could be dropped — keeping all retry info logic in a single reconcile path.

 \-on behalf of Scott via Cursor

mmalina:
> The PR describes a deliberate "dual reconciliation approach" — Flow A (RSC changes → all RPAs) handled by the new `ReleaseServiceConfig` controller, and Flow B (individual RPA changes → that RPA) handled by the RPA controller. That's a reasonable design, but the implementation ends up covering Flow A twice:
> 
> 1. The `ReleaseServiceConfig` controller (`EnsureAllReleasePlanAdmissionsHaveRetryInfo`) loops over all RPAs and patches them directly.
> 2. The `enqueueAllRPAs` watch added to the RPA controller also enqueues all RPAs when RSC changes, each of which then runs `EnsureRetryInformationIsSet`.
> 
> Both are triggered by the same `ReleaseServiceConfigPredicate`, so every qualifying RSC change results in each RPA being patched twice. The patches are idempotent so this won't produce incorrect state, but it doubles the API load for no benefit and makes the reconciliation logic harder to follow.
> 
> To match the intended design, I'd suggest picking one path for Flow A and removing the other. The `enqueueAllRPAs` watch in the RPA controller already covers it completely, so the dedicated `ReleaseServiceConfig` controller could be dropped — keeping all retry info logic in a single reconcile path.
> 
> -on behalf of Scott via Cursor

Thanks! It seems you're right and we don't really need a dedicated ReleaseServiceConfig controller at all - the RPA controller already watches for RP changes, so the same pattern can be used here. Updated.

swickersh:
`EnsureRetryInformationIsSet` in `adapter.go` has no unit tests (0% coverage per codecov). Given the rest of the adapter has solid test coverage, it would be good to add tests in `adapter_test.go` covering at least:

- RSC not found → `RetryInfo` set to disabled
- Pipeline matches a retryable pipeline → `RetryInfo` set to enabled with correct `MaxRetries`
- Tags match a disable condition → `RetryInfo` set to disabled
- `RetryInfo` unchanged → no patch issued (the `reflect.DeepEqual` short-circuit)
\-on behalf of Scott via Cursor

swickersh:
In `MatchPipeline`, both `Url` and `Revision` are compiled as regexes and matched with `MatchString`, which does substring matching. This means a pattern like `https://github.com/org/repo` would also match `https://github.com/org/repo-extra`, which is almost certainly not what operators intend when configuring retryable pipelines.

Either anchor the patterns by default (e.g. wrap with `^(?:...)$` before compiling), or clearly document in the `RetryablePipeline` type and user-facing docs that patterns must be explicitly anchored (e.g. `^https://github.com/org/repo$`) to get exact-match behaviour.

\-on behalf of Scott via Cursor

mmalina:
> `EnsureRetryInformationIsSet` in `adapter.go` has no unit tests (0% coverage per codecov). Given the rest of the adapter has solid test coverage, it would be good to add tests in `adapter_test.go` covering at least:
> 
> * RSC not found → `RetryInfo` set to disabled
> * Pipeline matches a retryable pipeline → `RetryInfo` set to enabled with correct `MaxRetries`
> * Tags match a disable condition → `RetryInfo` set to disabled
> * `RetryInfo` unchanged → no patch issued (the `reflect.DeepEqual` short-circuit)
>   -on behalf of Scott via Cursor

Good point. Tests added.

mmalina:
> In `MatchPipeline`, both `Url` and `Revision` are compiled as regexes and matched with `MatchString`, which does substring matching. This means a pattern like `https://github.com/org/repo` would also match `https://github.com/org/repo-extra`, which is almost certainly not what operators intend when configuring retryable pipelines.
> 
> Either anchor the patterns by default (e.g. wrap with `^(?:...)$` before compiling), or clearly document in the `RetryablePipeline` type and user-facing docs that patterns must be explicitly anchored (e.g. `^https://github.com/org/repo$`) to get exact-match behaviour.
> 
> -on behalf of Scott via Cursor

That's a good point. I am going with always adding `^` and `$`. Changed. The reason Claude did it this way is because IsPipelineOverridden has the same pattern (so possibly a bug as well - it's used to determine which pipelines are using trusted artifacts).

config/crd/bases/appstudio.redhat.com_releaseplanadmissions.yaml
```diff
@@ -1110,6 +1110,26 @@ spec:
...
+                    description: Enabled indicates whether retries are enabled for
+                      this RPA's pipeline
+                    type: boolean
+                  maxRetries:
+                    description: MaxRetries is the maximum number of retries if enabled
+                    type: integer
+                  reason:
+                    description: |-
+                      Reason explains why retries are enabled or disabled
+                      Examples: "disabled due to tag: production", "enabled with policy", "no matching pipeline"
```
swickersh, line 1239: Not a blocker, but a nice enhancement later might be including _**where**_ the disabling tag is from.
Like "disabled by tag 'production' in ReleasePlan `my-release-plan` (mapping.defaults.tags)"

You can resolve this comment without action

swickersh (approved):

config/crd/bases/appstudio.redhat.com_releaseplanadmissions.yaml
mmalina: Do you think it's worth creating a jira for this? I can do it.

mmalina:
e2e failed to provision kind cluster.
/retest

config/crd/bases/appstudio.redhat.com_releaseplanadmissions.yaml
swickersh: Only if you think it's worthwhile.  Seems like a nice-to-have to me.

swickersh (dismissed):

controllers/releaseplanadmission/adapter_test.go
```diff
@@ -201,6 +201,251 @@ var _ = Describe("ReleasePlanAdmission adapter", Ordered, func() {
...
+			Expect(!result.RequeueRequest && !result.CancelRequest).To(BeTrue())
+			Expect(err).NotTo(HaveOccurred())
+			Expect(adapter.releasePlanAdmission.Status.RetryInfo).ToNot(BeNil())
+			Expect(adapter.releasePlanAdmission.Status.RetryInfo.Enabled).To(BeTrue())
+			Expect(adapter.releasePlanAdmission.Status.RetryInfo.MaxRetries).ToNot(BeNil())
+			Expect(*adapter.releasePlanAdmission.Status.RetryInfo.MaxRetries).To(Equal(maxRetries))
+			Expect(adapter.releasePlanAdmission.Status.RetryInfo.Reason).To(Equal("retries enabled by policy"))
+		})
+
+		It("should set RetryInfo to disabled when tags match disable condition", func() {
```
seanconroy2021, line 337: Sorry for the slow review was trying to get this work with my own :( 

The test says "should set RetryInfo to disabled when tags match disable condition" but the matchedRP has no Spec.Data. So the disable-by-tags is never actually tested.

controllers/releaseplanadmission/adapter.go
```diff
@@ -74,3 +76,40 @@ func (a *adapter) EnsureMatchingInformationIsSet() (controller.OperationResult,
...
+// EnsureRetryInformationIsSet is an operation that will ensure that the ReleasePlanAdmission has updated
+// retry information in its status based on the current ReleaseServiceConfig.
+func (a *adapter) EnsureRetryInformationIsSet() (controller.OperationResult, error) {
+	// Get ReleaseServiceConfig
+	rsc, err := a.loader.GetReleaseServiceConfig(
+		a.ctx, a.client,
+		v1alpha1.ReleaseServiceConfigResourceName,
+		a.releasePlanAdmission.Namespace,
+	)
+	if err != nil && !errors.IsNotFound(err) {
```
seanconroy2021: If the RSC isn't found GetReleaseServiceConfig still returns a non nil empty. So DetermineRetryInfo never hits the rsc equal nil it seen an empty. RetryablePipelines would return "pipeline not configured for retries" instead of "no retry configuration available." If I am looking at this right?

Guess the fix would if errors.IsNotFound(err) set rsc equal nil else controller.RequeueWithError(err)

controllers/utils/retry/matcher.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/go-logr/logr"
+	"github.com/konflux-ci/release-service/api/v1alpha1"
+	tektonutils "github.com/konflux-ci/release-service/tekton/utils"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// DetermineRetryInfo determines the retry information for a given ReleasePlanAdmission.
+// It checks if the RPA's pipeline matches any configured retryable pipelines and if
+// tags from the RPA or matched ReleasePlans disable retries.
+func DetermineRetryInfo(
```
seanconroy2021: We only use RetryPolicy.MaxRetries from the RSC but never looks at RPA.Spec.Pipeline.MaxRetries. The idea was that users can override the default or set it to 0 to disable. https://github.com/konflux-ci/release-service/blob/38a346f0db0bbb3509a2a47351f1a33dad3efe1a/tekton/utils/pipeline.go#L80

controllers/releaseplanadmission/adapter_test.go
mmalina: No worries. Sorry for the late reaction :)

Good catch, thanks. Should be fixed now.

controllers/releaseplanadmission/adapter.go
mmalina: Thanks, should be fixed now.

controllers/utils/retry/matcher.go
mmalina: Good point. This was definitely missed. Added now.

config/crd/bases/appstudio.redhat.com_releaseplanadmissions.yaml
mmalina: I started writing the jira and then realized that I don't think it's worth the effort, sorry :) Because in reality, it will be {{ release_timestamp }} or {{ incrementer }} and it should be easy to find it. Feel free to create the jira yourself though if you feel strongly about it.

controllers/releaseplanadmission/adapter.go
```diff
@@ -74,3 +76,43 @@ func (a *adapter) EnsureMatchingInformationIsSet() (controller.OperationResult,
...
 }
+
+// EnsureRetryInformationIsSet is an operation that will ensure that the ReleasePlanAdmission has updated
+// retry information in its status based on the current ReleaseServiceConfig.
+func (a *adapter) EnsureRetryInformationIsSet() (controller.OperationResult, error) {
+	// Get ReleaseServiceConfig
+	rsc, err := a.loader.GetReleaseServiceConfig(
+		a.ctx, a.client,
+		v1alpha1.ReleaseServiceConfigResourceName,
+		a.releasePlanAdmission.Namespace,
```
seanconroy2021: The RSC CR live in the operators NS, so you should use `os.Getenv("SERVICE_NAMESPACE")`

mmalina: ok, added with a fallback to release-service

seanconroy2021 (dismissed):
LGTM - only need to squash the commits :)

mmalina:
> LGTM - only need to squash the commits :)

thanks for the review! commits squashed.

api/v1alpha1/releaseplanadmission_types.go
```diff
@@ -86,6 +86,20 @@ type MatchedReleasePlan struct {
 	Active bool `json:"active,omitempty"`
 }
 
+// RetryInfo contains information about retry configuration for a ReleasePlanAdmission.
+type RetryInfo struct {
```
johnbieren, line 90: Can we copy the Mitigations here too so the user doesn't have to look at the RSC?

controllers/releaseplanadmission/adapter.go
```diff
@@ -74,3 +77,48 @@ func (a *adapter) EnsureMatchingInformationIsSet() (controller.OperationResult,
...
+	if err != nil {
+		if errors.IsNotFound(err) {
+			rsc = nil
+		} else {
+			return controller.RequeueWithError(err)
+		}
+	}
+
+	// Get matched ReleasePlans
+	matchedRPs, err := a.loader.GetMatchingReleasePlans(a.ctx, a.client, a.releasePlanAdmission)
```
johnbieren: isn't it wasteful to load these if rsc is nil?

johnbieren (commented):
looks mostly good, nice work
I think it would be great to demo this to 1) show off the functionality and 2) give us confidence before putting it in prod

controllers/utils/retry/matcher.go
```diff
@@ -0,0 +1,259 @@
...
+	"github.com/konflux-ci/release-service/api/v1alpha1"
+	tektonutils "github.com/konflux-ci/release-service/tekton/utils"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// DetermineRetryInfo determines the retry information for a given ReleasePlanAdmission.
+// It checks if the RPA's pipeline matches any configured retryable pipelines and if
+// tags from the RPA or matched ReleasePlans disable retries.
+// The RPA's Pipeline.MaxRetries can override the RSC configuration.
+func DetermineRetryInfo(
```
davidmogar: Function parameters across the base code are defined in line (not one per line). I would keep that format to be consistent. Applies to multiple places.

controllers/utils/retry/matcher.go
```diff
@@ -0,0 +1,259 @@
...
+		Enabled:    true,
+		MaxRetries: &maxRetries,
+		Reason:     "retries enabled by policy",
+	}
+}
+
+// MatchPipeline checks if a pipeline matches any of the configured retryable pipelines.
+// It returns the first matching RetryablePipeline or nil if no match is found.
+// Follows the same pattern as IsPipelineOverridden from releaseserviceconfig_types.go.
+func MatchPipeline(
```
davidmogar: Nitpick. Given that this returns something, I would call it `GetRetryablePipeline` or something similar. Easy to understand.

api/v1alpha1/releaseplanadmission_types.go
mmalina: good point, added

controllers/releaseplanadmission/adapter.go
mmalina: good catch, fixed

controllers/utils/retry/matcher.go
mmalina: ok, fixed

controllers/utils/retry/matcher.go
mmalina: agreed, fixed

api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -760,6 +765,26 @@ func (in *ReleaseStatus) DeepCopy() *ReleaseStatus {
 	return out
 }
```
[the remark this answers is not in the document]
mmalina: This is really a redundant comment, as this also broke CI. Fixed now.

controllers/utils/predicates/predicates.go
```diff
@@ -46,7 +46,9 @@ func MatchPredicate() predicate.Predicate {
 			return haveApplicationsChanged(e.ObjectOld, e.ObjectNew) ||
 				hasBehaviorLabelChanged(e.ObjectOld, e.ObjectNew) ||
 				hasMatchConditionChanged(e.ObjectOld, e.ObjectNew) ||
```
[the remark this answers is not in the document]
mmalina: Good point.

  1. Created ReleasePlanAdmissionPredicate() - A new predicate that includes all
  MatchPredicate behavior PLUS pipeline and data changes
  2. Kept MatchPredicate() unchanged - Now only watches applications, behavior labels,
  match conditions, and source (no pipeline/data)
  3. Updated RPA controller - Uses ReleasePlanAdmissionPredicate() for the main resource
  4. ReleasePlan controller unaffected - Still uses MatchPredicate() and won't reconcile
  on pipeline/data changes it doesn't need

  This avoids the unnecessary reconciliation traffic on the ReleasePlan controller.

controllers/releaseplanadmission/adapter.go
```diff
@@ -74,3 +77,51 @@ func (a *adapter) EnsureMatchingInformationIsSet() (controller.OperationResult,
 
 	return controller.RequeueOnErrorOrContinue(a.client.Status().Patch(a.ctx, a.releasePlanAdmission, patch))
 }
+
+// EnsureRetryInformationIsSet is an operation that will ensure that the ReleasePlanAdmission has updated
+// retry information in its status based on the current ReleaseServiceConfig.
```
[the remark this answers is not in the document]
mmalina: Done! Now the RPA controller is consistent with the Release controller - if
  SERVICE_NAMESPACE is not set, it returns an error and requeues instead of silently
  falling back to "release-service". This ensures we don't accidentally read the RSC
  from the wrong namespace.

controllers/utils/retry/matcher.go
```diff
@@ -0,0 +1,251 @@
...
+		return &v1alpha1.RetryInfo{
+			Enabled:    true,
+			MaxRetries: rpa.Spec.Pipeline.MaxRetries,
+			Reason:     "retries enabled by RPA pipeline override",
+		}
+	}
+
+	// Try to match the pipeline against RSC retryable pipelines
+	matchedRetryable := GetMatchingRetryablePipeline(rpa.Spec.Pipeline, rsc.Spec.RetryablePipelines, logger)
+	if matchedRetryable == nil {
```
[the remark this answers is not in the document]
mmalina: Fixed! Now when the ReleaseServiceConfig is deleted (even though it's a singleton and
  shouldn't be), all RPAs will reconcile and update their RetryInfo status to reflect
  that there's no retry configuration available, preventing stale status.

controllers/utils/predicates/predicates.go
```diff
@@ -135,3 +137,67 @@ func hasSourceChanged(objectOld, objectNew client.Object) bool {
...
+
+// hasPipelineChanged returns true if the objects are ReleasePlanAdmissions or ReleasePlans
+// and their pipeline specifications have changed.
+func hasPipelineChanged(objectOld, objectNew client.Object) bool {
+	if releasePlanAdmissionOld, ok := objectOld.(*v1alpha1.ReleasePlanAdmission); ok {
+		if releasePlanAdmissionNew, ok := objectNew.(*v1alpha1.ReleasePlanAdmission); ok {
+			return !reflect.DeepEqual(releasePlanAdmissionOld.Spec.Pipeline, releasePlanAdmissionNew.Spec.Pipeline)
+		}
+	}
+
```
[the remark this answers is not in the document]
mmalina: Added comprehensive test coverage for hasPipelineChanged and hasDataChanged
  functions. The test suite now has 43 tests (up from 35), all passing.

controllers/releaseplanadmission/controller.go
```diff
@@ -65,18 +67,22 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 
 // Register registers the controller with the passed manager and log.
 func (c *Controller) Register(mgr ctrl.Manager, log *logr.Logger, _ cluster.Cluster) error {
 	c.client = mgr.GetClient()
+	c.log = *log
 
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(sigsctrl.Options{MaxConcurrentReconciles: c.MaxConcurrentReconciles}).
-		For(&v1alpha1.ReleasePlanAdmission{}, builder.WithPredicates(predicates.MatchPredicate())).
+		For(&v1alpha1.ReleasePlanAdmission{}, builder.WithPredicates(predicates.ReleasePlanAdmissionPredicate())).
```
johnbieren: Isn't it cleaner to have two predicates here instead of having the ReleasePlanAdmissionPredicate duplicate everything in MatchPredicate? I also think ReleasePlanAdmissionPredicate could use a better name. https://github.com/konflux-ci/release-service/blob/main/controllers/release/controller.go#L113 is an example of multiple predicates

mmalina: ok, good point, please check now

controllers/releaseplanadmission/controller.go
```diff
@@ -65,18 +68,22 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 
 // Register registers the controller with the passed manager and log.
 func (c *Controller) Register(mgr ctrl.Manager, log *logr.Logger, _ cluster.Cluster) error {
 	c.client = mgr.GetClient()
+	c.log = *log
 
 	return ctrl.NewControllerManagedBy(mgr).
 		WithOptions(sigsctrl.Options{MaxConcurrentReconciles: c.MaxConcurrentReconciles}).
-		For(&v1alpha1.ReleasePlanAdmission{}, builder.WithPredicates(predicates.MatchPredicate())).
+		For(&v1alpha1.ReleasePlanAdmission{}, builder.WithPredicates(predicate.Or(predicates.MatchPredicate(), predicates.RetryInfoPredicate()))).
```
johnbieren, line 82: what does the Or do? Isn't just listing the two of them enough? Like https://github.com/konflux-ci/release-service/blob/main/controllers/release/controller.go#L113

mmalina: It seems it's justified here, because otherwise both would need to be true. From claude:
```
You raise a good point, but there's an important difference! When you list multiple
  predicates with WithPredicates(p1, p2), they are AND'd together - ALL must return true for
  reconciliation to happen.

  But I need OR behavior here - reconcile if EITHER:
  - Match-related fields change (applications, labels, target, etc.) OR
  - Retry-related fields change (pipeline, data)

  Looking at the Release controller example you linked:
  For(&v1alpha1.Release{}, builder.WithPredicates(predicate.GenerationChangedPredicate{},
  predicates.ReleasePredicate()))

  This means reconcile only when BOTH:
  - Generation changed AND
  - Some Release-specific field changed

  That's AND logic, which makes sense for Release - they want to filter out status-only
  updates.

  But for RPA, I want:
  - Reconcile if match fields change (application, target, etc.)
  - OR reconcile if retry fields change (pipeline, data)

  So predicate.Or() is correct here. Without it, if I just listed both predicates, I'd only
  reconcile when BOTH match fields AND retry fields change, which would miss most updates!

  Does that make sense?
```

johnbieren: yeah i didn't realize it was and for a list instead of or

johnbieren (dismissed):
I think this should be demoed

ach912 (dismissed):
lgtm

controllers/utils/predicates/predicates.go
```diff
@@ -135,3 +137,67 @@ func hasSourceChanged(objectOld, objectNew client.Object) bool {
...
+	return false
+}
+
+// ReleaseServiceConfigPredicate returns a predicate which returns true when a ReleaseServiceConfig
+// is created or when its RetryablePipelines specification changes.
+func ReleaseServiceConfigPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			return true
+		},
```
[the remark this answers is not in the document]
mmalina: This is fixed now - DeleteFunc returns true as well.

mmalina:
## Test Coverage Improvements

Added comprehensive tests for the new controller and predicate code to achieve 100% coverage:

### 1. Controller Tests (`controllers/releaseplanadmission/controller_test.go`)

Added 3 test cases for the `enqueueAllRPAs` function:
- ✅ **Should enqueue all ReleasePlanAdmissions** - Verifies that when RSC changes, all RPAs are correctly enqueued for reconciliation
- ✅ **Should return empty slice when List fails** - Tests error handling when the K8s client fails to list RPAs
- ✅ **Should return empty slice when no RPAs exist** - Tests behavior when there are no RPAs in the cluster

Also added a mock `errorListClient` helper to simulate client errors.

**Result:** `enqueueAllRPAs` coverage: **60% → 100%** ✅

### 2. Predicate Tests (`controllers/utils/predicates/predicates_test.go`)

Added 8 test cases for the `ReleaseServiceConfigPredicate` function:
- ✅ Returns true for create events
- ✅ Returns true for delete events  
- ✅ Returns false for generic events
- ✅ Returns true when RetryablePipelines changes
- ✅ Returns false when RetryablePipelines doesn't change
- ✅ Returns false when both have empty RetryablePipelines
- ✅ Returns false when object is not ReleaseServiceConfig (type assertion edge case)
- ✅ Returns false when only ObjectOld is not ReleaseServiceConfig
- ✅ Returns false when only ObjectNew is not ReleaseServiceConfig

**Result:** `ReleaseServiceConfigPredicate` coverage: **75.51% → 100%** ✅

### Overall Coverage
```
ok  	github.com/konflux-ci/release-service/controllers/releaseplanadmission	7.897s	coverage: 83.0%
ok  	github.com/konflux-ci/release-service/controllers/utils/predicates	(cached)	coverage: 97.0%
```

All new code now has complete test coverage with no code changes required - only test additions.

mmalina:
Please see the second commit adding new tests - nothing changed in the code.

ach912 (approved):
lgtm

davidmogar (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-06-11T11:29:23.248Z |

Approved

seanconroy2021 (approved):
lgtm

mmalina:
e2e failed:
```
HEAD is now at 1ae5653 test(RELEASE-2122): add tests for controller and predicates
go: downloading go1.26.0 (linux/amd64)
Failed to compile cmd:
 
k8s.io/api/core/v1: /tmp/tmp.6Wivwdjulh/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.26.0.linux-amd64/pkg/tool/linux_amd64/compile: signal: killed
 
 
Ginkgo ran 1 suite in 28.881398482s
 
Test Suite Failed
```
So the compilation was killed - could be oom.

johnbieren:
@mmalina what is your planned schedule for putting this in staging/production of infra-deployments? I don't see any PRs open. I have another commit I need to get in after it

mmalina:
> @mmalina what is your planned schedule for putting this in staging/production of infra-deployments? I don't see any PRs open. I have another commit I need to get in after it

We need to merge this first: https://github.com/redhat-appstudio/infra-deployments/pull/12359
But it's currently failing on a CI issue - seems random. And I pinged devprod for help and so far no help. I pinged infra as well now. https://redhat-internal.slack.com/archives/C07PZ0U12MA/p1781617887451829?thread_ts=1781532465.945459&cid=C07PZ0U12MA

johnbieren:
> > @mmalina what is your planned schedule for putting this in staging/production of infra-deployments? I don't see any PRs open. I have another commit I need to get in after it
> 
> We need to merge this first: [redhat-appstudio/infra-deployments#12359](https://github.com/redhat-appstudio/infra-deployments/pull/12359) But it's currently failing on a CI issue - seems random. And I pinged devprod for help and so far no help. I pinged infra as well now. https://redhat-internal.slack.com/archives/C07PZ0U12MA/p1781617887451829?thread_ts=1781532465.945459&cid=C07PZ0U12MA

ack, thanks! I'll start watching that PR

PR #1544, opened 2026-04-30
feat: add distributed tracing for release PipelineRuns

ci-operator:
Emit waitDuration and executeDuration spans per release PipelineRun, parented under
the tekton.dev/pipelinerunSpanContext annotation that the Release CR carries forward.
When a Release CR is rejected at the validation gate (no PipelineRun ever runs),
emit a waitDuration span so the rejection surfaces in the trace instead of
disappearing. Span attributes follow the delivery tracing schema established in
[konflux-ci/architecture ADR 0062](https://github.com/konflux-ci/architecture/blob/main/ADR/0062-distributed-tracing.md);
sampling and label names are env-var configurable.

Validated end-to-end on a CRC dev cluster: build PR → Snapshot → Release → release
PR composes under one trace ID, with all four services emitting against the ADR-0062
schema.

Assisted-by: Claude Code

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,6 +21,7 @@ syncer/                # Cross-namespace Snapshot syncing
 tekton/                # PipelineRun builders, status helpers, watch predicates
 metadata/              # Label/annotation/finalizer constants and utilities
 metrics/               # Prometheus gauges, histograms, counters for release lifecycle
+tracing/               # OpenTelemetry distributed tracing for PipelineRuns
 config/                # Kustomize manifests (CRDs, RBAC, webhooks, samples)
 main.go                # Entry point — registers controllers and webhooks
 ```
--- a/README.md
+++ b/README.md
@@ -51,3 +51,53 @@ by default, this operator exports the following custom metrics:
 | release_pre_processing_duration_seconds          | Histogram | How long in seconds a Release takes to start processing             |
 | release_validation_duration_seconds              | Histogram | How long in seconds a Release takes to validate                     |
 | release_total                                    | Counter   | Total number of releases reconciled by the operator.                |
+
+## Distributed tracing
+
+The operator emits OpenTelemetry spans for each release PipelineRun (tenant-collectors, managed-collectors, tenant, managed, final) on completion. Tracing is enabled by pointing the controller at an OTLP/gRPC collector; when the endpoint env var is unset, the operator installs a noop tracer provider and emits nothing.
+
+### Configuration
+
+| Env var | Purpose | Default |
+|---|---|---|
+| `OTEL_EXPORTER_OTLP_ENDPOINT` | OTLP/gRPC collector URL. Unset disables tracing. `http://` sends spans in plaintext; `https://` uses TLS. | *(unset)* |
+| `OTEL_TRACES_SAMPLER` | `always_on`, `always_off`, `traceidratio`, `parentbased_always_off`, `parentbased_traceidratio`. | `parentbased_always_on` |
+| `OTEL_TRACES_SAMPLER_ARG` | Ratio for ratio-based samplers (e.g. `0.1`). | *(unused unless a ratio sampler is selected)* |
+| `TRACING_LABEL_ACTION` | Label key read from the `PipelineRun` or `Release` to populate `cicd.pipeline.action.name`. Empty string disables the attribute. | `delivery.tekton.dev/action` |
+| `TRACING_LABEL_APPLICATION` | Label key read from the `PipelineRun` or `Release` to populate `delivery.tekton.dev.application`. Empty string disables the attribute. | `delivery.tekton.dev/application` |
+| `TRACING_LABEL_COMPONENT` | Label key read from the `PipelineRun` or `Release` to populate `delivery.tekton.dev.component`. Empty string disables the attribute. | `delivery.tekton.dev/component` |
+
+### Emitted spans
+
+Per completed PipelineRun:
+
+- `waitDuration`: `pr.CreationTimestamp` to `pr.Status.StartTime`
+- `executeDuration`: `pr.Status.StartTime` to `pr.Status.CompletionTime`
+
+When a Release is rejected at validation and no PipelineRun is created, a single
+`waitDuration` span is emitted instead, spanning the Release's `CreationTimestamp`
+through the Validated condition's `LastTransitionTime` and carrying
+`cicd.pipeline.result=error` so the rejection still surfaces in the trace.
+
+Parenting follows the W3C Trace Context in the `tekton.dev/pipelinerunSpanContext`
+annotation. The annotation is copied forward from the Release CR onto each
+PipelineRun the controller creates, so all release-stage spans land under a
+single trace when a parent context is present upstream.
+
+### Span attributes
+
+Emitted spans may carry failure messages from PipelineRun and TaskRun conditions (`delivery.tekton.dev.result_message`). These can echo task parameters or step output; treat the OTLP collector at the same access-control level as controller logs.
+
+Span column values: `wait` = PipelineRun `waitDuration`, `execute` = PipelineRun `executeDuration`, `validation wait` = the waitDuration emitted on validation rejection.
+
+| Attribute | Span | Source |
+|---|---|---|
+| `namespace` | all | `PipelineRun` namespace (wait, execute); `Release` namespace (validation wait) |
+| `pipelinerun` | wait, execute | `PipelineRun` name |
+| `delivery.tekton.dev.pipelinerun_uid` | wait, execute | `PipelineRun` UID |
+| `release` | validation wait | `Release` name |
+| `cicd.pipeline.action.name` | all | Label on `PipelineRun` (wait, execute) or `Release` (validation wait); label key configurable via `TRACING_LABEL_ACTION` |
+| `delivery.tekton.dev.application` | all | Label on `PipelineRun` (wait, execute) or `Release` (validation wait); label key configurable via `TRACING_LABEL_APPLICATION` |
+| `delivery.tekton.dev.component` | all | Label on `PipelineRun` (wait, execute) or `Release` (validation wait); label key configurable via `TRACING_LABEL_COMPONENT` |
+| `cicd.pipeline.result` | execute, validation wait | execute: `Succeeded` condition mapped to the semconv `cicd.pipeline.result` enum (`success` / `failure` / `timeout` / `cancellation` / `error`). validation wait: always `error`. |
+| `delivery.tekton.dev.result_message` | execute, validation wait | execute: earliest failing TaskRun's `Succeeded` condition message, falling back to the PipelineRun's own condition message. validation wait: the Validated condition's message. Truncated to 1024 bytes (UTF-8 safe); omitted when empty. |
--- a/api/v1alpha1/release_types.go
+++ b/api/v1alpha1/release_types.go
@@ -422,6 +422,11 @@ func (r *Release) IsValid() bool {
 	return meta.IsStatusConditionTrue(r.Status.Conditions, validatedConditionType.String())
 }
 
+// GetValidatedCondition returns the Release's Validated condition, or nil if not present.
+func (r *Release) GetValidatedCondition() *metav1.Condition {
+	return meta.FindStatusCondition(r.Status.Conditions, validatedConditionType.String())
+}
+
 // IsFailed checks whether the Release has failed.
 func (r *Release) IsFailed() bool {
 	condition := meta.FindStatusCondition(r.Status.Conditions, releasedConditionType.String())
--- a/controllers/release/adapter.go
+++ b/controllers/release/adapter.go
@@ -38,6 +38,7 @@ import (
 	"github.com/konflux-ci/release-service/syncer"
 	"github.com/konflux-ci/release-service/tekton"
 	"github.com/konflux-ci/release-service/tekton/utils"
+	"github.com/konflux-ci/release-service/tracing"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbac "k8s.io/api/rbac/v1"
@@ -734,6 +735,7 @@ func (a *adapter) EnsureReleaseExpirationTimeIsAdded() (controller.OperationResu
 // validation checks.
 func (a *adapter) EnsureReleaseIsValid() (controller.OperationResult, error) {
 	patch := client.MergeFrom(a.release.DeepCopy())
+	releaseAlreadyFinished := a.release.HasReleaseFinished()
 
 	result := controller.Validate(a.validations...)
 	if !result.Valid {
@@ -743,19 +745,24 @@ func (a *adapter) EnsureReleaseIsValid() (controller.OperationResult, error) {
 		a.release.MarkReleaseFailed("Release validation failed")
 	}
 
-	// IsReleasing will be false if MarkReleaseFailed was called
+	// IsReleasing will be false if MarkReleaseFailed was called (this or a prior reconcile).
 	if a.release.IsReleasing() {
 		a.release.MarkValidated()
 		return controller.RequeueOnErrorOrContinue(a.client.Status().Patch(a.ctx, a.release, patch))
 	}
 
+	err := a.client.Status().Patch(a.ctx, a.release, patch)
+	if err == nil && !releaseAlreadyFinished && !result.Valid {
+		a.emitValidationFailureWaitSpan()
+	}
+
 	if !a.release.AreAllProcessingPhasesFinished() {
 		a.logger.Info("EnsureReleaseIsValid: release finished but not all phases finished, continuing so we can complete properly",
 			"Release", fmt.Sprintf("%s/%s", a.release.Namespace, a.release.Name))
-		return controller.RequeueOnErrorOrContinue(a.client.Status().Patch(a.ctx, a.release, patch))
+		return controller.RequeueOnErrorOrContinue(err)
 	}
 
-	return controller.RequeueOnErrorOrStop(a.client.Status().Patch(a.ctx, a.release, patch))
+	return controller.RequeueOnErrorOrStop(err)
 }
 
 // EnsureTenantPipelineProcessingIsTracked is an operation that will ensure that the Release Tenant PipelineRun status
@@ -1173,6 +1180,7 @@ func (a *adapter) getCollectorsPipelineRunBuilder(pipelineType metadata.Pipeline
 
 	return utils.NewPipelineRunBuilder(pipelineType.String(), namespace).
 		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, integrationgitops.PipelinesAsCodePrefix)).
+		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, metadata.SpanContextAnnotation)).
 		WithFinalizer(metadata.ReleaseFinalizer).
 		WithLabels(map[string]string{
 			metadata.PipelinesTypeLabel:    pipelineType.String(),
@@ -1314,6 +1322,7 @@ func (a *adapter) createTenantCollectorsPipelineRun(releasePlan *v1alpha1.Releas
 func (a *adapter) createFinalPipelineRun(releasePlan *v1alpha1.ReleasePlan, snapshot *applicationapiv1alpha1.Snapshot) (*tektonv1.PipelineRun, error) {
 	builder := utils.NewPipelineRunBuilder(metadata.FinalPipelineType.String(), releasePlan.Namespace).
 		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, integrationgitops.PipelinesAsCodePrefix)).
+		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, metadata.SpanContextAnnotation)).
 		WithFinalizer(metadata.ReleaseFinalizer).
 		WithLabels(map[string]string{
 			metadata.ApplicationNameLabel:  releasePlan.Spec.Application,
@@ -1364,6 +1373,7 @@ func (a *adapter) createFinalPipelineRun(releasePlan *v1alpha1.ReleasePlan, snap
 func (a *adapter) createManagedPipelineRun(resources *loader.ProcessingResources, taskRunSpecs []tektonv1.PipelineTaskRunSpec, timeouts tektonv1.TimeoutFields) (*tektonv1.PipelineRun, error) {
 	builder := utils.NewPipelineRunBuilder(metadata.ManagedPipelineType.String(), resources.ReleasePlanAdmission.Namespace).
 		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, integrationgitops.PipelinesAsCodePrefix)).
+		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, metadata.SpanContextAnnotation)).
 		WithFinalizer(metadata.ReleaseFinalizer).
 		WithLabels(map[string]string{
 			metadata.ApplicationNameLabel:  resources.ReleasePlan.Spec.Application,
@@ -1420,6 +1430,7 @@ func (a *adapter) createManagedPipelineRun(resources *loader.ProcessingResources
 func (a *adapter) createTenantPipelineRun(releasePlan *v1alpha1.ReleasePlan, snapshot *applicationapiv1alpha1.Snapshot) (*tektonv1.PipelineRun, error) {
 	builder := utils.NewPipelineRunBuilder(metadata.TenantPipelineType.String(), releasePlan.Namespace).
 		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, integrationgitops.PipelinesAsCodePrefix)).
+		WithAnnotations(metadata.GetAnnotationsWithPrefix(a.release, metadata.SpanContextAnnotation)).
 		WithFinalizer(metadata.ReleaseFinalizer).
 		WithLabels(map[string]string{
 			metadata.ApplicationNameLabel:  releasePlan.Spec.Application,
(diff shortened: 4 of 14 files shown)


mmalina:
LGTM, but I would like somebody else with more expertise in the operator code to take a look as well. Maybe @seanconroy2021 can take a look?

Also, a conflict needs to be resolved.

controllers/release/adapter.go
```diff
@@ -1598,6 +1603,37 @@ func (a *adapter) registerManagedProcessingData(releasePipelineRun *tektonv1.Pip
 	return a.client.Status().Patch(a.ctx, a.release, patch)
 }
 
+// emitTimingSpans emits timing spans for a completed PipelineRun if not already emitted.
```
seanconroy2021: The method has no direct unit tests the register ProcessingStatus tests uses it indirectly but without checking cases.

controllers/release/adapter_test.go
```diff
@@ -4000,6 +4000,22 @@ var _ = Describe("Release adapter", Ordered, func() {
 			Expect(pipelineRun.GetAnnotations()[handler.TypeAnnotation]).To(ContainSubstring("Release"))
 		})
 
+		It("propagates the span context annotation when present", func() {
```
seanconroy2021, line 5096: Span context annotation tests works only for createFinalPipelineRun. The same annotation is in 4 other pipeline creations all of which need to be covered.

tracing/tracing.go
```diff
@@ -0,0 +1,246 @@
...
+			earliestMsg = cond.Message
+		}
+	}
+	return earliestMsg
+}
+
+var setupLog = ctrl.Log.WithName("tracing")
+
+type TracerProvider struct {
+	provider trace.TracerProvider
```
seanconroy2021: The provider field is set but never used the otel.SetTracerProvider() at line 230 is the actual one. Maybe we should remove this field or using it directly rather than relying on the global one.

tracing/tracing.go
```diff
@@ -0,0 +1,246 @@
...
+	shutdown func(context.Context) error
+}
+
+func New(ctx context.Context) (*TracerProvider, error) {
```
seanconroy2021: New() returns (*TracerProvider, error) but every  method returns a nil error failures are handled by falling back to a nooperation provider. This makes the if err != nil check in main.go dead code and gives the idea that initialization can fail. 

It also means if someone later changes New() to actually return an error with a nil  the tracerProvider.Shutdown() in main.go would panic. Either surface the error so we can decide what to do or drop the error return and just return *TracerProvider.

seanconroy2021:
@ci-operator any update on this or can it be moved to draft for the time being ?

tracing/tracing.go
ci-operator: Went with the first option and removed it. `TracerProvider` now carries only the `shutdown` field, and the actual tracer flows through `otel.SetTracerProvider`, which is the only place the SDK looks anyway.

tracing/tracing.go
ci-operator: Dropped the error return per the second option. `New()` is now `(ctx context.Context) *TracerProvider`; exporter and resource errors log inline and fall back to a noop-backed provider, so initialization always succeeds. `main.go` is updated to match, and the latent nil-panic path you flagged is gone.

controllers/release/adapter_test.go
ci-operator: Good catch. Added the propagate/absent pair to the `createManagedCollectorsPipelineRun`, `createTenantCollectorsPipelineRun`, `createTenantPipelineRun`, and `createManagedPipelineRun` `When` blocks. The createFinalPipelineRun tests are unchanged.

controllers/release/adapter.go
ci-operator: Agreed. The register-ProcessingStatus path exercises it but doesn't assert on the annotation. Added a dedicated `When("emitTimingSpans is called")` block in adapter_test.go covering nil PipelineRun (no-op), already-annotated PipelineRun (no re-emission), the noop-provider path (no annotation written), and a completed-run path that asserts the annotation is patched.

seanconroy2021 (dismissed):
lgtm

mmalina:
CI is failing on:
```
Error: tracing/tracing_test.go:116:11: attr.Value.Emit is deprecated: Use [Value.String] instead.  (SA1019)
Error: Process completed with exit code 1.
```
@ci-operator PTAL

controllers/release/adapter.go
```diff
@@ -735,6 +727,7 @@ func (a *adapter) EnsureReleaseIsValid() (controller.OperationResult, error) {
 			return controller.RequeueWithError(result.Err)
```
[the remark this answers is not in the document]
ci-operator: The new HEAD fixes this - emitValidationFailureWaitSpan now checks delivery.tekton.dev/timingEmitted on the Release before emitting and writes the annotation after a successful emit, matching how emitTimingSpans handles PipelineRuns.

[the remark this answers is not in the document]
ci-operator: Span emission is intentionally best-effort here so that tracing failures cannot block reconciliation; OTel errors are handled by the SDK's own error handler.

tracing/timing_spans.go
```diff
@@ -0,0 +1,218 @@
...
+		PipelineRunKey.String(releaseName),
+	)
+	for _, m := range []struct {
+		labelName string
+		key       attribute.Key
+	}{
+		{labels.Action, semconv.CICDPipelineActionNameKey},
+		{labels.Application, DeliveryApplicationKey},
+		{labels.Component, DeliveryComponentKey},
+	} {
```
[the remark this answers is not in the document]
ci-operator: EmitTimingSpans is only called from register*ProcessingStatus paths, each of which checks tekton.IsPipelineRunDone(pr) first, so CompletionTime is always set when it runs.

mmalina:
> @mmalina sorry for the delay. The deprecated attr.Value.Emit() call in tracing/tracing_test.go now uses attr.Value.String(). I also added an idempotency check on emitValidationFailureWaitSpan, adapter-level tests for that path, and a README cleanup. Tests + vet + staticcheck pass.

No worries. I triggered tests again.

tracing/tracing.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+type LabelNames struct {
+	Action      string
+	Application string
+	Component   string
+}
+
+// LoadLabelNames: an explicit empty env value disables the corresponding attribute.
+func LoadLabelNames() LabelNames {
```
mmalina: From Claude:
```
LoadLabelNames() called on every span emission 🔧

  tracing/timing_spans.go:162
  if !tracing.EmitTimingSpans(a.ctx, a.client, pipelineRun, tracing.LoadLabelNames(),
  spanContext) {

  Issue: LoadLabelNames() reads 3 environment variables on every call. Should be cached at
  startup.

  Recommendation: Initialize label names once in main.go or store in adapter.
```
WDYT?

tracing/tracing.go
```diff
@@ -0,0 +1,241 @@
...
+
+func EarliestFailingTaskRunMessage(ctx context.Context, c client.Client, pr *tektonv1.PipelineRun) string {
+	if len(pr.Status.ChildReferences) == 0 {
+		return ""
+	}
+	var (
+		earliestTime *metav1.Time
+		earliestMsg  string
+	)
+	for _, ref := range pr.Status.ChildReferences {
```
mmalina:
```
EarliestFailingTaskRunMessage fetches all TaskRuns ⚠️

  tracing/timing_spans.go:1646-1674
  for _, ref := range pr.Status.ChildReferences {
      tr := &tektonv1.TaskRun{}
      if err := c.Get(ctx, types.NamespacedName{Namespace: pr.Namespace, Name: ref.Name}, tr);
  err != nil {
          continue
      }

  Issue: For a PipelineRun with 20 TaskRuns, this makes 20 sequential API calls. Could be
  expensive.

  Recommendation: Consider batching with client.List() or accepting the PipelineRun's condition
  message as "good enough" for most cases.
```

tracing/tracing.go
```diff
@@ -0,0 +1,241 @@
...
+)
+
+const AttrNamespace = "delivery.tekton.dev"
+
+const (
+	SpanWaitDuration    = "waitDuration"
+	SpanExecuteDuration = "executeDuration"
+)
+
+const MaxResultMessageLen = 1024
```
mmalina, line 65: Why 1024? Maybe add a comment explaining?

tracing/tracing.go
ci-operator: Resolved by initializing at startup via `tracing.SetupLabelNames()` from main.go - matches the codebase's existing `cache.SetupXxxCache` pattern, env reads happen once at controller startup, and `LoadLabelNames` is now just a getter. Tests call `SetupLabelNames` directly after env mutation.

tracing/tracing.go
ci-operator: Resolved by taking your `client.List` suggestion - one call filtered by `MatchingLabels{"tekton.dev/pipelineRun": pr.Name}` regardless of TaskRun count. Test fixtures gained the label, and the "silently skips a missing TaskRun" spec was renamed since List returns only what it finds. If TaskRuns are already pruned when emission runs, List returns empty and `resolveFailureMessage` falls back to the PipelineRun's own condition message - same safety net the per-ref `Get` path used for missing TaskRuns.

tracing/tracing.go
ci-operator: Done - added a godoc on `MaxResultMessageLen` explaining the 1024 cap (typical tracing-backend tag-value limit).

seanconroy2021 (dismissed):
lgtm

mmalina (dismissed):

tracing/tracing.go
johnbieren, line 1: Can you add docstrings to the functions here and in `timing_spans.go`? I think it would help to understand what the functions are used for

ci-operator: Added docstrings to all the exported functions across both files. Let me know if any of them still don't make it clear what they're for.

tracing/timing_spans.go
```diff
@@ -0,0 +1,211 @@
...
+	EmitWaitDuration(parentCtx, pr, labels)
+
+	if pr.Status.CompletionTime != nil {
+		EmitExecuteDuration(parentCtx, pr, labels, resolveFailureMessage(ctx, c, pr))
+	}
+
+	return true
+}
+
+func resolveFailureMessage(ctx context.Context, c client.Client, pr *tektonv1.PipelineRun) string {
```
johnbieren, line 201: missing docstring

tracing/timing_spans.go
```diff
@@ -0,0 +1,211 @@
...
+		if m.labelName == "" {
+			continue
+		}
+		if v := prLabels[m.labelName]; v != "" {
+			span.SetAttributes(m.key.String(v))
+		}
+	}
+}
+
+func setOutcomeAttributes(span trace.Span, pr *tektonv1.PipelineRun, failureMsg string) {
```
johnbieren, line 77: missing docstring

tracing/timing_spans.go
```diff
@@ -0,0 +1,211 @@
...
+	if err := json.Unmarshal([]byte(jsonCarrier), &carrier); err != nil {
+		setupLog.Info("ignoring malformed span context annotation", "error", err)
+		return context.Background(), false
+	}
+	ctx := otel.GetTextMapPropagator().Extract(context.Background(), propagation.MapCarrier(carrier))
+	sc := trace.SpanContextFromContext(ctx)
+	return ctx, sc.IsValid()
+}
+
+func setCommonAttributes(span trace.Span, pr *tektonv1.PipelineRun, labels LabelNames) {
```
johnbieren, line 52: missing docstring

tracing/tracing.go
```diff
@@ -0,0 +1,252 @@
...
+func LoadLabelNames() LabelNames { return labelNames }
+
+func envOr(key, fallback string) string {
+	if v, ok := os.LookupEnv(key); ok {
+		return v
+	}
+	return fallback
+}
+
+func samplerFromEnv() sdktrace.Sampler {
```
johnbieren, line 155: missing docstring

tracing/timing_spans.go
ci-operator: Docstring added.

tracing/timing_spans.go
ci-operator: Docstring added.

tracing/timing_spans.go
ci-operator: Docstring added.

tracing/tracing.go
ci-operator: Docstring added.

tracing/timing_spans.go
johnbieren: did you forget to `git add` or something? I still don't see them

ci-operator: Yes - added, but lost after rebase - sorry for the churn! The docstrings are in now.

controllers/release/adapter.go
```diff
@@ -735,6 +737,7 @@ func (a *adapter) EnsureReleaseIsValid() (controller.OperationResult, error) {
...
 		}
 		a.release.MarkReleaseFailed("Release validation failed")
```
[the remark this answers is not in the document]
davidmogar: I think this might be right. Double patch.

tracing/timing_spans.go
```diff
@@ -0,0 +1,214 @@
...
+		{labels.Action, semconv.CICDPipelineActionNameKey},
+		{labels.Application, DeliveryApplicationKey},
+		{labels.Component, DeliveryComponentKey},
+	} {
+		if m.labelName == "" {
+			continue
+		}
+		if v := releaseLabels[m.labelName]; v != "" {
+			span.SetAttributes(m.key.String(v))
+		}
```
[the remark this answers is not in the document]
ci-operator: `EmitTimingSpans` now requires both StartTime and CompletionTime, emitting both spans atomically. The path where only waitDuration was emitted (and the annotation persisted, blocking executeDuration) is structurally impossible now.

tracing/timing_spans.go
```diff
@@ -0,0 +1,214 @@
...
+	span.End(trace.WithTimestamp(end))
+}
+
+// EmitReleaseValidationFailureWait records the time between a Release's creation and validation rejection, plus its result and failure message.
+func EmitReleaseValidationFailureWait(
+	parentCtx context.Context,
+	releaseName, releaseNamespace string,
+	releaseLabels map[string]string,
+	createdAt, validatedAt time.Time,
+	validationMessage string,
```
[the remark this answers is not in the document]
ci-operator: Added `ReleaseKey` (`"release"`); `EmitReleaseValidationFailureWait` uses it instead of `PipelineRunKey`, since no PipelineRun exists in the validation-failure path.

AGENTS.md
```diff
@@ -21,6 +21,7 @@ syncer/                # Cross-namespace Snapshot syncing
 tekton/                # PipelineRun builders, status helpers, watch predicates
 metadata/              # Label/annotation/finalizer constants and utilities
 metrics/               # Prometheus gauges, histograms, counters for release lifecycle
+tracing/               # OpenTelemetry distributed tracing (per konflux-ci/architecture ADR 0062)
```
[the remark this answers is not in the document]
ci-operator: The previous run of this review recommended this change under `architectural-alignment`: *"Add `tracing/  # OpenTelemetry distributed tracing for PipelineRuns` to the Repository Structure section."* The change is a single directory entry pointing at the new package.

controllers/release/adapter.go
```diff
@@ -1689,6 +1696,69 @@ func (a *adapter) registerManagedProcessingData(releasePipelineRun *tektonv1.Pip
 	return a.client.Status().Patch(a.ctx, a.release, patch)
```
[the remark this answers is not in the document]
ci-operator: Reverted to fire-and-forget since tracing is best-effort; patch errors are logged, not propagated. Also resolves the parallel **idempotency** Low on `emitValidationFailureWaitSpan` - both functions now share the same pattern.

tracing/tracing.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+	switch cond.Reason {
+	case tektonv1.PipelineRunReasonCancelled.String(),
+		tektonv1.PipelineRunReasonCancelledRunningFinally.String():
+		return semconv.CICDPipelineResultCancellation
+	case tektonv1.PipelineRunReasonTimedOut.String():
+		return semconv.CICDPipelineResultTimeout
+	case tektonv1.PipelineRunReasonFailed.String():
+		return semconv.CICDPipelineResultFailure
+	}
```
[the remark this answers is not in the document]
ci-operator: Added `PipelineRunReasonStoppedRunningFinally`. All three of Tekton's user-cessation condition reasons (`Cancelled`, `CancelledRunningFinally`, `StoppedRunningFinally`) now map to `cancellation`.

controllers/release/adapter.go
```diff
@@ -1689,6 +1696,69 @@ func (a *adapter) registerManagedProcessingData(releasePipelineRun *tektonv1.Pip
...
+		a.release.Labels,
+		createdAt, validatedAt,
+		cond.Message,
+		tracing.LoadLabelNames(),
+	) {
+		a.logger.Info("Skipped release-validation waitDuration emission", "release.Name", a.release.Name)
+		return
+	}
+
+	patch := client.MergeFrom(a.release.DeepCopy())
```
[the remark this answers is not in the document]
ci-operator: Reverted to fire-and-forget since tracing is best-effort; patch errors are logged, not propagated. Also resolves the parallel **idempotency** Low on `emitValidationFailureWaitSpan` - both functions now share the same pattern.

tracing/tracing.go
```diff
@@ -0,0 +1,254 @@
...
+		setupLog.Error(err, "failed to create OTLP exporter, using noop tracer provider")
+		return &TracerProvider{shutdown: func(context.Context) error { return nil }}
+	}
+
+	res, err := resource.Merge(
+		resource.Default(),
+		resource.NewWithAttributes(
+			semconv.SchemaURL,
+			semconv.ServiceName(TracerName),
+		),
```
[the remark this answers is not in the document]
ci-operator: README OTEL endpoint row notes `http://` sends spans in plaintext; `https://` uses TLS.

metadata/labels.go
```diff
@@ -60,6 +60,12 @@ const (
 const (
 	// PipelinesAsCodePrefix contains the prefix applied to labels and annotations copied from Pipelines as Code resources.
 	PipelinesAsCodePrefix = "pac.test.appstudio.openshift.io"
+
```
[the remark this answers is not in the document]
ci-operator: Tracing annotation comments describe purpose; namespace is visible in the value.

metadata/labels.go
```diff
@@ -60,6 +60,12 @@ const (
...
+	// SpanContextAnnotation is the annotation key for propagating span context for distributed tracing.
+	SpanContextAnnotation = "tekton.dev/pipelinerunSpanContext"
+
```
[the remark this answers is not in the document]
ci-operator: Tracing annotation comments describe purpose; namespace is visible in the value.

metadata/labels.go
```diff
@@ -60,6 +60,12 @@ const (
 const (
```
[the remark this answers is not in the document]
ci-operator: Tracing annotations moved to their own const block.

controllers/release/adapter.go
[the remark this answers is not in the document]
ci-operator: Yeah - emitValidationFailureWaitSpan now patches a deep-copy of a.release; otherwise controller-runtime's in-place update would clobber the in-memory mutations the outer status patch needs. Test: `preserves in-memory status mutations on a.release across the annotation patch`.

AGENTS.md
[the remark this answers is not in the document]
ci-operator: Addresses the `architectural-alignment` finding (add `tracing/` to Repository Structure). The previous run's `architectural-decision` Low suggested an ADR 0062 reference here; we carry that rationale in `metadata/labels.go` where the namespace choice lives, keeping this file free of cross-repo references.

metadata/labels.go
```diff
@@ -62,6 +62,15 @@ const (
 	PipelinesAsCodePrefix = "pac.test.appstudio.openshift.io"
 )
 
+// Distributed tracing annotations
+const (
+	// SpanContextAnnotation propagates span context.
+	SpanContextAnnotation = "tekton.dev/pipelinerunSpanContext"
+
```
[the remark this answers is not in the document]
ci-operator: `delivery.tekton.dev` is the shared cross-service tracing namespace used by PaC and integration-service for the same attributes. Renaming would break cross-service consistency.

tracing/timing_spans.go
```diff
@@ -0,0 +1,211 @@
...
+	labels LabelNames,
+) bool {
+	if _, ok := otel.GetTracerProvider().(*sdktrace.TracerProvider); !ok {
+		return false
```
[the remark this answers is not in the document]
ci-operator: Called only from reconciliation, where `CreationTimestamp` is set by the API server.

tracing/tracing.go
```diff
@@ -0,0 +1,255 @@
...
+	)
+	if err != nil {
+		setupLog.Error(err, "failed to create resource")
```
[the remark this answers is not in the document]
ci-operator: Documented in the README.

controllers/release/adapter.go
```diff
@@ -1689,6 +1696,68 @@ func (a *adapter) registerManagedProcessingData(releasePipelineRun *tektonv1.Pip
 	return a.client.Status().Patch(a.ctx, a.release, patch)
 }
 
+// emitValidationFailureWaitSpan emits a waitDuration span measuring how long a Release waited before rejection.
```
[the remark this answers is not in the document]
ci-operator: No PipelineRun exists in the validation-failure path.

metadata/labels.go
```diff
@@ -62,6 +62,15 @@ const (
 	PipelinesAsCodePrefix = "pac.test.appstudio.openshift.io"
 )
 
+// Distributed tracing annotations
+const (
+	// SpanContextAnnotation propagates span context; tekton.dev domain follows Tekton ecosystem convention.
+	SpanContextAnnotation = "tekton.dev/pipelinerunSpanContext"
```
seanconroy2021, line 68: Are these new labels then owned by PAC or Tekton ?

ci-operator: These aren't PaC-owned. `tekton.dev/pipelinerunSpanContext` is Tekton's own contract for cross-controller coordination - specifically for adopting an external traceparent, which is exactly how release-service, PaC, and integration-service use it. `delivery.tekton.dev/timingEmitted` is Konflux-CI's; the annotation is per-controller dedupe, and the shared `delivery.tekton.dev` domain distinguishes delivery-chain traces from execution traces (per the labels.go comment).

seanconroy2021: Ack thanks for explaining :)

controllers/release/adapter.go
```diff
@@ -1794,7 +1768,11 @@ func (a *adapter) registerTenantCollectorsProcessingStatus(pipelineRun *tektonv1
 		a.release.MarkReleaseFailed("Release processing failed on tenant collectors pipelineRun")
 	}
 
-	return a.client.Status().Patch(a.ctx, a.release, patch)
+	if err := a.client.Status().Patch(a.ctx, a.release, patch); err != nil {
```
seanconroy2021: The patch then emit block is repeated across all five register functions. It may be a good idea to use a small helper function like patchStatusAndEmitSpans(patch, pipelineRun). It would be cleaner and easier to test. This is a nit, so you can ignore it if you don't think so.

controllers/release/adapter.go
seanconroy2021, line 736: This is a bit hard to follow with the justFailedValidation variable and the nested guard. Could we capture wasAlreadyFinished at the top before anything runs, then use that to decide whether to emit? That way we don't need the extra check around MarkReleaseFailed either since it already skips internally when finished.

ci-operator: Captured the finished state up front and gated the emit on it directly. Dropped the wrapping guard around MarkReleaseFailed since its internal idempotency check already handles that path.

controllers/release/adapter.go
ci-operator: Extracted the pattern into a small helper on the adapter. All five register methods delegate to it now instead of inlining the block.

seanconroy2021 (dismissed):
LGTM

api/v1alpha1/release_types.go
```diff
@@ -395,6 +395,11 @@ func (r *Release) IsValid() bool {
 	return meta.IsStatusConditionTrue(r.Status.Conditions, validatedConditionType.String())
 }
 
+// GetValidatedCondition returns the Release's Validated condition, or nil if not present.
+func (r *Release) GetValidatedCondition() *metav1.Condition {
```
johnbieren, line 426: can you add unit tests for this in release_types_test.go?

ci-operator: Added a GetValidatedCondition block after IsValid in release_types_test.go. One case sets Released and Validated together and asserts we get the Validated condition back, exercising the type filter. Another asserts nil when Validated is absent.

johnbieren: perfect, thanks

johnbieren:
Changes lgtm but you have a conflict to resolve now

johnbieren (dismissed):

controllers/release/adapter_test.go
```diff
@@ -2317,6 +2351,35 @@ var _ = Describe("Release adapter", Ordered, func() {
...
+		It("does not re-emit the validation-failure span when the Release has already been marked failed", func() {
+			prev := otel.GetTracerProvider()
+			exporter := &recordingExporter{}
+			tp := sdktrace.NewTracerProvider(sdktrace.WithSyncer(exporter))
+			otel.SetTracerProvider(tp)
+			DeferCleanup(func() {
+				otel.SetTracerProvider(prev)
+				_ = tp.Shutdown(adapter.ctx)
+			})
+
```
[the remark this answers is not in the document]
ci-operator: Added MarkValidationFailed to the test setup so a Validated condition is present. Test now exercises the releaseAlreadyFinished guard.

README.md
```diff
@@ -51,3 +51,48 @@ by default, this operator exports the following custom metrics:
 | release_pre_processing_duration_seconds          | Histogram | How long in seconds a Release takes to start processing             |
 | release_validation_duration_seconds              | Histogram | How long in seconds a Release takes to validate                     |
 | release_total                                    | Counter   | Total number of releases reconciled by the operator.                |
+
+## Distributed tracing
+
```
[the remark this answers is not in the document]
ci-operator: Span column now names wait, execute, validation wait. Release row added. cicd.pipeline.result and result_message extended to validation wait.

tracing/timing_spans.go
```diff
@@ -0,0 +1,211 @@
...
+
+	span.SetAttributes(
+		NamespaceKey.String(releaseNamespace),
+		ReleaseKey.String(releaseName),
+	)
+	for _, m := range []struct {
+		labelName string
+		key       attribute.Key
+	}{
+		{labels.Action, semconv.CICDPipelineActionNameKey},
```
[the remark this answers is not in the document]
ci-operator: Sensitivity paragraph added above the attributes table.

tracing/timing_spans.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	span.SetAttributes(semconv.CICDPipelineResultError)
+	if validationMessage != "" {
+		span.SetAttributes(DeliveryResultMessageKey.String(TruncateResultMessage(validationMessage)))
+	}
+
+	span.End(trace.WithTimestamp(validatedAt))
+	return true
+}
```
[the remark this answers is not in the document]
ci-operator: Provider check extracted into a shared helper. Emits a one-time otel.Handle warning when the provider is neither sdktrace nor noop.

tracing/tracing.go
```diff
@@ -0,0 +1,257 @@
...
+)
+
+// ResultEnum returns the cicd.pipeline.result span attribute matching the PipelineRun's terminal condition.
+func ResultEnum(cond *apis.Condition) attribute.KeyValue {
+	if cond.Status == corev1.ConditionTrue {
+		return semconv.CICDPipelineResultSuccess
+	}
+	switch cond.Reason {
+	case tektonv1.PipelineRunReasonCancelled.String(),
+		tektonv1.PipelineRunReasonCancelledRunningFinally.String(),
```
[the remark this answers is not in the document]
ci-operator: Nil guard added. Nil-case test added.

controllers/release/adapter.go
```diff
@@ -1911,6 +1922,49 @@ func (a *adapter) registerManagedProcessingData(releasePipelineRun *tektonv1.Pip
...
+// emitTimingSpans emits timing spans for a completed PipelineRun.
+func (a *adapter) emitTimingSpans(pipelineRun *tektonv1.PipelineRun) {
+	if pipelineRun == nil {
+		return
+	}
+	spanContext := pipelineRun.Annotations[metadata.SpanContextAnnotation]
+	tracing.EmitTimingSpans(a.ctx, a.client, pipelineRun, tracing.LoadLabelNames(), spanContext)
+}
+
+// patchStatusAndEmitSpans persists the Release status and, on success, emits timing spans.
```
seanconroy2021, line 1959: Missing tests for this new function?

ci-operator: Added tests for both branches; failure case exercises the guard.

seanconroy2021 (approved):

controllers/release/adapter.go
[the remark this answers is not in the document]
johnbieren: @davidmogar should this be resolved now?

ach912 (approved):

controllers/release/adapter.go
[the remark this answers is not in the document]
davidmogar: I think so. I don't see any double patching now.

davidmogar (approved):

johnbieren (approved):

ach912:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @ach912 |
| **Timestamp** | 2026-07-16T13:28:20.438Z |

Approved

PR #1560, opened 2026-05-05
feat(RELEASE-2422): add AGENTS.md

FilipNikolovski:
Added AGENTS.md file, and symlinked CLAUDE.md to it.

Assisted-by: Claude Code

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,64 @@
+# AGENTS.md
+
+Release Service is a Kubernetes operator that orchestrates software release pipelines in Konflux CI. It watches Release CRs, validates them, creates Tekton PipelineRuns, tracks their progress, and records results.
+
+## Technology Stack
+
+- **Language**: Go 
+- **Framework:** controller-runtime
+- **CRDs**: Release, ReleasePlan, ReleasePlanAdmission, ReleaseServiceConfig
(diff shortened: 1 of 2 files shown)


mmalina (approved):

davidmogar (approved):

ach912 (approved):

johnbieren (commented):
I know this is already merged, but I think it would be nice for it to actually explain some of our tendencies in this repo. Things like our operation functions use gates to ensure the controller is idempotent. We only patch the resource in certain functions (you can look at the release validation workflow to see this). Also, this is over 60 lines. https://redhat.atlassian.net/browse/KONFLUX-12747 says it cannot be, and that a CI check should ensure that (which was added in the catalog repo PR)

FilipNikolovski:
> I know this is already merged, but I think it would be nice for it to actually explain some of our tendencies in this repo. Things like our operation functions use gates to ensure the controller is idempotent. We only patch the resource in certain functions (you can look at the release validation workflow to see this). Also, this is over 60 lines. https://redhat.atlassian.net/browse/KONFLUX-12747 says it cannot be, and that a CI check should ensure that (which was added in the catalog repo PR)

Yes, I realized this after the fact, I've created a new PR for the CI check https://github.com/konflux-ci/release-service/pull/1596, I'll also address your feedback over there.

PR #1590, opened 2026-05-12
fix(RELEASE-2092): fix KubeArchive connectivity and logging

querti:
The KubeArchive client was connecting via the external route, which is blocked by the controller's egress NetworkPolicy, causing a 30-second hang followed by silent failure. Switch to the in-cluster service URL (kubearchive-api-server.product-kubearchive.svc.cluster.local:8081), load the OpenShift service-serving CA for TLS verification, and add a 10-second HTTP client timeout.

Remove the Client struct and caching mechanism — discovery and client construction are lightweight and the cache was ineffective since the loader was recreated on every reconciliation.

Add error-level logging on all failure paths so KubeArchive issues are visible in controller logs. Allow egress to the KubeArchive API server on port 8081 in the NetworkPolicy.

Assisted-by: Cursor

### DIFF
--- a/kubearchive/client.go
+++ b/kubearchive/client.go
@@ -2,6 +2,7 @@ package kubearchive
 
 import (
 	"context"
+	"crypto/x509"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -9,7 +10,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
-	"sync"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


kubearchive/client.go
```diff
@@ -170,67 +176,72 @@ func (c *Client) getFromList(ctx context.Context, httpClient *http.Client,
...
 	if err != nil {
 		return nil, fmt.Errorf("loading in-cluster config: %w", err)
 	}
-	cfg.TLSClientConfig.CAFile = ""
-	cfg.TLSClientConfig.CAData = nil
-	return rest.HTTPClientFor(cfg)
+
+	pool := x509.NewCertPool()
+	if !pool.AppendCertsFromPEM(caData) {
+		return nil, fmt.Errorf("failed to parse service-serving CA bundle")
```
happybhati, line 223: Did you get a chance to test this TLS path on a live cluster?

My understanding is that `rest.TransportFor` with a bearer token returns a wrapped `RoundTripper` (e.g. `bearerTokenRoundTripper`), not a bare `*http.Transport` — so this type assertion might always be `false` and the CA pool never actually gets injected. If that's the case, requests to the in-cluster service URL would fail TLS verification since the system cert pool won't have the OpenShift service-serving CA.

Could be I'm wrong about how `rest.TransportFor` behaves here — is this something you verified works end-to-end, or am I missing something?

loader/loader.go
```diff
@@ -45,25 +45,23 @@ type ObjectLoader interface {
...
+// resources from KubeArchive, matching kubearchive.Get.
+type KubeArchiveGetFunc func(ctx context.Context, cli client.Client, gvr schema.GroupVersionResource,
+	namespace, name string, obj client.Object) error
 
 type loader struct {
-	kaClient KubeArchiveClient
+	kaGet KubeArchiveGetFunc
 }
 
 func NewLoader() ObjectLoader {
```
happybhati, line 57: Small question on the caching removal — the PR description says the cache was ineffective because the loader is recreated per reconciliation, but looking at how `NewLoader()` is wired into the controller, it seems like the loader is created once at setup time and reused across reconciliations. If that's the case, the old `Client` struct caching was actually working.

Is the removal intentional just for simplicity, or was there a specific case where it caused issues? Just want to make sure we're not adding 2 extra ConfigMap GETs (kubearchive-api-url + openshift-service-ca.crt) on every archived Snapshot lookup unnecessarily. Happy to be corrected if I'm reading the controller wiring wrong!

happybhati (commented):
Overall the approach looks right — switching to the in-cluster service URL + adding the NetworkPolicy egress rule makes sense, and the error logging improvements are a nice cleanup. Left two inline questions on things I wasn't sure about.

kubearchive/client.go
querti: I have tested it on a local Kind cluster and the test had succeeded. But questioning Claude again, apparently the test was flawed due to using the same CA cert, which invalidated the test results. So this was indeed a bug. THe code has bene rewritten to correctly inject the certificate and I have now tested the negative scenario as well (wrong CA) to see if the connection failed.

Good catch, thanks!

loader/loader.go
querti: The caching was written by an LLM, as you can imagine, but I saw in the logs with my own eyes that the cache was recreated on every reconciliation loop, defeating the purpose and just creating noise in the logs. According to Claude, implementing a persistent cache would require significantly more complicated implementation, so I decided to just remove it. Kubearchive code path is only a fallback, so it's not expected to trigger frequently, and the two saved GETs are not worth the added complexity in my opinion. WDYT?

kubearchive/client.go
```diff
@@ -170,67 +175,66 @@ func (c *Client) getFromList(ctx context.Context, httpClient *http.Client,
...
+// server over the cluster network, avoiding external route egress restrictions.
 func discoverAPIURL(ctx context.Context, cli client.Client) (string, error) {
 	cm := &corev1.ConfigMap{}
 	if err := cli.Get(ctx, types.NamespacedName{Namespace: kaNamespace, Name: apiURLConfigMapName}, cm); err != nil {
 		return "", fmt.Errorf("ConfigMap %q not found in namespace %q: %w", apiURLConfigMapName, kaNamespace, err)
 	}
-	if u, ok := cm.Data[apiURLConfigMapKey]; ok &&
-		(strings.HasPrefix(u, "http://") || strings.HasPrefix(u, "https://")) {
-		return strings.TrimRight(u, "/"), nil
+	return fmt.Sprintf("https://%s.%s.svc.cluster.local:%d", kaServiceName, kaNamespace, kaServicePort), nil
```
theflockers, line 188: I would rather always return what is in the ConfigMap or the error. IMHO having the default might give the wrong assumption that the url is correct.

If we want to have a "default" url,  maybe you should get the actual domain name to add to the Sprintf.

loader/loader.go
happybhati: That makes sense — if the cache was getting recreated every loop in practice, removing it is the right call. And agreed, since this is a fallback path it's not worth the added complexity. Thanks for the context!

happybhati (approved):
LGTM — the TLS fix looks good and the caching explanation makes sense. Approving pending resolution of Leandro's open comment on the hardcoded in-cluster URL in `discoverAPIURL`.

kubearchive/client.go
querti: The configmap fetch now only works as a sanity check - that the Kubearchive is deployed. The reason why we construct the internal url instead of using the external one is that egress network policy blocks all connections by default. The policy is much easier to set up with internal connections - we only need namespace and port. If we were to use external connection we would also need IP address which is too open and also more fragile. WDYT?

ach912:
@querti the following items are needed before this PR can be approved for production:



### - [ ] Risk Assessment

Add a `## Risk Assessment` section to the PR description:

```markdown
## Risk Assessment

**Risk Level:** <Low / Medium / High / Very High>
**Description:** <Brief description of what could go wrong>
**Rollback:** <How to revert if needed>
```

Replace the `<…>` placeholders with your actual values.

### - [ ] Staging Validation

Add a comment with the staging PR URL in this format:

```markdown
## Validation

Staging PR: <your-staging-pr-url>
```

The dashboard needs the full GitHub PR URL to check if it has been merged and calculate soak time.
Replace the `<…>` placeholder with the actual staging PR link.



---

> Posted by the Konflux approval dashboard.

johnbieren (approved):

davidmogar (approved):

ach912 (approved):

ach912:
@querti the following items are needed before this PR can be approved for production:



### - [ ] Risk Assessment

Add a `## Risk Assessment` section to the PR description:

```markdown
## Risk Assessment

**Risk Level:** <Low / Medium / High / Very High>
**Description:** <Brief description of what could go wrong>
**Rollback:** <How to revert if needed>
```

Replace the `<…>` placeholders with your actual values.

### - [ ] Staging Validation

Add a comment with the staging PR URL in this format:

```markdown
## Validation

Staging PR: <your-staging-pr-url>
```

The dashboard needs the full GitHub PR URL to check if it has been merged and calculate soak time.
Replace the `<…>` placeholder with the actual staging PR link.



---

> Posted by the Konflux approval dashboard.

davidmogar:
## Production Approval Record

| Field | Value |
|---|---|
| **Action** | APPROVED |
| **Reviewer** | @davidmogar |
| **Timestamp** | 2026-05-25T10:31:09.567Z |

PR #1596, opened 2026-05-13
feat(RELEASE-2422): add AGENTS.md lint check and document key patterns

FilipNikolovski:
Add CI job to enforce AGENTS.md stays under 60 non-empty lines.
Add error handling, idempotency, and patching discipline to
Key Patterns section in AGENTS.md.

Assisted-by: Claude Code

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -113,6 +113,18 @@ jobs:
           fetch-depth: 0
       - name: Check if dockerimage build is working
         run: docker build -f ./Dockerfile .
+  check-agents-md:
+    name: Check AGENTS.md line count
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+      - name: Enforce AGENTS.md max 60 lines
+        run: |
+          lines=$(grep -c '.' AGENTS.md)
+          if [ "$lines" -gt 60 ]; then
+            echo "AGENTS.md has $lines non-empty lines (max 60)"
+            exit 1
+          fi
   gitlint:
     name: Run gitlint checks
     runs-on: ubuntu-latest
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,13 +26,10 @@ main.go                # Entry point — registers controllers and webhooks
 ```
 
 ## Architecture
-
 ### Adapter Pattern
-
 Each controller delegates to an **adapter** (`adapter.go`) that holds the K8s client, resource under reconciliation, an `ObjectLoader`, and a `Syncer`. All domain logic lives in adapter methods, not in the controller.
 
 ### Reconciliation Pipeline
-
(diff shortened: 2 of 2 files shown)


mmalina (dismissed):

mmalina:
Could it be 60 non-empty lines? It seems kinda crazy to remove empty lines (which take no context space, but improve human readability) just to satisfy this arbitrary rule.

FilipNikolovski:
> Could it be 60 non-empty lines? It seems kinda crazy to remove empty lines (which take no context space, but improve human readability) just to satisfy this arbitrary rule.

We could do that, not sure if the 60 lines is a hard rule though, because otherwise the file would have more than 60 actual lines.

mmalina:
> > Could it be 60 non-empty lines? It seems kinda crazy to remove empty lines (which take no context space, but improve human readability) just to satisfy this arbitrary rule.
> 
> We could do that, not sure if the 60 lines is a hard rule though, because otherwise the file would have more than 60 actual lines.

I'm not really sure what you're saying :) If you mean that with my proposal the file could have more than 60 lines, then absolutely, that's the point - I propose that only non-empty lines would count towards the limit. So let's do it?

FilipNikolovski:
> > > Could it be 60 non-empty lines? It seems kinda crazy to remove empty lines (which take no context space, but improve human readability) just to satisfy this arbitrary rule.
> 
> > 
> 
> > We could do that, not sure if the 60 lines is a hard rule though, because otherwise the file would have more than 60 actual lines.
> 
> 
> 
> I'm not really sure what you're saying :) If you mean that with my proposal the file could have more than 60 lines, then absolutely, that's the point - I propose that only non-empty lines would count towards the limit. So let's do it?

I mean I'm not sure why the 60 lines limit should be there in the first place, that's why I'm asking. Why should only non empty line count towards this arbitrary limit?

mmalina:
> > > > Could it be 60 non-empty lines? It seems kinda crazy to remove empty lines (which take no context space, but improve human readability) just to satisfy this arbitrary rule.
> > 
> > 
> > > 
> > 
> > 
> > > We could do that, not sure if the 60 lines is a hard rule though, because otherwise the file would have more than 60 actual lines.
> > 
> > 
> > I'm not really sure what you're saying :) If you mean that with my proposal the file could have more than 60 lines, then absolutely, that's the point - I propose that only non-empty lines would count towards the limit. So let's do it?
> 
> I mean I'm not sure why the 60 lines limit should be there in the first place, that's why I'm asking. Why should only non empty line count towards this arbitrary limit?

Because of what I said? ;) Because if it includes empty lines, then as a result we will end up removing empty lines to fit in the limit, but those empty lines take up no context - which is likely the main motivation for having a limit.

FilipNikolovski:
@mmalina I've updated the CI check, and also updated the AGENTS.md file as per @johnbieren suggestions.

johnbieren (approved):

happybhati (approved):

davidmogar (approved):

ach912 (approved):

PR #1748, opened 2026-06-30
docs(RELEASE-2588): created ADRs

davidmogar:
Add directory to hold documents capturing architectural decisions.

Generated-by: Claude


seanconroy2021 (approved):
LGTM - So when will we create a new record for each ticket, or only for big changes?

davidmogar:
> LGTM - So when will we create a new record for each ticket, or only for big changes?

TBD but it makes sense to do it for important changes. That way AI will work better.

PR #1864, opened 2026-08-05
test(loader): fix flaky GetReleasePipelineRunAttempt test

mmalina:
## Summary
- `GetReleasePipelineRunAttempt` test created PipelineRuns and
  immediately asserted on the lookup result via the cached client,
  racing the informer cache sync
- when the cache hadn't observed the create yet, the lookup returned
  nil and the test panicked on a nil pointer dereference instead of
  failing cleanly
- this has been failing roughly half of all PR CI runs over the last
  day, unrelated to the PRs' own changes, e.g.
  https://github.com/konflux-ci/release-service/pull/1848
- wraps each assertion in `Eventually()`, matching the pattern
  already used elsewhere in this suite for cache-backed reads

## Test plan
- [x] `go test ./loader/...` passes
- [x] ran the loader suite 5x in a row locally with no failures
  (previously flaky roughly 50% of the time)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -787,17 +787,20 @@ var _ = Describe("Release Adapter", Ordered, func() {
 			retryPipelineRun2.Labels[metadata.ReleaseAttemptLabel] = "2"
 			Expect(k8sClient.Create(ctx, retryPipelineRun2)).To(Succeed())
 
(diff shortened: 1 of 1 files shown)


seanconroy2021 (approved):
LGTM :)

PR #1940, opened 2026-09-09
fix(RELEASE-2681): don't hard-fail on network errors with token

happybhati:
## Summary

Follow-up to #1753 (RELEASE-2681), which added authenticated GitHub branch→SHA
resolution (`GITHUB_TOKEN`) and made `ResolveBranchToSHA` failures hard-fail
instead of silently falling back, to stop rate-limit exhaustion from passing
raw branch names as SHAs.

That PR caused a staging regression and was rolled back: staging's egress
`NetworkPolicy` does not allow traffic to `github.com`, so `ResolveBranchToSHA`
hit a network-level error on every git-resolver `ReleasePlanAdmission`. The
old `failHard` check treated *any* GitHub-URL failure as fatal once
`GITHUB_TOKEN` was configured, so every one of those releases hard-failed and
stalled instead of falling back to the branch name, until Sean reverted the
deployed image.

This PR:

1. **Narrows `failHard`** to only trigger on a confirmed rate limit or a
   token explicitly rejected as bad/expired. Any other remote-access failure
   (network, DNS, TLS, timeout — including the exact staging case above) now
   falls back to the branch name, same as before `GITHUB_TOKEN` existed.
2. **Adds GitHub's published IP ranges** to the controller's egress
   `NetworkPolicy`, additively.

## Related Issues

Fixes #

## Testing

- [ ] Tests added/updated
- [ ] All tests pass (`make test`)
- [ ] Changes verified locally

## Checklist

- [ ] Code follows Kubernetes coding conventions
- [ ] Commit messages follow conventional format: `type(RELEASE-NNNN): description`
- [ ] Documentation updated if needed
- [ ] Generated manifests updated (`make manifests generate`)

## Controller Promotion (skip if this is a Mintmaker/bot PR)

Once your infra-deployments development PR is merged, you are responsible for
promoting the release-service controller to staging and production.

- [ ] Infra-deployments `development` PR merged (automated by pipeline)
- [ ] Controller promoted to `staging` — [promote-overlay script](https://github.com/konflux-ci/release-service-utils/tree/main/ci/promote-overlay)
- [ ] Controller promoted to `production` — [promote-overlay script](https://github.com/konflux-ci/release-service-utils/tree/main/ci/promote-overlay)

### DIFF
--- a/config/manager/network_policy.yaml
+++ b/config/manager/network_policy.yaml
@@ -52,3 +52,105 @@ spec:
       ports:
         - port: 5353
           protocol: UDP
+    # RELEASE-2681: allow HTTPS egress to github.com so ResolveBranchToSHA can
+    # authenticate with GITHUB_TOKEN instead of silently falling back to the
+    # branch name. Full "git" range set from https://api.github.com/meta as of
+    # 2026-09-09 - GitHub rotates the individual /32s over time, so re-fetch
+    # that endpoint and update this list if resolution starts failing again
+    # with a network/TLS error instead of a clean auth response.
+    - to:
+      - ipBlock:
+          cidr: 140.82.112.0/20
+      - ipBlock:
+          cidr: 143.55.64.0/20
+      - ipBlock:
+          cidr: 172.182.252.133/32
+      - ipBlock:
+          cidr: 172.182.252.135/32
+      - ipBlock:
+          cidr: 185.199.108.0/22
+      - ipBlock:
+          cidr: 192.30.252.0/22
+      - ipBlock:
+          cidr: 20.175.192.146/32
+      - ipBlock:
+          cidr: 20.175.192.147/32
+      - ipBlock:
+          cidr: 20.199.39.227/32
+      - ipBlock:
+          cidr: 20.199.39.232/32
+      - ipBlock:
+          cidr: 20.200.245.247/32
+      - ipBlock:
+          cidr: 20.200.245.248/32
+      - ipBlock:
+          cidr: 20.201.28.151/32
+      - ipBlock:
+          cidr: 20.201.28.152/32
+      - ipBlock:
+          cidr: 20.205.243.160/32
+      - ipBlock:
+          cidr: 20.205.243.166/32
+      - ipBlock:
+          cidr: 20.207.73.82/32
+      - ipBlock:
+          cidr: 20.207.73.83/32
+      - ipBlock:
+          cidr: 20.217.135.4/32
+      - ipBlock:
+          cidr: 20.217.135.5/32
+      - ipBlock:
+          cidr: 20.233.83.145/32
+      - ipBlock:
+          cidr: 20.233.83.149/32
+      - ipBlock:
+          cidr: 20.26.156.214/32
+      - ipBlock:
+          cidr: 20.26.156.215/32
+      - ipBlock:
+          cidr: 20.27.177.113/32
+      - ipBlock:
+          cidr: 20.27.177.118/32
+      - ipBlock:
+          cidr: 20.29.134.19/32
+      - ipBlock:
+          cidr: 20.29.134.23/32
+      - ipBlock:
+          cidr: 20.87.245.0/32
+      - ipBlock:
+          cidr: 20.87.245.4/32
+      - ipBlock:
+          cidr: 4.208.26.197/32
+      - ipBlock:
+          cidr: 4.208.26.198/32
+      - ipBlock:
+          cidr: 4.225.11.194/32
+      - ipBlock:
+          cidr: 4.225.11.200/32
+      - ipBlock:
+          cidr: 4.228.31.145/32
+      - ipBlock:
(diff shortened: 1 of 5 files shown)


seanconroy2021 (approved):
LGTM

happybhati:
> > On the trusted-ca thing — good catch, that explains a lot. I looked around infra-deployments and [build-service](https://github.com/redhat-appstudio/infra-deployments/blob/0c5d5074b53c736413feefc1c7708bf6b7987222/components/build-service/components/rh-certs/kustomization.yaml)/[mintmaker](https://github.com/redhat-appstudio/infra-deployments/blob/bff38c1914022f3921da124b79a306df76799fda/components/mintmaker/components/rh-certs/kustomization.yaml)/[integration](https://github.com/redhat-appstudio/infra-deployments/blob/67db5d12e0c531cda56c9bea1bff8a66fd670a01/components/integration/rh-certs/kustomization.yaml) all already solve this with a small reusable rh-certs kustomize component ([ConfigMap labeled config.openshift.io/inject-trusted-cabundle, mounted at /etc/pki/ca-trust/extracted/pem](https://github.com/redhat-appstudio/infra-deployments/blob/0c5d5074b53c736413feefc1c7708bf6b7987222/components/build-service/components/rh-certs/add-rh-certs-patch.yaml)) — [release-service](https://issues.redhat.com/browse/RELEASE-service) just never had it. Given [our image is ubi-micro](https://github.com/konflux-ci/release-service/blob/bffa7e4038321a3e5ba49a6db35cf13523d90f7e/Dockerfile#L31-L33) with literally zero CA bundle baked in, seems like the same pattern you already tested would work here too. WDYT?
> 
> Yeah, I was using the trusted-ca CM on Kind and it was working. So say just copy one of the other services to match Release :) So you dont need any change in this PR then, since it's going to be a patch in infra deployments right?

Thanks for the quick review, Sean! 🙏 Created the infra PR for dev and staging: [#14128](https://github.com/redhat-appstudio/infra-deployments/pull/14128) — used the same rh-certs pattern build-service/mintmaker/integration already have. Production follows once this is verified working.

config/manager/network_policy.yaml
```diff
@@ -52,3 +52,105 @@ spec:
       ports:
         - port: 5353
           protocol: UDP
+    # RELEASE-2681: allow HTTPS egress to github.com so ResolveBranchToSHA can
+    # authenticate with GITHUB_TOKEN instead of silently falling back to the
+    # branch name. Full "git" range set from https://api.github.com/meta as of
+    # 2026-09-09 - GitHub rotates the individual /32s over time, so re-fetch
+    # that endpoint and update this list if resolution starts failing again
+    # with a network/TLS error instead of a clean auth response.
```
davidmogar, line 60: This is going to bite us again in the future. It would be nice to have at least an alert to know if we need to update...

happybhati: Yeah, this is the same problem integration-service hit in [#266](https://github.com/konflux-ci/integration-service/pull/266) — they just dropped the NetworkPolicy instead of trying to keep GitHub IPs up to date. You called that out too [here](https://redhat-internal.slack.com/archives/C04J47GL936/p1741090421731499?thread_ts=1741083675.576519&cid=C04J47GL936) (GitHub IPs being too dynamic). We never really closed the prod-sec question from the SeaLights thread.

happybhati: I am not sure what alert we can add about this we have the fallback to existing method if NP get blocked but I only tested in kind 
so let me see if I can get ROSA cluster to test the same again, and we can drop NP as integration-service did(can discuss offline).

happybhati: Tried the RELEASE-2681 operator (#1940) on a ROSA/GPTE OpenShift 4.20 sandbox. Same ingredients as stage: OVN NetworkPolicy, ubi-micro (no CA store), OpenShift trusted-ca inject.

Case 1 — GitHub CIDRs in the NP + trusted-ca mounted
Branch development on https://github.com/konflux-ci/release-service-catalog resolved to SHA 7174d754985d8b5e9dab2622f418b4182cbbe93e (taskGitRevision). Matches current development HEAD. SHA pin works when the pod can actually talk to GitHub over TLS.

Case 2 — same operator, GitHub CIDRs stripped (stage-like: github.com blocked)
Did not hang or fail-hard. PipelineRun was created with taskGitRevision=development (branch name fallback). Tekton later failed inside the pipeline, which is expected — the operator recovered, the catalog fetch still needs GitHub.

So: with this PR, “can’t reach GitHub” falls back the way it did after Sean reverted the stage image. It does not stall the Release the way #1753 did.

SHA pin on real stage/prod still needs egress (this PR’s CIDRs, or dropping that NP) and trusted-ca (infra-deployments #14128). This PR by itself should not re-break stage.

Code: [#1940](https://github.com/konflux-ci/release-service/pull/1940)

CA mount: [#14128](https://github.com/redhat-appstudio/infra-deployments/pull/14128)

